PR #10, opened 2023-10-06
Fix pathing and related things to allow rtd to autobuild docs

qciaran:


docs/conf.py
```diff
@@ -28,18 +28,17 @@
...
 
-from pecos import __author__, __version__
+from pecos import __version__
 
 # -- Project information -----------------------------------------------------
 
 project = "PECOS"
 copyright = (
-    "2018-2023, The PECOS Developers. Copyright 2018 National Technology & Engineering Solutions of Sandia, LLC (NTESS)"
+    "2018-2023 The PECOS Developers. Copyright 2018 National Technology & Engineering Solutions of Sandia, LLC (NTESS)"
```
qartik, line 39: @qciaran Sphinx requires the comma, see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-copyright

qartik: In fact, we can provide a sequence now:
> The value may now be a sequence of copyright statements in the above form, which will be displayed each to their own line.

qciaran: Ah okay, thanks!

PR #11, opened 2023-10-06
get rtd to install needed packages for documentation

qciaran:


.readthedocs.yaml
```diff
@@ -30,6 +30,6 @@ sphinx:
 # Optional but recommended, declare the Python requirements required
 # to build your documentation
 # See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
-# python:
-#    install:
-#    - requirements: docs/requirements.txt
+python:
+  install:
+    - requirements: requirements.txt
```
qartik, line 35: Shouldn't this be?
```suggestion
    - requirements: requirements/docs.txt
```

qartik: or perhaps add docs.txt in your primary requirements file.

qciaran: What it had by default is docs/requirements.txt. So yet another requirements file... We could create one, but I am currently just trying to get rtd to work.

qciaran: I guess I wont be lazy, and I'll add docs/requirements.txt. I'll do the copyright sequence too... :D

qartik: actually I was referring to the existing file: https://github.com/PECOS-packages/PECOS/blob/master/requirements/docs.txt

qciaran: That folder/file is in the `master` branch but was removed in the `development` branch. Although could add it back. Or do what sphinx suggested by default (i.e., `docs/requirements.txt`) and copy some of those in there. hmmmm

qartik: Ah, I didn't realize that.

qciaran:
Getting rtd to autobuild docs

PR #13, opened 2023-10-09
Until refactored, measurements need to accept **params

ciaranra:


ciaranra:
Fixes stabilizer measurement issues

PR #20, opened 2023-11-07
fix: errors found in end to end testing with pytket-phir

qartik:
The changes in this PR are currently needed to be able to run the `phirc` CLI tool available as part of https://github.com/CQCL/pytket-phir

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,10 @@ lint: metadeps  ## Run all quality checks / linting / reformatting
 # Testing
 # -------
 
-tests: venv install metadeps  ## Run tests on the Python package (not including optional dependencies)
+tests: install metadeps  ## Run tests on the Python package (not including optional dependencies)
 	$(VENV_BIN)/pytest tests -m "not optional_dependency"
 
-tests-dep: venv install-all metadeps ## Run tests on the Python package only for optional dependencies
+tests-dep: install-all metadeps ## Run tests on the Python package only for optional dependencies
 	$(VENV_BIN)/pytest tests -m optional_dependency
 
 doctests:  ## Run doctests with pytest
--- a/python/pecos/reps/pypmir/op_types.py
+++ b/python/pecos/reps/pypmir/op_types.py
@@ -19,11 +19,13 @@ class Op:
         args: list | None = None,
         returns: list | None = None,
         metadata: dict | None = None,
+        angles: tuple[list[float], str] | None = None,
     ) -> None:
         self.name = name
         self.args = args
         self.returns = returns
         self.metadata = metadata
+        self.angles = angles
 
         if returns is not None:
             for r in returns:
@@ -38,16 +40,27 @@ class Op:
                     msg = f"Returns not of correct form of cvar (str) or cbit ([str, int]): {returns}"
                     raise TypeError(msg)
 
+    def __str__(self) -> str:
+        return f"{self.name}, {self.args}, {self.returns}, {self.metadata}, {self.angles}"
+
 
 class QOp(Op):
     """Quantum operation."""
 
-    def __init__(self, name: str, args: list, returns: list | None = None, metadata: dict | None = None) -> None:
+    def __init__(
+        self,
+        name: str,
+        args: list,
+        returns: list | None = None,
+        metadata: dict | None = None,
+        angles: tuple[list[float], str] | None = None,
+    ) -> None:
         super().__init__(
             name=name,
             args=args,
             returns=returns,
             metadata=metadata,
+            angles=angles,
         )
 
 
--- a/python/pecos/reps/pypmir/op_types.py
+++ b/python/pecos/reps/pypmir/op_types.py
@@ -19,13 +19,11 @@ class Op:
         args: list | None = None,
         returns: list | None = None,
         metadata: dict | None = None,
-        angles: tuple[list[float], str] | None = None,
     ) -> None:
         self.name = name
         self.args = args
         self.returns = returns
         self.metadata = metadata
-        self.angles = angles
 
         if returns is not None:
             for r in returns:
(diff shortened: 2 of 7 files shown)


python/pecos/simulators/parent_sim_classes.py
```diff
@@ -34,13 +34,16 @@ def run_gate(self, symbol: str, locations: set[int] | set[tuple[int, ...]], **pa
         output = {}
 
         if params.get("simulate_gate", True):
-            for location in locations:
-                results = self.bindings[symbol](self, location, **params)
+            if locations:
```
qartik, line 37: This code block may require careful changes. I ran into a problem where two-qubit gates such as `RZZ` are being passed their arguments one by one (`location` here) and, hence failing.

See
```py
RZZ, [19, 0], None, {'angles': [0.0]}, None
{'angles': [0.0], 'angle': 0.0}
19
Traceback (most recent call last):
  File "/Users/kartik.singhal/GH/pytket-phir/./pc.py", line 329, in <module>
    main()
  File "/Users/kartik.singhal/GH/pytket-phir/./pc.py", line 300, in main
    print(HybridEngine(qsim="state-vector").run(program=p, shots=10))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kartik.singhal/GH/PECOS/python/pecos/engines/hybrid_engine.py", line 168, in run
    measurements = self.qsim.run(noisy_buffered_qops)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kartik.singhal/GH/PECOS/python/pecos/simulators/quantum_simulator.py", line 67, in run
    output = self.state.run_gate(op.name, op.args, **op.metadata)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kartik.singhal/GH/PECOS/python/pecos/simulators/parent_sim_classes.py", line 42, in run_gate
    results = self.bindings[symbol](self, location, **params)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kartik.singhal/GH/PECOS/python/pecos/simulators/projectq/gates_two_qubit.py", line 161, in RZZ
    q1 = state.qids[qubits[0]]
                    ~~~~~~^^^
TypeError: 'int' object is not subscriptable
```

python/pecos/simulators/quantum_simulator.py
```diff
@@ -61,6 +61,8 @@ def run(self, qops: list[QOp]) -> list:
         """
         meas = []
         for op in qops:
+            if op.metadata is None:
+                op.metadata = {}
```
qartik, line 65: Alternatively, perhaps, the default value of `op.metadata` could be `{}` in the constructor.

Makefile
```diff
@@ -55,10 +55,10 @@ lint: metadeps  ## Run all quality checks / linting / reformatting
 # Testing
 # -------
 
-tests: venv install metadeps  ## Run tests on the Python package (not including optional dependencies)
+tests: install metadeps  ## Run tests on the Python package (not including optional dependencies)
 	$(VENV_BIN)/pytest tests -m "not optional_dependency"
 
-tests-dep: venv install-all metadeps ## Run tests on the Python package only for optional dependencies
```
qartik, line 61: Calling `venv` deletes my virtual environment. I'd never want that while testing.

qciaran: This is intentional as I want to test in a clean environment; however, I can understand how that is frustrating in active development. ` build-full` probably satisfies my need for now. So feel free to remove `venv` from the two tests targets.

python/pecos/simulators/quantum_simulator.py
qciaran: I am not quite sure which constructor you mean. I think this probably serves both the `QuantumCircuit` and PHIR-based data structures. I would say, unless it is causing an issue, for now leave it as is as there probably needs to be addressed in another wave of refactoring. I would like to simplify and get rid of this but it might break other things.

python/pecos/reps/pypmir/op_types.py
```diff
@@ -19,11 +19,13 @@ def __init__(
         args: list | None = None,
         returns: list | None = None,
         metadata: dict | None = None,
+        angles: tuple[list[float], str] | None = None,
     ) -> None:
         self.name = name
         self.args = args
         self.returns = returns
         self.metadata = metadata
+        self.angles = angles
```
qciaran, line 28: I don't think a general operation needs angles attached to them as that is something special to QOps. Is this sort of thing needed for the pydantic translation, or you would rather have the parent class have all possible fields?

python/pecos/simulators/quantum_simulator.py
```diff
@@ -61,6 +61,10 @@ def run(self, qops: list[QOp]) -> list:
...
+            if self.backend == "state-vector" and op.name == "RZZ" and op.metadata["angles"][0] == 0.0:
+                continue
```
qciaran, line 67: I think I would rather have gate transformations (which I think this is essentially) live in the phir -> pypmir conversion and be the same for all simulators, when possible.

qciaran: It might be better to replace it with an "I" gate with the corresponding duration in case an error model or something down the pipeline is using durations to in act idling.

qciaran (changes requested):
Looks good except I have one question on adding angles to Ops and think converting RZZ(0) should be done earlier and  be replaced by "I" rather than skipped over.

qartik:
@qciaran thanks I just pushed updates with suggested changes. Let me know how it looks!

qciaran (approved):

qciaran:
Alright, changes looks good. Thanks!

PR #21, opened 2023-11-16
Feature more simulators

PabloAndresCQ:
Draft PR to gather initial feedback about current approach to add simulators.

Currently adding:

- A wrapper for cuStateVec
- A CoinToss  "simulator"


PabloAndresCQ:
Pre-commit failing because CI `ruff` output seems to disagree with my local `ruff` output. Might be a versioning issue?

Either way, hopefully this is fine for now just to gather initial feedback.

qartik:
> Pre-commit failing because CI `ruff` output seems to disagree with my local `ruff` output. Might be a versioning issue?

Yeah, most likely. Could you also include an update to ruff version in your PR to `v0.1.5` at
https://github.com/PECOS-packages/PECOS/blob/development/.pre-commit-config.yaml#L22

python/pecos/simulators/cointoss/bindings.py
```diff
@@ -0,0 +1,43 @@
+# TODO: Include license information?
+
+from pecos.simulators.cointoss.gates import ignore_gate, measure
+
+# Supporting gates from table:
+#   https://github.com/CQCL/phir/blob/main/phir_spec_qasm.md#table-ii---quantum-operations
```
qartik, line 6:
```suggestion
#   https://github.com/CQCL/phir/blob/main/spec.md#table-ii---quantum-operations

```

python/pecos/simulators/custatevec/bindings.py
```diff
@@ -0,0 +1,46 @@
+# TODO: Include license information?
+
+import pecos.simulators.custatevec.gates_one_qubit as one_q
+import pecos.simulators.custatevec.gates_two_qubit as two_q
+from pecos.simulators.custatevec.gates_init import init_zero
+from pecos.simulators.custatevec.gates_meas import meas_z
+
+# Supporting gates from table:
+#   https://github.com/CQCL/phir/blob/main/phir_spec_qasm.md#table-ii---quantum-operations
```
qartik, line 9:
```suggestion
#   https://github.com/CQCL/phir/blob/main/spec.md#table-ii---quantum-operations

```

PabloAndresCQ:
> You can locally test whether the CI will pass by issuing `pre-commit run --all-files` or `make lint`.

I'm doing:
```
python -m pip install pre-commit
pre-commit run --all-files
```
and everything passes in my machine, but not in CI :/ Ruff version seems to be 0.1.5.

If I try `make lint` I get `make: .venv/bin/python: Command not found` which seems to be just some absolute path I'd need to change.

qartik:
> `.venv/bin/python: Command not found` which seems to be just some absolute path I'd need to change.

It sounds like you were not inside a virtual environment. Consider doing the following listed under [installation instructions](https://github.com/PECOS-packages/PECOS/tree/development#installation):
```
python -m venv .venv
source .venv/bin/activate
pip install -U pip setuptools
pip install -r requirements.txt
pre-commit install
pip install -e .
```


I just pushed the fix so it should be good now.

pyproject.toml
```diff
@@ -57,6 +57,8 @@ simulators = [
     "cython",
     "pybind11>=2.2.3,<3.0",
     "projectq>=0.5.0,<0.9.0",
+    "cuquantum-python>=23.6.0",
+    "cupy>=10.4.0",
```
PabloAndresCQ, line 61: Looks like CI is no longer pleased with me requiring CuPy as an optional dependency:

>       ERROR: CUDA could not be found on your system.
>       
>       HINT: You are trying to build CuPy from source, which is NOT recommended for general use.
>             Please consider using binary packages instead.

Any suggestions on what to do? I'm happy to just remove CuPy and cuquantum-python from the optional dependencies. To install these you need a machine with a GPU, which I don't expect is the case for the machine running CI?

qartik: @PabloAndresCQ You can create a new dependency section named `gpu` or `cuda` (and refer it in README and elsewhere for this sim), like so:
```
gpu = [
    "cuquantum-python>=23.6.0",
    "cupy>=10.4.0",
]
```

Since `simulators` in included in `all` below, the CI step at https://github.com/PECOS-packages/PECOS/blob/development/.github/workflows/python-app.yml#L41-L45 is including these dependencies in it.

PabloAndresCQ: Awesome, thanks! :)

ciaranra (approved):
Code LGTM and new simulation work for me.

qciaran (approved):
LGTM too lol

PR #27, opened 2023-12-07
Bug phir if statement

qciaran:
Fixed a bug where for blocks the qop buffer wasn't being yielded all the way down to the engine.


qartik (approved):
LGTM

PR #34, opened 2023-12-15
Add resolved name

qciaran:
Adding a .sim_name to PyPMIR to resolve names, primarily for SQ gates, to simulators such as stabilizer sims will recognize. Future work may need to resolve TQ Clifford gates.


qciaran:
Note to self: Seems to currently work for RZZ, RZ, and R1XY. Make it work for general SQ gates.

qciaran:
Note to self: consider adding `op.clifford`, which gives the Clifford symbol if equivalent to a Clifford and 'False` if not.

PR #38, opened 2024-01-31
Initial support for barriers (skips barriers for now).

qciaran:
Adds barriers to PyPMIR

### DIFF
--- a/python/pecos/classical_interpreters/phir_classical_interpreter.py
+++ b/python/pecos/classical_interpreters/phir_classical_interpreter.py
@@ -53,6 +53,8 @@ class PHIRClassicalInterpreter(ClassicalInterpreter):
         self.cenv = None
         self.cid2dtype = None
 
+        self.phir_validate = True
+
         self.reset()
 
     def _reset_env(self):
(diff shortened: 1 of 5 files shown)


qartik (approved):

qciaran:
I should of added a quick barrier test. Let me add that before merging.

qciaran:
@qartik Alright, I added tests and fixed the errors I encountered.

python/pecos/classical_interpreters/phir_classical_interpreter.py
```diff
@@ -81,7 +83,7 @@ def init(self, program: str | (dict | QuantumCircuit), foreign_obj: ForeignObjec
             self.program = self.program.to_phir_dict()
 
         # Assume PHIR dict format, validate PHIR
-        if isinstance(self.program, dict):
+        if isinstance(self.program, dict) and self.phir_validate:
             PHIRModel.model_validate(self.program)
```
qartik, line 87: When does this validation kick in? Since we haven't made a new phir release yet, the JSON tests you added should fail validation checks if this line is reachable for them.

qartik: Ah, never mind, I see you setting them to `false` in the tests.

qartik (approved):

python/pecos/classical_interpreters/phir_classical_interpreter.py
qciaran: Yeah, I was in the awkward position that the `phir` package wont validate the barriers right now. I'll get rid of turning it off once I can depend a version of `phir` that allows it.

ciaranra (approved):

PR #53, opened 2024-03-09
PyPECOS

ciaranra:
Restructures the project


ciaranra:
This is essentially the old version of PR #77

PR #57, opened 2024-04-23
64 bit signed arithmetic support

qartik:
Finish `BinArray2` with 64-bit signed as the default type and migrate to it from `BinArray`.

Also, add a unit test file to expose interesting edge cases.


python/pecos/classical_interpreters/phir_classical_interpreter.py
```diff
@@ -261,8 +261,10 @@ def assign_int(self, cvar, val: int):
             cval &= ~(1 << i)
             cval |= (val & 1) << i
 
-        # mask off bits give the size of the register
-        cval &= (1 << size) - 1
+        if not isinstance(cval, (np.int64, np.int32)):
```
qartik, line 264: I wish I could exclude all signed types instead of enumerating them.

python/pecos/engines/cvm/binarray2.py
```diff
@@ -41,7 +43,8 @@ def set(self, value):
 
     def new_val(self, value):
         b = BinArray2(self.size, value, self.dtype)
-        b.clamp(self.size)
+        if self.dtype in (np.uint32, np.uint64):
```
qartik, line 46: I wish I could include all unsigned types instead of enumerating them.

tests/unit/test_binarray.py
```diff
@@ -0,0 +1,106 @@
+from typing import Final
+
+import numpy as np
+from hypothesis import assume, given
+from hypothesis import strategies as st
+from pecos.engines.cvm.binarray2 import BinArray2 as BinArray
+
+# from pecos.engines.cvm.binarray import BinArray
```
qartik, line 8:
```suggestion
```

tests/unit/test_binarray.py
qartik, line 1: FYI, these tests aren't comprehensive. A lot more can be done.

python/pecos/classical_interpreters/phir_classical_interpreter.py
qciaran: @dlucchetti Thoughts on not masking off signed ints?

python/pecos/engines/cvm/binarray2.py
qciaran: @dlucchetti same question...

dlucchetti (commented):
Seems reasonable.  Not 100% sure on the issue around masking width.  Seems that it may be best to integrate this to get some broader testing.  I believe it should only impact QIR.

python/pecos/classical_interpreters/phir_classical_interpreter.py
ciaranra: So maybe this is a future selves problem?

qartik: @dlucchetti @ciaranra It's not clear if there is something actionable here. Are we converging on merging this PR for and do some integration testing?

Like I mentioned elsewhere, I did some of this integration testing using this branch already with the internal emulator and compiler tests. Please let me know if there is more we can do here before merging.

qciaran: It does sound like maybe future problems... although not having a consistent approach gets me a little antsy...

What PR were you considering merging this with?

I haven't looked closely at the tests yet. I was planning to sit down and look at them and play around with the implementation. Should we be moving some tests from the internal emulator or compiler to here for better coverage/regression testing?

qartik: > What PR were you considering merging this with?

I meant this PR into `development` branch and making a new release. Though it sounds like you don't feel comfortable with it just yet.

> I haven't looked closely at the tests yet. I was planning to sit down and look at them and play around with the implementation.

Thanks, that might be nice. I don't have the best picture of what was happening in the existing `BinArray` with the unsigned bitstring assumptions.

> Should we be moving some tests from the internal emulator or compiler to here for better coverage/regression testing?

That would be great. I don't know if it makes sense to move any compiler tests here, but if the PHIR classical interpreter was also using `BinArray2`, we could create several tests as PHIR input.

ciaranra: > I meant this PR into `development` branch and making a new release. Though it sounds like you don't feel comfortable with it just yet.

I think I might be fine with it. Just feels a little bit like weird/undefined behavior. I wonder if the solution would just be to raise an exception if a size argument was given for signed ints when PHIR is translated to the Python representation in PECOS.

> That would be great. I don't know if it makes sense to move any compiler tests here, but if the PHIR classical interpreter was also using `BinArray2`, we could create several tests as PHIR input.

Hmmm, good point... although maybe not worth if we move away from `BinArray2` to the PHIR interpreter...

Alright maybe for this PR I will just look more closely into the tests/play around with it. Although, if there is a obvious place to insert an exception for signed ints with size... that might be good to add.

qartik: > I wonder if the solution would just be to raise an exception if a size argument was given for signed ints when PHIR is translated to the Python representation in PECOS.

I tried that, but we currently have several test cases that use that, e.g. https://github.com/PECOS-packages/PECOS/blob/development/tests/integration/phir/example1_no_wasm.json#L12-L25

> maybe not worth if we move away from `BinArray2` to the PHIR interpreter...

Oh, is PHIR interpreter going to the primary classical interpreter?

python/pecos/reps/pypmir/pypmir.py
```diff
@@ -25,6 +27,20 @@
...
+    "i64": np.int64,
+}
+
+unsigned_data_types = {
+    "u8": np.uint8,
+    "u16": np.uint16,
+    "u32": np.uint32,
+    "u64": np.uint64,
+}
+
```
qartik, line 43: Unsure if this is the best module to place these data types.

python/pecos/reps/pypmir/pypmir.py
```diff
@@ -229,10 +245,14 @@ def from_phir(cls, phir: dict, name_resolver=None) -> PyPMIR:
                 name = o["data"]
 
                 if name == "cvar_define":
+                    data_type = o["data_type"]
+                    size = int(data_type[1:])
+                    if "size" in o:
+                        size = o["size"]
```
qartik, line 251: This takes care of default sizes for signed integers, but overrides them if specified.

ruff.toml
```diff
@@ -112,6 +112,9 @@ ignore = [
     # TODO: Remove to improve error handling...
     "TRY002", # Custom exceptions
     "TRY003", # Avoid specifying long messages outside the exception class
+
+    # UP031 should be enabled after fixing all suggestions in a separate PR
+    "UP031",
```
qartik, line 117: Ruff made a bit of a change that's affecting a lot of files: https://github.com/astral-sh/ruff/pull/11019 Ignoring for now.

tests/integration/test_phir_64_bit.py
```diff
@@ -0,0 +1,35 @@
...
+    phir = {
+        "format": "PHIR/JSON",
+        "version": "0.1.0",
+        "ops": [
+            {"data": "cvar_define", "data_type": "i32", "variable": "var_i32"},
+            {"data": "cvar_define", "data_type": "u32", "variable": "var_u32", "size": 32},
+            {"data": "cvar_define", "data_type": "i64", "variable": "var_i64"},
+            {"data": "cvar_define", "data_type": "u64", "variable": "var_u64", "size": 64},
+            {"data": "cvar_define", "data_type": "i32", "variable": "var_i32neg"},
+            {"data": "cvar_define", "data_type": "i64", "variable": "var_i64neg"},
```
qartik, line 18: I omit `"size"` in the signed variables above, which currently fails PHIR validation. Created a PR to allow this (the spec already marks the size field as optional): https://github.com/CQCL/phir/pull/84

python/pecos/classical_interpreters/phir_classical_interpreter.py
qciaran: > Oh, is PHIR interpreter going to be the primary classical interpreter?

What I mean is that internal emulator does not use PHIR/the PHIR interpreter right now. However, it would be nice to switch it to that so there is one emulation process. The PHIR interpreter doesn't use `BinArray`/`BinArray2`.

python/pecos/reps/pypmir/pypmir.py
qciaran: It's probably fine for now. I have been thinking some refactoring of PyPMIR might be good. So more thought can probably be put into a future PR.

qciaran (changes requested):
I tried some examples myself. Everything worked and the code makes sense.

So LGTM except I think we should keep to 0.6.0.dev* and I will try to work on getting to a release sooner than later.

qciaran (approved):
LGTM

ciaranra (approved):
LGTM

PR #55, opened 2024-04-08
Adding a numpy-based basic statevector simulator for more exhaustive testing

PabloAndresCQ:
- Adding a numpy-based basic statevector simulator.
- Adding more exhaustive tests comparing simulators against the numpy basic simulator.

PR #58, opened 2024-04-26
Update: Update cuquantum elements to support 24.0

dlucchetti:
Update for the latest cuquantum library 0.24.0 interface.  Tested on A100.


python/pecos/simulators/custatevec/bindings.py
```diff
@@ -51,6 +54,7 @@
     "SYY": two_q.SYY,
     "SYYdg": two_q.SYYdg,
     "SZZ": two_q.SZZ,
+    "SZZ": two_q.SZZ,
```
qartik, line 57:
```suggestion
```

dlucchetti: Fixed!  That Actually was supposed to have the SqrtZZ op.  Copy-paste error.

python/pecos/simulators/custatevec/state.py
```diff
@@ -97,6 +99,13 @@ def free(ptr, size, stream):
         mem_handler = (malloc, free, "GPU memory handler")
         cusv.set_device_mem_handler(self.libhandle, mem_handler)
 
+    def reset(self):
+        """Reset the quantum state for another run without reinitializing."""
+        # Initialize all qubits in the zero state
```
PabloAndresCQ, line 104: As far as I can tell, this is not resetting the state to the zero state. The lines you would need to add to achieve so are:

```
        # Allocate the statevector in GPU and initialize it to |0>
        self.vector = cp.zeros(shape=2**num_qubits, dtype=self.cp_type)
        self.vector[0] = 1
```

PabloAndresCQ (changes requested):
Thanks Dominic! Indeed, there was a breaking change introduced in the latest release of custatevec, I had not notice!

Just one comment below on your new `reset` function.

PabloAndresCQ:
I found a bug on the cuStateVec wrapper, on the definition of the RYY gate: https://github.com/PECOS-packages/PECOS/pull/55/commits/3e17bc01acedd0a4cf6f8d77aa1d84b0c98246c7

Perhaps it's worth doing a cherry-pick of this commit to this branch? Apologies, I should probably have created a PR for that bugfix on its own, rather than add it on a feature PR.

On an unrelated note: you might want to update the version requirement in the `tests/integration/state_sim_tests/test_statevec.py` file, line 24. Alternatively, we could just remove that check from the test file, it is unnecessary since the pyproject.toml is already dealing with versioning, so we should just check whether it is installed or not.

PabloAndresCQ:
@dlucchetti I'm happy to take this off your hands and add the change I suggested + make the CI checks pass for this PR. Give me a 👍 if that'd be helpful.

PabloAndresCQ:
I ended up creating a separate branch, since there's some stuff in this PR I'm not sure I understand. The new PR is #64 and it includes most of the changes appearing here. The differences are:
- Lack of `init_helper`
- Different implementation of `reset`
- No special case for `"angle" in params` in the `RZ` gate. I expect that, if this is necessary, we would need it in all of the single-qubit gate methods, not just the `RZ`. If this is a fix that was needed for the use of `HybridEngine`, have a look at #63 where I did a small change to `parent_sim_classes.py` which appears to be enough to solve this.

ciaranra:
This looks like an older PR @dlucchetti  do we want to still move forward with this? If so, looks like there are changes still to be made.

ciaranra:
This PR still relevant or should it be closed? @dlucchetti

ciaranra:
Closing due to likely being stale

PR #62, opened 2024-05-09
Adding specific error messaging and enforced maximum duration to WASM calls

nealerickson-qtm:


tests/integration/test_phir.py
qciaran, line 1: From the file history, this file was first created in 2023. So far I have been putting the year at which the file was first created/edited in the copyright. However, I believe it is fairly common to include a range of years. We could switch to that format (I might need to double check this is what is done.).

python/pecos/foreign_objects/wasmtime.py
qciaran, line 1: LGTM or sorry about that. :D

tests/integration/test_phir.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 The PECOS developers
+# Copyright 2024 The PECOS developers
```
qciaran, line 1:
```suggestion
# Copyright 2023 The PECOS developers
```

ciaranra (changes requested):
LGTM as long as the date for the test_phir.py file is reverted.

ciaranra (approved):

ciaranra:
LGTM again

PR #64, opened 2024-05-22
cuStateVec bugfixes and slight refactor

PabloAndresCQ:
The objective of this PR is to solve some issues with the cuStateVec simulator, which were appearing in PRs #55 and #58. The intention is that this PR is merged first, then the latter two updated accordingly.

This PR does the following changes:
- Updated calls to cuStateVec to match new API from cuStateVec 1.6.0 (duplicates approach from #58).
- Added some more bindings to cuStateVec (duplicates approach from #58)
- Adds a reset method to cuStateVec (corrected from #58)
- Adds a bugfix on RYY gate (commit cherry-picked from #55)
- Refactors `self.vector` so that it returns a `numpy` array for easier integration and testing (duplicates approach from #55)

Tests passing on NVIDIA GPUs using cuQuantum 24.03 (and cuStateVec 1.6.0).


python/pecos/simulators/custatevec/bindings.py
```diff
@@ -28,6 +36,7 @@
     "RY": one_q.RY,
     "RZ": one_q.RZ,
     "R1XY": one_q.R1XY,
+    "RXY1Q": one_q.R1XY,
     "SX": one_q.SX,
     "SXdg": one_q.SXdg,
```
dlucchetti, line 41: A few more gates that we should define...  

SqrtX == SX
SqrtXd == SXdg
SqrtYd == SY
SqrtYd == SYdg
SqrtZd == SZ
SqrtZd == SZdg

Seen these in circuits in the wild, so seems like we need them.

dlucchetti (commented):
Thank you for the updated PR, Pablo!  I have a couple of additional gates that would be good to include, but otherwise looks good!  

Thanks again!

python/pecos/simulators/custatevec/bindings.py
PabloAndresCQ: Sure thing, I've added these in. 

However, rather than adding these synonyms manually to each individual simulator, I think we should make some choice of canonical names (e.g. those from the [PHIR specs](https://github.com/CQCL/phir/blob/main/spec.md#table-ii---quantum-operations)) and add a "centralised translator" in `parent_sim_classes.py` (or similar) that all commands go through before reaching the backend. I'm happy to create a separate PR to add this in, if it'd be helpful.

dlucchetti: I think the problem is that we have made multiple decisions on the canonical names over time and every shift creates a compatibility problem when interfacing with the package.  Being a bit more exhaustive temporarily and calling out which would be deprecated seems like the right balancing act for ease of forward movement?

ciaranra: Yeah, I think being more exhaustive right now is fine. If it isn't already, very soon, everything can conform to the PHIR spec naming. (There might be some older style names like SqrtX left over in internal software that need to be fixed, and then we can clean up PECOS.)

pyproject.toml
```diff
@@ -59,7 +59,8 @@ simulators = [
     "projectq>=0.5.0,<0.9.0",
 ]
 cuda = [
-    "cuquantum-python>=23.6.0",
+    "cuquantum-python>=24.03.0",
+    "custatevec>=1.6.0",
     "cupy>=10.4.0",
```
ciaranra, line 64: One thing that is slightly annoying is that I cannot install this directly from pip in Linux. So `pip install .[all,cuda]` fails. (I could probably figure it out and get cupy, e.g., install via binaries or building from source...) If I install via `conda install -c conda-forge cuquantum-python>=24.03.0`, conda sits there thinking for a very long time. `conda install -c conda-forge cuquantum-python` worked for me and installs the latest version. It also installs both custatevec and cupy. So the extra custatevec and cupy requirements are not needed, at least for conda.

Long story short... I was able to use conda to install and successfully run the associated tests... It feels like maybe not for this PR but in the future we should maybe document how to install (at least reference the appropriate urls). And ideally add a makefile target to set it all up for development.

ciaranra (approved):
LGTM. Tests pass. Nothing out of place that I can see.

PR #67, opened 2024-06-27
Update: Some minor fixes.

dlucchetti:
As it was there were some issues with cuquantum when the reset was invoked and for some gate invocations with singular invocation names.


ciaranra:
LTGM however looks like you made ruff angry. (I don't know why I had to manually approve the workflow... otherwise, you could of seen this sooner. weird)

python/pecos/simulators/custatevec/state.py
```diff
@@ -101,8 +103,10 @@ def free(ptr, size, stream):
     def reset(self):
         """Reset the quantum state for another run without reinitializing."""
         # Initialize all qubits in the zero state
-        self.vector = cp.zeros(shape=2**self.num_qubits, dtype=self.cp_type)
-        self.vector[0] = 1
+        # self.vector = cp.zeros(shape=2**self.num_qubits, dtype=self.cp_type)
+        # self.vector[0] = 1
+        cusv.destroy(self.libhandle)
+        self.__init_helper()
```
PabloAndresCQ, line 109: Why is this necessary? The docstring explicitly says that there's no reinitialisation (following the dosctring that existed for ProjectQ), but here we're doing initialisation from scratch, including reinitialisation of the library via `cusv.create()`, which is a noticeable overhead.

Also, how is it ensuring that the state is initialised to `|0>`?

dlucchetti: I'll dig in a bit more here.  In my testing I was seeing plenty of examples where not starting from scratch was causing issues, but I may be pulling out inappropriate weaponry to resolve those issues.  My *naive* understanding of what I have requested is that in destroying the reference / re-initializing it I have requested that it start *completely* fresh and that (likely due to data leakage concern?) it likely zeros out the memory when starting up.  

I have done many rounds of GHZ state creation tests which return 'reasonable' results which to me suggests that it is not being polluted by end-state, but I would like to dig in just a bit more to appreciate if the current flow is what *should* be done.  

Speed alone is itself a good reason to re-investigate.

PabloAndresCQ: Hmm, I see. Could you share more details on what issues you were getting (and the examples that caused those issues)? 

I've just realised that in the version of `reset` you have updated, what's being reset is `self.vector`, but what keeps the state of the computation is `self.cupy_vector`. Perhaps that's the bug that was causing the issue? See [here](https://github.com/PECOS-packages/PECOS/pull/68/files#diff-24c8d5cdf817ad6597c96f264700fabc826fce21235245bf8c7f22cc41b39fb6L104). Apologies -- I should have pointed it out when I created that PR (somehow, it didn't register as a bug in my mind; but it obviously is).

To be honest, if that's the bug, I'm quite surprised that your approach solved it. With `cusv.destroy()` and `cusv.create()` you are reinitialising the CuStateVec library, but the vector itself is kept by CuPy, so I would expect it's not affected. I suppose that since we pass the pointer of `self.cupy_vector` to CuStateVec, it's free to do whatever with it, and it _may_ zero it out at some point... but it's undocumented behaviour as far as I can tell, and I wouldn't be surprised if it was the source of bugs later on if there are internal changes on how CuStateVec works.

ciaranra:
This PR still relevant or should it be closed? @dlucchetti

ciaranra:
Closing due to likely being stale

PR #68, opened 2024-07-02
[feature] Adding support for QuEST and Qulacs statevector simulators

PabloAndresCQ:
This PR adds the following:

- Support to run statevector simulations on [Qulacs](https://docs.qulacs.org/en/latest/intro/0_about.html).
- Support to run statevector simulations on [QuEST](https://quest.qtechtheory.org/).
- Support and tests of HybridEngine with Qulacs, QuEST and CuStateVec.

### Additional notes

- I have done a [survey](https://cqc.atlassian.net/wiki/spaces/RESEARCH/pages/3053191172/Survey+of+statevector+simulators) of statevector simulators and concluded that Qulacs and QuEST were the best third-party statevector simulators to support.
- Beyond 20-25 qubits, the simulators from Cirq, Qiskit and QPanda become dominant. 
  - I have not provided support for these because they do not support "on demand gates and measurements" required to use `HybridEngine`. 
  - Besides, I expect that beyond this point `CuStateVec` would beat these CPU-based simulators, so there might not be a need to support these. 
    - This conjecture is _not_ supported by the benchmarking paper I've based my survey on. In it, Qiskit using cuQuantum is shown to have poor performance. I believe this is misrepresenting the power of CuStateVec, though, and it'd be worth doing some benchmarking on our side.
- I believe we should provide support for a 32-bit precision statevector simulator. 
  - This can be easily done for `CuStateVec` and I'd be happy to do it (in another PR) if there's interest. 
  - Moreover, QuEST supports 32-bit precision and it's particularly good at it. Jake would be keen on providing support for this (in another PR) if there's interest.
  - Looking at the survey paper (see confluence page above), it's not obvious that there'll be any noticeable speed-up (at least on CPU-based simulators). Maybe this varies between computers, though. The difference in terms of memory is just providing an extra qubit, which is not going to make that much of a difference.


PabloAndresCQ:
These last two commits correspond to small bugfixes in the code added by this PR, which I've found while working on the MPS PR.

This PR is ready for review.

ciaranra:
LGTM. The tests look great. I tried it all out on Linux and Windows. No problems.

ciaranra (approved):
LGTM

PR #72, opened 2024-08-26
Allow passing ancillas to `Steane` class

perlinm:
This should allow different logical qubits to share ancillas, with something like
```python
program = pecos.slr.Main(
    ancillas := pecos.slr.QReg("ancillas", size=3),
    logical_qubit_0 := pecos.qeclib.steane.steane_class.Steane("lq_0", ancillas=ancillas),
    logical_qubit_1 := pecos.qeclib.steane.steane_class.Steane("lq_1", ancillas=ancillas),
    ...
)
```

Warning: I am not sure how to test for unintended side effects, so I'm not sure whether this breaks anything...

UPDATE: for what it's worth, I've played around with some simulations that do/don't use this PR, and (1) this PR runs without error, and (2) the results from both simulations look to be identical.

### DIFF
--- a/python/pecos/qeclib/steane/steane_class.py
+++ b/python/pecos/qeclib/steane/steane_class.py
@@ -33,10 +33,10 @@ class Steane(Vars):
     This represents one particular choice of Steane protocols. For finer control construct your own class
     or utilize the library of Steane code protocols directly."""
 
-    def __init__(self, name: str, default_rus_limit: int = 3):
+    def __init__(self, name: str, default_rus_limit: int = 3, ancillas: QReg | None = None):
         super().__init__()
         self.d = QReg(f"{name}_d", 7)
-        self.a = QReg(f"{name}_a", 3)
+        self.a = ancillas or QReg(f"{name}_a", 3)
         self.c = CReg(f"{name}_c", 32)
 
         # TODO: Make it so I can put these in self.c... need to convert things like if(c) and c = a ^ b, a = 0;
@@ -59,7 +59,6 @@ class Steane(Vars):
 
         self.vars = [
             self.d,
-            self.a,
             self.c,
             self.syn_meas,
             self.last_raw_syn_x,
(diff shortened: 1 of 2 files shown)


perlinm:
TODO: throw an error if the ancillas that were passed to a `Steane` instance were not initialized in the parent circuit (`pecos.slr.Main`).

perlinm:
On second thought, I think the TODO above is overkill, or at least out-of-scope for this PR.  PECOS already allows for something like
```python
qubit = pecos.qeclib.steane.steane_class.Steane("lq")
program = pecos.slr.Main(qubit.qec())
```
which I actually find to be useful to represent and examine subcircuits.  Leaving it to the user to make sure that ancillas are properly declared seems reasonable on similar grounds IMO.

One alternative that I considered (and pushed to this PR, but then tentatively reverted) is to have `Steane` instances always keep track of their ancillas (as done [here](https://github.com/PECOS-packages/PECOS/blob/feat-qeclib-steane/python/pecos/qeclib/steane/steane_class.py#L60-L62)), but then (to deal with the fact that multiple Steane/Vars instances will contain the same ancillas/Var object) modify `Vars.extend` to something like
```python
class Vars:
    ...

    def extend(self, vars_obj: "Vars"):
        if isinstance(vars_obj, Vars):
            # identify and add only new variables
            new_vars = [var for var in vars_obj.vars if var not in self]
            self.vars.extend(new_vars)
        else:
            ...

    def __contains__(self, var: object) -> bool:
        return any(var is self_var for self_var in self.vars) 
```
This way the same ancillas object won't be added to `pecos.slr.Main` multiple times.  However, I reverted this proposed change because IMO just making `Steane` not keep track of "external" ancillas seems (A) cleaner and (B) consistent with current PECOS "practices".

qciaran:
I'll look into edge cases with multiple or no declaration later. It seems to be something that needs to be studied more deeply, e.g., more tests and checking proper error messaging in general.

ciaranra (approved):
LGTM

PR #73, opened 2024-08-29
[Bugfixes] Solving some issues for qemulator MPS release

PabloAndresCQ:
This PR solves some issues found when updating qemulator to be able to use `MPS`, `QuEST`, `Qulacs` and `CoinToss`.


ciaranra (approved):
Changes make sense

PR #74, opened 2024-08-30
Use `default_rus_limit` in `Steane` class

perlinm:
Some `Steane` methods (such as [`Steane.pz`](https://github.com/PECOS-packages/PECOS/blob/feat-qeclib-steane/python/pecos/qeclib/steane/steane_class.py#L90-L93)) ask for an `rus_limit: int | None = None` argument (`rus` = repeat until success).  If `rus_limit is None`, these methods set `rus_limit = self.default_rus_limit`.

However, some of the newer `Steane` methods require an `rus_limit: int | None` argument, without setting a default.  This PR sets the default to `None`.


qciaran (approved):
LGTM. It does indeed seem these were missing.

PR #76, opened 2024-09-13
Typo fixed in the implementation of `p` and `tdg` in `steane_class.py`

PabloAndresCQ:


ciaranra (approved):
Looks great! Thanks for finding those typos!

This suggests we need to start implementing tests to verify we are getting what we expect.

PR #78, opened 2024-09-19
Fix bug in `Steane.prep_tdg_plus_state`

perlinm:
`Tdg |+> = Sdg T |+>`


ciaranra (approved):
Looks right. Good catch.
RZ(-pi/4) = RZ(-pi/2)RZ(pi/4) 
RZ(-pi/4) != RZ(pi)RZ(pi/4)

PR #82, opened 2024-10-11
Fix RUS reject bits and add QEC flag bits

perlinm:
Specifically:
- The RUS reject bit for logical Steane T|+> state prep was previously set incorrectly (it was always 0).  This PR fixes that.
- This PR also adds a reject bit for other Steane RUS state prep protocols.
- This PR also adds the option to set a flag bit to 1 if a QEC cycle had nontrivial syndromes.


python/pecos/qeclib/steane/steane_class.py
```diff
@@ -218,40 +197,31 @@ def nonft_prep_t_plus_state(self):
...
                 flag_z=self.flag_z,
                 flags=self.flags,
                 last_raw_syn_x=self.last_raw_syn_x,
                 last_raw_syn_z=self.last_raw_syn_z,
-                limit=rus_limit,
+                limit=rus_limit or self.default_rus_limit,
             ),
         )
-        if reject is not None:
-            block.extend(reject.set(self.log))
```
perlinm, line 241: This is an example of the `reject` bit being set incorrectly.  It is always set to `self.log`, even though `self.log` is unused and does not change in this method.

perlinm:
The last commit (`8f0e928a41ef75dd68b8a1403a76f37c0d015820`, "Automatic lint fixes") has greatly **_increased_** the diff with `PECOS-packages:development`, as it has applied a bunch of changes that are unrelated to this PR.  Maybe these fixes should be applied to `PECOS-packages:development` and this PR should merge those changes instead (to better keep track of what is being changed in this PR)?

perlinm:
I (tentatively) reverted the lint fixes in `8f0e928a41ef75dd68b8a1403a76f37c0d015820`, which mostly changed code that was not touched by this PR.  This makes it easier to see what is actually being changed here.  We can add the lint fixes back once the changes are reviewed (though I still think that it would be best to make the lint fixes in the upstream branch).

ciaranra:
Alright. No worries about reverting. I sort of like getting everything to pass first. But this works. Sorry for the slow response. Will try to get to this as soon as I can.

perlinm:
Thanks!  A minor update on this front: while RUS rejection bits (specifically for T|+> states) do not behave as expected in PECOS simulations (both state vector and MPS), rejection rates seem reasonable in (a) my own state vector simulations of PECOS-produced QASM, and (b) actual hardware experiments.  So I suspect that there is funny business happening with something common to the simulators.

python/pecos/qeclib/steane/steane_class.py
ciaranra: Yeah, that is obviously wrong.

python/pecos/qeclib/steane/steane_class.py
```diff
@@ -97,115 +97,94 @@ def __init__(self, name: str, default_rus_limit: int = 3, ancillas: QReg | None
...
-            rus_limit = self.default_rus_limit
-
         return PrepRUS(
             q=self.d,
             a=self.a[0],
             init=self.verify_prep[0],
-            limit=rus_limit,
+            limit=rus_limit or self.default_rus_limit,
             state="+X",
+            reject=reject or self.scratch[2],
```
ciaranra, line 127: I think potentially moving the bit outside of the logical qubit object gets me bit nervous as a user might accidentally having different logical qubits override the same register. But fine if the value is constant. So `rus_limit` is probably okay. (Maybe worth adding a Const type.) 

Generally, I would like information generated by operations on logical qubits saved "privately" and when a user want it made public, to copy it to another place, e.g., see the pattern in the logical measurements.

python/pecos/qeclib/steane/qec/qec_3parallel.py
```diff
@@ -48,3 +49,5 @@ def __init__(
             FlagLookupQASMActiveCorrectionX(q, syn_x, syndromes, last_raw_syn_x, pf_z, flag_x, flags, scratch),
             FlagLookupQASMActiveCorrectionZ(q, syn_z, syndromes, last_raw_syn_z, pf_x, flag_z, flags, scratch),
         )
+        if flag_bit is not None:
+            self.extend(If(flags != 0).Then(flag_bit.set(1)))
```
ciaranra, line 53: I don't think extra information should be extracted from lower level operations. To prevent the accumulation of arguments, it is most preferable for this info to be extracted at the user level. Next, best would be the class operations. Perhaps the best compromise might be to expose flag/syndrome information to the user and have them if off of that information.

ciaranra: Maybe I need to think about making variables obviously private using underscores and provide getter methods to expose useful info publicly to make it easier for users.

python/pecos/qeclib/steane/steane_class.py
perlinm: IMO this is fully on the user.  If I write a classical program, I get to decide when to set what variables, and how to use those variables throughout the program.  Here also I am allowing the user to set variables (bit registers) that user manages.  In principle, the user could even add things like conditional logic to the program based on the values of those variables.

I think it **does not** make sense to save all rejection bits to `Steane` class variables.  The Steane class only has (and, IMO, only should have) a "local" view of info/data attached to one logical qubit at the time of initialization.  It should not know global data such as (say) how many RUS protocols are in the program written by the user, which determine the size of a register that stores all RUS rejection bits.

perlinm: Sorry -- I am using strong language like "should" above for ease of communication.  I am ultimately just after data like RUS rejection bits and QEC/QED syndrome flags, so I'm happy to discuss how to best extract that info 🙂

python/pecos/qeclib/steane/qec/qec_3parallel.py
perlinm: I am not sure what you mean by "lower level operations", but as you suggest what I am really trying to do here is expose flag/syndrome info.  My thinking was to give the user the reigns here because it seems to me like the `Steane` class cannot know a priori how many QEC cycles are in a program, and therefore how many syndrome bits to save (or even whether to save them).

python/pecos/qeclib/steane/steane_class.py
ciaranra: There might of been some confusion. I do not wish each Steane class to know the global information. It was more about privately protecting information locally for the Steane code and not let the users accidently override it by replacing the private register as I believe what you were code is proposing would do. Doing so could cause one register to be assigned to different independent codes.

ciaranra:
Re-looking through this with fresh eyes and forgetting any fancy language tricks like introducing const, etc. I think I would be okay with this if the user assigned variables (i.e., `reject` and `flag_bit`) were moved out of qeclib and moved into `steane_class` and done in a similar way to `log` in `def mz()`.

python/pecos/qeclib/steane/steane_class.py
perlinm: The option to pass a `reject` bit certainly lets the user override `Bit`s assigned to a `Steane` class -- which was also true before this PR.  Nothing prevents a user from using (say) `logical_qubit.flag_z[1]` somewhere in their `pecos.slr.Main` program.  Adding underscores to signal private data may be a good idea, but I would argue is outside the scope of this PR.

One point to note: as far as I can tell, the scratch bit that rejection data was stored to (and is still stored to by default) is not actually used for anything by the `Steane` class, and moreover is not even stored in any persistent memory.  The `Steane` class simply wrote rejection data to `self.scratch[2]`, and happily kept overwriting that bit at every RUS protocol.  The same bit is also written over by various methods in `python/pecos/qeclib/steane/decoders/lookup.py` (which is fair play!  That's why it's called a `scratch` bit 🙂).  This PR lets a user save rejection data to a (persistent) bit register of their choice.

perlinm:
1. I consolidated the changes in this PR so that everything is inside `steane_class.py` (plus one typo fix).
2. What is the difference between how `reject` and `flag_bit` are used vs. `log` in `def mz`?  I thought I was using them the same way.

ciaranra:
Thanks!
The difference is that the information being gathered/functionality being added is just for the user and not required for the functionality of the library (only for gathering info for the user-side program). Essentially, I think it is a better separation of concerns. Plus, I think it is better to maintain style consistency. Additionally, it buries the optional user thing less further down layers of abstraction.

python/pecos/qeclib/steane/steane_class.py
```diff
@@ -218,40 +162,31 @@ def nonft_prep_t_plus_state(self):
...
-
-        block = Block(
+        return Block(
             self.scratch.set(0),
             PrepEncodeTPlusFTRUS(
                 d=self.d,
                 a=self.a,
                 out=self.scratch,
-                reject=self.scratch[2],
+                reject=reject or self.scratch[2],
```
ciaranra, line 171: No longer supports `reject`

```suggestion
```

python/pecos/qeclib/steane/steane_class.py
```diff
@@ -97,117 +97,61 @@ def __init__(self, name: str, default_rus_limit: int = 3, ancillas: QReg | None
...
-        return PrepRUS(
+        block = PrepRUS(
             q=self.d,
             a=self.a[0],
             init=self.verify_prep[0],
-            limit=rus_limit,
-            state="+X",
+            limit=rus_limit or self.default_rus_limit,
+            state=state,
+            reject=reject or self.scratch[2],
```
ciaranra: `reject` kwarg is no longer supported in this version. Looks like this is leading to an error in the tests.

```suggestion
```

perlinm: Oops.  Fixed 🙂

perlinm:
@ciaranra is the current use of `reject` and `flag_bit` correct then?  Where "current use" is something like having `def p(...)` accept a `reject: Bit` and end with
```python
        if reject is not None:
            block.extend(reject.set(self.verify_prep[0]))
```
analogously to how `def m(...)` accepts a `log: Bit` and ends with
```python
        if log is not None:
            block.extend(log.set(self.log))
```
`Steane.qec` uses `flag_bit` the same way as well.

perlinm:
Tangentially related to this PR: the `reject` bit in `Steane.prep_t_plus_state` is required.  Should I make it optional?

UPDATE: I went ahead and made it optional, along with all cases of `log: Bit | None` (which previously allowed a `None` argument but did not set a default).  Although in retrospect maybe this was deliberate syntactic salt to make sure that unrecorded measurements are deliberate.  If so, let me know and I can change this back.

perlinm:
It looks like to pass some regression tests we would need to change a bunch of the `tests/regression/test_qasm/regression_qasm/local_steane_code_program.telep_*.qasm` files.

Looking through (say) [this error log](https://github.com/PECOS-packages/PECOS/actions/runs/12146586712/job/33870854277?pr=82) I think the changes to the QASM files are correct, but given the sensitive nature of regression tests I'll wait to mess with them pending your guidance @ciaranra.

ciaranra:
@perlinm Alright, sounds good. I'll take a look at the regression tests shortly and try to fix them. I heard from some others about approaches/packages that make these sort of tests less onerous (walks through the regression tests, asks if the diffs are okay and rebuilds the regression test if yes).

ciaranra:
(unless you want to fix them @perlinm. right now it has to be a boring manual fix.)

perlinm:
I'm happy to report that the regression test fixes were simple and explainable.  It took just a few short regex commands 🙂

In total there were three one-line changes to make to each of the `tests/regression/test_qasm/regression_qasm/*t_gate_*.qasm` files.  The regex find+replace commands for these changes (from the `tests/regression/test_qasm/regression_qasm` directory) are...

(syntax: `sd -F "<FIND>" "<REPLACE>" <FILES>`)

1 and 2.
```sh
sd -F "m_reject[0] = saux_c[2];" "m_reject[0] = saux_scratch[2];" *t_gate_*.qasm
sd -F "m_reject[1] = saux_c[2];" "m_reject[1] = saux_scratch[2];" *t_gate_*.qasm
```
These changes can be attributed to [this diff](https://github.com/PECOS-packages/PECOS/pull/82/files#diff-bb85cf9a4613b22cc4b2fab58eeeaf45fe64b111b2abec3682ee6e1f6e21fd50L241-R180) (old line 241, new line 180) in `Steane.prep_t_plus_state`, fixing a bug in which the `reject` bit was incorrectly assigned `self.log` (= `self.c[2]`) instead of `self.scratch[2]` (see [this comment](https://github.com/PECOS-packages/PECOS/pull/82#discussion_r1796247027)).

3.
```sh
sd -F "m_reject[0] = sin_c[2];
" "" *t_gate_*.qasm
```
This change can be attributed to the removal of [these lines](https://github.com/PECOS-packages/PECOS/pull/82/files#diff-bb85cf9a4613b22cc4b2fab58eeeaf45fe64b111b2abec3682ee6e1f6e21fd50L314-L317) (old lines 314--317), which set the `reject` bit in `Steane.t` (likewise in `Steane.tdg`) to `self.log` (= `self.c[2]`).  Aside from setting `reject` to the wrong bit (as in `Steane.prep_t_plus_state`, above), these lines were redundant with the fact that `prep_t_plus_state` already sets the `reject` bit.

python/pecos/qeclib/steane/steane_class.py
```diff
@@ -97,117 +97,46 @@ def __init__(self, name: str, default_rus_limit: int = 3, ancillas: QReg | None
...
-                msg = f"State {state} is not implemented!"
-                raise NotImplementedError(msg)
-
-    def px(self, rus_limit: int | None = None):
-        """Prepare logical |+X>, a.k.a. |+>"""
-
-        if rus_limit is None:
-            rus_limit = self.default_rus_limit
+        if state in ["X", "Y", "Z"]:
+            state = f"+{state}"
```
perlinm: As minor cleanup, I made `Steane.p` call `PrepRUS`, and then turned `Steane.pz/pnz/px/etc.` into one-line calls to `Steane.p`.  As a result,  match/case here was redundant with the match/case in in PrepRUS -> LogZeroRot ([here](https://github.com/PECOS-packages/PECOS/blob/a0989e7aec73dadbbf610a6b75ceb1df66db8c4b/python/pecos/qeclib/steane/preps/pauli_states.py#L153-L179)), so it can be removed.

perlinm: I likewise made `Steane.mx/my/mz` call `Steane.m`.

perlinm: Incidentally, @ciaranra would you want to make "X", "Y", and "Z" recognized as "+X", "+Y", and "+Z" in [`LogZeroRot`](https://github.com/PECOS-packages/PECOS/blob/a0989e7aec73dadbbf610a6b75ceb1df66db8c4b/python/pecos/qeclib/steane/preps/pauli_states.py#L153-L179) as well?

ciaranra: Yeah, that would be good to do to keep things consistent.

perlinm: done 🙂

ciaranra (approved):
Looks good. Thanks for the PR!

Issue #86, opened 2024-10-25
`KeyError: 'H4'` with certain parameter values


cqc-alec:
Version: quantum-pecos 0.6.0.dev5

```python
from pecos.engines.hybrid_engine import HybridEngine

phir = """
{
    "format": "PHIR/JSON",
    "version": "0.1.0",
    "metadata":
    {
        "source": "pytket-phir v0.8.1"
    },
    "ops":
    [
        {
            "data": "qvar_define",
            "data_type": "qubits",
            "variable": "q",
            "size": 1
        },
        {
            "qop": "R1XY",
            "angles":
            [
                [
                    1.0,
                    1.75
                ],
                "pi"
            ],
            "args":
            [
                [
                    "q",
                    0
                ]
            ]
        }
    ]
}
"""

engine = HybridEngine(qsim="state-vector")
engine.run(phir, shots=10)
```
Output:
```
Traceback (most recent call last):
  File "/home/alec/r/pytket-quantinuum/play/john.py", line 51, in <module>
    r = engine.run(phir, shots=10)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alec/r/pytket-quantinuum/env/lib/python3.12/site-packages/pecos/engines/hybrid_engine.py", line 191, in run
    temp_meas = self.qsim.run(noisy_qops)
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alec/r/pytket-quantinuum/env/lib/python3.12/site-packages/pecos/simulators/quantum_simulator.py", line 95, in run
    output = self.state.run_gate(op.sim_name, op.args, **op.metadata)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alec/r/pytket-quantinuum/env/lib/python3.12/site-packages/pecos/simulators/parent_sim_classes.py", line 43, in run_gate
    results = self.bindings[symbol](self, location, **params)
              ~~~~~~~~~~~~~^^^^^^^^
KeyError: 'H4'
```

If "1.75" is changed to "1.76", there is no error!

cqc-alec:
The bug does not appear with quantum-pecos 0.6.0.dev3: it seems to have been introduced in quantum-pecos 0.6.0.dev4.

PabloAndresCQ:
[...]

PR #87, opened 2024-10-28
More Clifford gates supported in state-vector and MPS simulators

PabloAndresCQ:
This fixes the bug reported in #86 

It simply adds to the simulators the Clifford gates listed [here](https://github.com/PECOS-packages/PECOS/blob/0d7b4f9b3032b0e61031e6dbc1979ab94693d9db/pecos/circuit_converters/std2chs.py#L55).


ciaranra:
Hmmm, should probably just turn off the Clifford conversion for simulators supporting non-Clifford gates. But this works well too. Working out the right way to turn on and off the conversion might be a more complex discussion.

ciaranra (approved):
It looks like all appropriate gates have been added.

PR #96, opened 2024-12-09
Unroll broadcasted operations when producing QASM from SLR

### LINKED ISSUE #95: `QReg`s not permuted by `pecos.slr.Permute`
It looks like `pecos.slr.Permute` correctly permutes individual qubits, but does not relabel qubit registers.  What I mean is best understood by an example:
```python
#!/usr/bin/env python3
import pecos.qeclib.qubit
import pecos.slr

qreg_a = pecos.slr.QReg("qreg_a", 1)
qreg_b = pecos.slr.QReg("qreg_b", 1)
program = pecos.slr.Main(
    pecos.slr.Permute(qreg_a, qreg_b),
    pecos.qeclib.qubit.X(qreg_a),
    pecos.qeclib.qubit.X(qreg_a[0]),
)

print(program.qasm())
```
Based on my understanding of how `Permute` is supposed to work, I expect this script to print X gates on `qreg_b` and `qreg_b[0]`.  However, on the current `PECOS:development` branch (commit `eca0ce7af8057a7fbd6e512d499081e8bb9bffad`), the above script instead prints
```sh
OPENQASM 2.0;
include "hqslib1.inc";
// Generated using: PECOS version 0.6.0.dev6
// Permuting: qreg_b[0] -> qreg_a[0]
x qreg_a;
x qreg_b[0];
```
so the qubit is being "relabeled" properly, but the register is not.  Regardless of my prior expectations, this behavior does not seem to be self-consistent.  Any use of `Permute` therefore messes up all following QASM in a `pecos.slr.Main` program.

In particular, this issue currently prevents us from running logical circuits with (1) the teleportation variant of a T gate ([`Steane.t_tel`](https://github.com/PECOS-packages/PECOS/blob/eca0ce7af8057a7fbd6e512d499081e8bb9bffad/python/pecos/qeclib/steane/steane_class.py#L270-L285)), and (2) Steane-type QEC with one-bit teleportation.

Not sure if this helps, but the issue **might** be in how qubits/registers/etc are permuted in [`gen_qasm.process_permute`](https://github.com/PECOS-packages/PECOS/blob/eca0ce7af8057a7fbd6e512d499081e8bb9bffad/python/pecos/slr/gen_codes/gen_qasm.py#L348).  The circular references between `QReg`s and `Qubit`s make it a bit difficult for me to reason about how to solve the issue, though...

FWIW, I think using `Permute` would be unnecessary if hardware could "natively" recognize a virtual SWAP instruction in QASM, as in something like `swap qreg_0[0] qreg_1[0];`.  That would be both a nice feature to have and a more elegant solution than permuting register names in QASM.

perlinm:
fixes #95

In my view this fix is a bit of a hack around peculiarities of the `Permute` instruction, which (at least in its current implementation) is fundamentally inconsistent with broadcasting instructions across a qubit register (see [comment](https://github.com/PECOS-packages/PECOS/issues/95#issuecomment-2526567715)).  The fix in this PR is to just unroll broadcast operations, and make them act on the individual qubits of a register.

I'm happy to fix the broken tests if+when the solution in this PR is approved.

If I were to opine a bit: in my view `Permute` looks to be a re-invention of the SWAP gate, and it would be much more favorable to just use SWAPs instead of `Permute`s, and in turn have hardware recognize SWAPs as virtual (handled by some below-QASM level of control software, essentially mutating the map from qubit label --> physical qubit/ion).  Inserting a single SWAP/Permute into a circuit shouldn't have the effect of (a) modifying all downstream QASM and (b) making visual inspection and debugging of QASM impossible, because the QASM no longer agrees with the SLR it was produced from (asking SLR to `pecos.qeclib.qubit.X(qreg_a[0])` can now produce the QASM `x qreg_b[2]`, for example).  And of course broadcasting instructions across a register is a nice and natural capability to have.  Having said that, we still need `Permute` for now, so hopefully this is an acceptable fix 🙂


python/pecos/slr/gen_codes/gen_qasm.py
```diff
@@ -291,14 +292,22 @@ def qgate_qasm(self, op, repr_str: str | None = None):
...
+            elif isinstance(q, tuple):
                 if len(q) != op.qsize:
                     msg = f"Expected size {op.qsize} got size {len(q)}"
                     raise Exception(msg)
                 qs = ",".join([str(qi) for qi in q])
                 str_list.append(f"{repr_str} {qs};")
 
-            str_list.append(f"{repr_str} {str(q)};")
+            else:
+                str_list.append(f"{repr_str} {q};")
```
perlinm: It looks to me like that the unconditional final `str_list.append(...)` was a bug, and should have been inside an `else` statement.  Otherwise this method first acts the gate on every `qs in q` and then again appends a gate on (the tuple) `q`.

ciaranra: To me, SWAP is a unitary gate that could possibly be converted to a physical permute if a compiler is smart enough. There is a need to distinguish the two operations as you may what fine control and choose one over the other and do not want to rely on a compiler.

As far as rewriting QASM, the user should no longer be responsible with debugging that. This is the job of the language (e.g., SLR)/compiler. If the underlying base language had support for Permute, then that would make things cleaner; however, SLR needs to metaprogram features that are not supported by the base language which is essentially what higher languages have to do when converting to lower languages. They provide more facilities then the simpler underlying language provides.

That being said. I want to rewrite the Permute object to be a little more friendly to code generation and just stores the info in a nice manner to be handed off to the code generator function.

perlinm: Upgrading SWAP to a "native gate" of the hardware is perfectly compatible with thinking of it as a unitary gate.  Relabeling qubits **_does_** implement the desired unitary operation, and it is an **advantage** of trapped ion platform that it can perform SWAP gates perfectly 🙂!  Not all platforms can do this.  This is also a capability that can be leveraged by circuit-level compilers (search [this paper](https://arxiv.org/pdf/2309.05157) for "swap mirroring", for example), but doing so requires that the hardware recognize SWAP as a native gate.  This essentially "exposes" the virtual SWAP capability to compilers that work on the level of (say) `pytket`, `cirq`, or `qiskit`.

There is also precedent here: having a virtual native SWAP gate is directly analogous to virtual Rz gates on IBM devices.  IBM devices accept Rz instructions, but do not actually apply any physical operations any qubits when asked to "apply" an Rz gate (they instead make a software-level update of a rotating frame).

From the perspective of compiler development, the way to get "fine control" and execute a specific sequence of physical operations is to build the circuit you want, and run it in "verbatim" mode to bypass/disable any compiler.  Another option is to have compilation flags that toggle various compiler features.  There is no control or choice removed by upgrading the SWAP gate to a native virtual operation (and in fact, at the moment you already "rely on a compiler" to convert a SWAP instruction into 3 MS gates).

Final point:
> As far as rewriting QASM, the user should no longer be responsible with debugging that.

As a matter of practice, we have definitely resorted to examining QASM to debug PECOS and would appreciate the option to continue doing so 🙂

In any case, what are your thoughts of unrolling broadcast operations in this PR?  Some fix for #95 is necessary to use both `Permute` and (say) `Steane.sz` in the same SLR (and `Permute` is used for `Steane.t_tel`).

ciaranra (approved):
LGTM

perlinm:
Thanks for the review and test fixes!

PR #99, opened 2024-12-10
Bump py 0.6.0.dev7

ciaranra:
Bump the Python PECOS version from 0.6.0.dev6 to 0.6.0.dev7


ciaranra:
Didn't mean to branch off of feat/color488

PR #101, opened 2024-12-10
Fixing that macos aarch64 and x86_64 zips are producing the same wheels

ciaranra:
Trying to fix wheel building

### DIFF
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -159,7 +159,7 @@ jobs:
           fi
 
       - name: Test wheel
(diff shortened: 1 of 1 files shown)


qartik:
It might be a slightly heavy-weight solution but cibuildwheel is quite amazing and helps avoid lot of future maintenance. See https://cibuildwheel.pypa.io/en/stable/

ciaranra:
Hmmmm, okay. I'll take a look at it. I got things working now though. I'll put in an issue to explore that.

ciaranra:
Related issue added: #102

PR #111, opened 2024-12-15
Upgrade to using uv instead of pip

ciaranra:
Use uv instead of pip for workflows and development


Makefile
```diff
@@ -30,37 +30,24 @@ venv:  ## Build Python virtual environment
...
-	$(VENV_BIN)/pip install -r temp_requirements.txt
-
-	@echo "Cleaning up temporary files..."
-	@rm temp_requirements.txt
+installreqs: ## Install Python project requirements from lockfiles
+	@unset CONDA_PREFIX \
+	&& $(VENV_BIN)/python -m pip install --upgrade uv
+	@echo "Installing from lockfiles..."
+	cd python/pecos-rslib && $(VENV_BIN)/uv sync
+	cd python/quantum-pecos && $(VENV_BIN)/uv sync
```
qartik: If you are using `uv` then it should not be installed inside a virtual environment, instead it should be installed globally on a system (ie, README should say it's required). It handles the virtual environment for you and can even handle python versions.

Then to get the project ready, all one needs to do is `uv sync` assuming all dependencies are correctly specified in `pyproject.toml`.

Specifically, you will not need to prefix with `$(VENV_BIN)/` anywhere in the `Makefile`.

ciaranra: Ah just saw this. Yeah, I think I happened to arrive at the same conclusion. Switching this to a draft PR

ciaranra: hmmm, looking through things I see more `$(VENV_BIN)/` that I can get rid of using `uvx` although I seem to be having issues getting `pytest` to work. :thinking:

ciaranra: NVM, `uv run pytest` works and `uvx pytest` doesn't for some reason.

qartik: `uvx` shouldn't generally be needed for a project. Its primary goal is to be able to run a python command that's not even installed on the system, along the lines of [pipx](https://github.com/pypa/pipx). Only useful for system utilities or rare invocations of such programs, IMHO.

ciaranra: Makes sense. Reading up on `uvx`, it sounds like it creates an isolated environment to run a command while `uv run` uses your venv. So makes sense that what you say is best practice. thanks!

.github/workflows/python-release.yml
```diff
@@ -66,13 +66,13 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
+      - name: Install uv
+        run: pip install uv
+
```
qartik:
```suggestion
      - name: Install the latest version of uv
        uses: astral-sh/setup-uv@v4
        with:
          enable-cache: true
```

You don't need/want to be using `pip` for installations.

.github/workflows/python-release.yml
```diff
@@ -130,13 +128,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install uv
+        run: pip install uv
+
```
qartik: same as above

.github/workflows/python-release.yml
```diff
@@ -187,29 +183,31 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
+      - name: Install uv
+        run: pip install uv
+
```
qartik: ditto (also these steps will not require repeating once cibuildwheel is in use)

.github/workflows/python-test.yml
```diff
@@ -50,25 +47,9 @@ jobs:
...
       run: |
-        python -m pip install --upgrade pip setuptools
-        pip install maturin pre-commit
-        echo "Temporarily removing pecos-rslib from requirements... because we need to build it"
-        grep -v "pecos-rslib" python/quantum-pecos/requirements.txt > temp_requirements.txt
-        echo "Installing project requirements (excluding pecos-rslib)..."
-        pip install -r temp_requirements.txt
-        echo "Cleaning up temporary files..."
-        rm temp_requirements.txt
+        python -m pip install --upgrade uv
```
qartik: better to use https://github.com/astral-sh/setup-uv

.github/workflows/python-release.yml
```diff
@@ -233,24 +231,23 @@ jobs:
...
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install uv
+        run: pip install uv
+
```
qartik: last one

DEVELOPMENT.md
```diff
@@ -14,34 +14,17 @@ For developers who want to contribute or modify PECOS:
    ```sh
    make venv
    ```
-   <details>
-   <summary>or manually set up a Python virtual environment for develop of this project's code.</summary>
 
-   On Linux/Mac:
+   or manually set up a Python [virtual environment using uv](https://docs.astral.sh/uv/getting-started/installation/) for develop of this project's code.
```
qartik:
```suggestion
   or manually set up a Python [virtual environment using uv](https://docs.astral.sh/uv/getting-started/installation/) for development.
```

DEVELOPMENT.md
```diff
@@ -14,34 +14,17 @@ For developers who want to contribute or modify PECOS:
...
+   To do so, navigate to the root of this project and run:
 
    ```sh
-   python -m venv .venv
-   .\venv\Scripts\activate
-   pip install -U pip setuptools
-   pip install -r python/quantum-pecos/requirements.txt
-   make metadeps
-   pre-commit install
+   python -m pip install --upgrade uv
```
qartik: `uv` is best installed systemwide, I'd point people to https://docs.astral.sh/uv/getting-started/installation/ instead of suggesting `pip install` here.

Makefile
```diff
@@ -5,90 +5,53 @@ PYTHONPATH := $(shell which python 2>/dev/null || which python3 2>/dev/null)
...
-
 # Virtual environment
 # -------------------
 
 .PHONY: venv
-venv:  ## Build Python virtual environment
-	$(PYTHONPATH) -m venv $(VENV)
+venv:  ## Create virtual environment and install all dependencies
+	@echo "Installing uv..."
+	$(PYTHONPATH) -m pip install --upgrade uv
```
qartik: If the `Makefile` is intended to do the `uv` install, which I recommend against, it may be better to use `pipx install uv`. My recommendation is to ask people to install `uv` systemwide and only provide `sync` step here.

Makefile
```diff
@@ -5,90 +5,53 @@ PYTHONPATH := $(shell which python 2>/dev/null || which python3 2>/dev/null)
...
-		--output-file=python/quantum-pecos/requirements.txt \
-		--strip-extras $(PYPROJECT_PATH)
-	@echo "Restoring original pyproject.toml..."
-	@mv $(PYPROJECT_PATH).bak $(PYPROJECT_PATH)
-	@echo "Adding pecos-rslib back to requirements.txt..."
-	@echo "pecos-rslib==$(call get_version)" >> python/quantum-pecos/requirements.txt
-	@echo "numpy==1.26.4 ; python_version < \"3.13\"" >> python/quantum-pecos/requirements.txt
+updatereqs:  ## Generate/update lockfiles for both packages
+	@echo "Ensuring uv is installed..."
+	$(PYTHONPATH) -m pip install --upgrade uv
```
qartik: If installed system-wide, `uv` has a self update command: https://docs.astral.sh/uv/getting-started/installation/#upgrading-uv

README.md
```diff
@@ -100,7 +100,7 @@ pecos = "0.x.x"  # Replace with the latest version
 ## Development Setup
 
 If you are interested in editing or developing the code in this project, see this
-[development documentation](development.md) to get started.
+[development documentation](DEVELOPMENT) to get started.
```
qartik:
```suggestion
[development documentation](DEVELOPMENT.md) to get started.
```

qartik (commented):
My primary suggestion is not use pip to install `uv` anywhere. I haven't tried any steps yet, will be happy to once this concern is settled.

Makefile
ciaranra: I'll leave it in just to lower the entry bar but recommend installing uv system wide over using the make command in DEVELOPMENT.md. It looks like `pipx` is not install by default, so rather then requiring something else to install, I'll just leave it as it is for now.

ciaranra:
Okay, here goes with the workflows... I think I had difficulty using the uv actions for everything last night but hopefully can get them working for everything....

ciaranra:
Woooh hooo workflows seem to be working

ciaranra:
Alright, I think I have fixed all the issues you raise @qartik 

Going to see if all the artifacts look okay.

ciaranra:
Artifacts seemingly check out

DEVELOPMENT.md
qartik, line 1: Tried the instructions, they work. Now that we are using `uv`, mentions of venv and virtual environment both here and in Makefile (make venv) feel unnecessary. I will let you take an executive decision.

Makefile
```diff
@@ -5,90 +5,53 @@ PYTHONPATH := $(shell which python 2>/dev/null || which python3 2>/dev/null)
...
-metadeps: upgrade-pip  ## Install extra dependencies used to develop/build this package
-	$(VENV_BIN)/pip install -U setuptools pip-tools pre-commit maturin
-
-# Function to extract version from pyproject.toml
-define get_version
-$(shell grep -m 1 'version\s*=' $(PYPROJECT_PATH) | sed 's/.*version\s*=\s*"\(.*\)".*/\1/')
-endef
+installreqs: ## Install Python project requirements to root .venv
+	@echo "Installing requirements..."
+	@unset CONDA_PREFIX && uv sync
```
qartik: `@unset CONDA_PREFIX &&` shows up in a lot of places in this Makefile, why is it necessary? Even if it was set, isn't `uv` taking care of the correct paths?

qartik (approved):
Left a question and a minor suggestion, LGTM.

Makefile
ciaranra: Probably. Will remove for now.

PR #114, opened 2024-12-16
Permute all variables of the Steane code

perlinm:
~Note that the current implementation of `Permute`-ing classical registers may result in errors before #116 is fixed.~

~One possible solution to this issue is to construct classical logic that swaps the values of classical registers.  On further thought, I'll leave this PR as a draft until that is implemented...~

Classical registers are now swapped "manually" to avoid issue #116

### DIFF
--- a/python/quantum-pecos/src/pecos/qeclib/steane/steane_class.py
+++ b/python/quantum-pecos/src/pecos/qeclib/steane/steane_class.py
@@ -424,7 +424,9 @@ class Steane(Vars):
         )
         for var_a, var_b in zip(self.vars, other.vars):
             if isinstance(var_a, CReg):
-                block.extend(var_a.set(var_a ^ var_b))
-                block.extend(var_a.set(var_b ^ var_a))
-                block.extend(var_a.set(var_a ^ var_b))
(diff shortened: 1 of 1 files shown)


python/quantum-pecos/src/pecos/qeclib/steane/steane_class.py
```diff
@@ -370,19 +370,17 @@ def cz(self, target: Steane):
...
+            q=self.d,
+            meas_basis=meas_basis,
+            meas=self.raw_meas,
+            log_raw=self.log_raw,
+            log=self.log,
+            syn_meas=self.syn_meas,
+            pf_x=self.pf_x,
+            pf_z=self.pf_z,
+            last_raw_syn_x=self.last_raw_syn_x,
+            last_raw_syn_z=self.last_raw_syn_z,
```
perlinm, line 381: This is just some minor cleanup that comes for the ride in this PR

perlinm:
Note that this permutation may result in errors before #116 is fixed.

ciaranra (approved):
LGTM

Issue #112, opened 2024-12-16
Lookup decoder checks for `flag == 6` when flag is a `Bit`


perlinm:
I'm not sure what exactly is going on, but I do not think this makes sense to do.  See this [code block](https://github.com/PECOS-packages/PECOS/blob/07aafb73a48a41c437a7b6d0f4017491d7dabe92/python/quantum-pecos/src/pecos/qeclib/steane/decoders/lookup.py#L30-L77) (first line says that `flag: Bit`, last line checks for `flag == 6`).

perlinm:
I think this is just a typing error, fixed in #115

perlinm:
fixed in #115

PR #115, opened 2024-12-16
Add Steane-type and teleportation-based QEC for the Steane code

perlinm:
Teleportation-based QEC is essentially just one-bit teleportation, in which only one type of error (X or Z) is preserved in the teleported state.  Teleporting in two different ways thereby corrects all errors.

I used state vector simulations to verify that these circuits correct all physical single-qubit errors by:
1. Initializing a logical Steane qubit in a random logical state.
2. Applying a single-qubit X, Y, or Z error.
3. Running a `qubit.steane_qec` cycle.
4. Asserting that the (physical) state vector of the logical qubit is equal to the state vector before the applied error (up to global phase).

I am not entirely certain that I am using/storing classical registers correctly, so @ciaranra  please take a look.

Finally: the methods here write a `flag` qubit to remember whether there were non-trivial syndromes.  It may make sense to "guarantee" that this flag qubit is set correctly and remove [this suggestive comment](https://github.com/PECOS-packages/PECOS/blob/07aafb73a48a41c437a7b6d0f4017491d7dabe92/python/quantum-pecos/src/pecos/qeclib/steane/decoders/lookup.py#L232).  Thoughts?


perlinm:
Something funny is going on in my testing, so I'm going to mark this as a draft until I figure it out...

UPDATE: I think I fixed it.  I'm still not sure I'm setting classical registers correctly though.

python/quantum-pecos/src/pecos/qeclib/steane/steane_class.py
```diff
@@ -400,7 +402,7 @@ def mz(self, log: Bit | None = None):
         """Logical destructive measurement of the logical Z operator."""
         return self.m("Z", log=log)
 
-    def qec(self, flag_bit: Bit | None = None):
+    def qec(self, flag: Bit | None = None):
```
perlinm: I also renamed `flag_bit` --> `flag` in this PR, for consistency with conventions elsewhere in the repo.

perlinm:
I think this PR fixes #112, which appears to just be an incorrect type hint

ciaranra:
Good questions. At first glance everything looks good... but I'll comb over everything shortly.

More future facing stuff (not this PR):
I am putting the final touches on SLR->PHIR, which should allow for establishing proper internal tests for most of the code (might need to think about non-Clifford stuff). Hopefully that will help confirming behavior.  Also, probably would be good to do a documentation sweep and check/establish for style consistency and provide guides in the documentation so we can keep straight how things are approached.

ciaranra (approved):
LGTM

PR #118, opened 2024-12-17
Add T gates with active error correction

perlinm:


perlinm:
@ciaranra I'm actually realizing that, in direct analogy with `qec_steane_[xz]` and `qec_tel_[xz]` in #115, there are actually four variants of the T gate with "partial QEC".  This PR has a newly added `t_cor` with active correction of X errors, and the old `t_tel` with passive correction of X errors.  ~It is also possible to construct variants of [`t_cor`, `t_tel`] that [actively, passively] correct Z errors.~ (UPDATE: I am no longer sure of this claim.)

~I think I should add all four variants (`t_[cor|tel]_[xz]`) to this PR.  Or maybe we can merge this PR and add a "completion" with all four variants in a follow-up PR.  Should I simply rename the old `t_tel` (breaking backwards compatibility), or should I leave a `t_tel` method that calls one of `t_tel_x` or `t_tel_z` by default?~ (UPDATE: no longer relevant)

~Ditto all of the above for the `tdg_*` methods, although as a separate question: to simplify and clean up the zoo of T gate methods, what do you think of making all the `tdg*` methods just call a matching `t*` method followed by `sdg`?  This is what the method to prepare a Tdg state does, for example.~ (UPDATE: no longer relevant)

ciaranra (approved):
LGTM

PR #120, opened 2025-01-23
Feat/more sims

ciaranra:
Adds additional simulators: state vector sim and Pauli propagation. Plus appropriate traits.


qciaran:
Will merge this functionality with a larger PR

PR #122, opened 2025-02-05
fix argument to SlrConverter.generate

perlinm:
`skip_headers` is a positional argument to `SlrConverter.generate`, so currently calling `SlrConverter.qasm` simply fails.  While we're at it, this PR also makes `skip_headers` a positional argument to `SlrConverter.qasm` for consistency, and fixes what I think is a type hinting error(?).


perlinm:
Looks like this issue was fixed in commit [9f29a1e21968b6bfa1ce5ffb5c8d18a08e433f12](https://github.com/PECOS-packages/PECOS/commit/9f29a1e21968b6bfa1ce5ffb5c8d18a08e433f12), so I'll close it.

PR #125, opened 2025-02-28
Raise an error if a user tries to construct a `CReg` with more than 64 bits

perlinm:
My understanding is that `CReg`s are implemented by 64-bit integers.  This PR adds a helpful error message as soon as a user tries to instantiate an invalid `CReg`.

### DIFF
--- a/python/quantum-pecos/src/pecos/slr/vars.py
+++ b/python/quantum-pecos/src/pecos/slr/vars.py
@@ -124,6 +124,9 @@ class CReg(Reg, PyCOp):
             result: Whether this register is a result register (default True)
         """
 
+        if size > 64:
+            raise ValueError(f"Classical registers are limited to storing 64 bits (requested: {size})")
+
         super().__init__(sym, size, elem_type=Bit)
         self.result = result
 


ciaranra:
SLR should be agnostic to the implementation of the underlying language. Such checks should be at the code gen level.

perlinm:
Makes sense.  Where does the actual restriction to 64-bit registers come into play then?

ciaranra:
Just introduced the fix here for commit 7b59de2a2bf21f9476945b52cd71f34773c5b3f7 on PR #121

ciaranra:
Closing since resolved by new commit to PR #121

perlinm:
@ciaranra I don't this issue is resolved.  I get a missing package error from just trying to import `pecos.slr`:
```python
$ python
Python 3.11.9 (main, Apr 19 2024, 11:43:47) [Clang 14.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pecos.slr
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/D648438/src/PECOS/python/quantum-pecos/src/pecos/slr/__init__.py", line 17, in <module>
    from pecos.slr.slr_converter import SlrConverter
  File "/Users/D648438/src/PECOS/python/quantum-pecos/src/pecos/slr/slr_converter.py", line 16, in <module>
    from pecos.slr.gen_codes import Language, QASMGenerator, QIRGenerator
  File "/Users/D648438/src/PECOS/python/quantum-pecos/src/pecos/slr/gen_codes/__init__.py", line 13, in <module>
    from pecos.slr.gen_codes.gen_qir import QIRGenerator
  File "/Users/D648438/src/PECOS/python/quantum-pecos/src/pecos/slr/gen_codes/gen_qir.py", line 18, in <module>
    from llvmlite import binding, ir
ModuleNotFoundError: No module named 'llvmlite'
```

perlinm:
Should there be a size check for `SLR --> QASM` as well?

UPDATE: oh right, that doesn't make sense.  Then new question: when does/should this size check happen in a `SLR --> QASM --> job submission` pipeline?

ciaranra:
Yeah, the QASM thing is a bit tricky of where it makes sense to check the CReg size 🤔 The implementation at Quantinuum used to be limited to 32 but it might be 64 on all machines now... I think OpenQASM itself doesn't limit it. So... Depends on how we are thinking about this QASM... What SLR/Quantinuum is supporting is a super set of OpenQASM 2.0... 

At the very least, it is currently checked during compilation when sent for submission to Quantinuum machines. You can also run code through the syntax checker if you just want to check the syntax and not run on hardware or emulators.

So checked during `job submission --compile--> run`

PR #128, opened 2025-03-03
Fixes test warnings

ciaranra:


ciaranra:
Appears to be an issue due to the difference between numpy <2 and >2

Issue #116, opened 2024-12-17
Classical registers are set incorrectly after being `Permute`d


perlinm:
This example:
```python
import pecos.slr
prog = pecos.slr.Main(
    a := pecos.slr.CReg("a", size=1),
    b := pecos.slr.CReg("b", size=1),
    pecos.slr.Permute(a, b),
    a[0].set(1),
    a.set(1),
)
print(prog.qasm())
```
prints
```sh
...
// Permuting: b[0] -> a[0]
b[0] = 1;
a = 1;
```
This behavior will cause problems when using the values of classical registers that have been permuted.

perlinm:
One resolution is to have `Permute(a, b)` actually swap classical bits by compiling down to something like
```
a ^= b;
b ^= a;
a ^= b;
```

ciaranra:
This should hopefully be fixed in PR #130. The XOR trick you suggested is used for whole register permutation. A temporary register and assignments are used for bit assignments. Therefore, CReg/Bit symbols/indices don't change since classical information is moved. QReg/Qubit symbols do change (relabeling is used) since quantum information can't be copied and we want to avoid gating.

ciaranra:
I believe this was solved... feel free if you (@perlinm ) feel differently, and I'll reopen this.

PR #130, opened 2025-03-08
Getting permute working for both QASM and QIR

ciaranra:
Should allow things like `Permute([a[0], b[3], c[1]], [c[1], a[0], b[3]]]` as well as `Permute(a, b)`, which is just shorthand for `Permute([a[0], ..., a[n], b[0], ..., b[n]], [b[0], ..., b[n], a[0], ..., a[n]])`.

Hopefully didn't screw up any other code generation. All other previous code gen tests pass.


ciaranra:
Not quite right yet... working to address Issue #116

ciaranra:
New pushes might solve #116 but probably should run test cases for QIR through a simulator that supports the language

Issue #147, opened 2025-07-07
Scaling of `with_p_average_idle_*` methods in Rust PECOS seems wrong


PabloAndresCQ:
In Python (see [line](https://github.com/CQCL/qemulator/blob/4fafeee6fd472c1cd7e596ab6258d0e870c0a883/python/qemulator/quantinuum/error_model/error_h1.py#L160)):

```
# Experimentalists are reporting average error rate for dephasing, need to convert to total.
self.error_params["quadratic_dephasing_rate"] *= np.sqrt(3 / 2)
self.error_params["linear_dephasing_rate"] *= 3 / 2
```

However, in Rust:
- In `p_average_idle_linear_rate` there is a square root:
https://github.com/PECOS-packages/PECOS/blob/71604d5c6a3a1d476a485dd6b10ed182032c32fa/crates/pecos-engines/src/noise/general/builder.rs#L340
- In `p_average_idle_quadratic_rate` the square root acts on both the rate and the 3/2 constant, but it should only act on the constant:
https://github.com/PECOS-packages/PECOS/blob/71604d5c6a3a1d476a485dd6b10ed182032c32fa/crates/pecos-engines/src/noise/general/builder.rs#L362

PR #148, opened 2025-07-07
Fix `with_p_average` scaling

PabloAndresCQ:
- Fix for #147 
- Renamed `with_p_average_*` to `with_average_p_*` assuming the previous name was a typo.


qciaran:
LGTM

Issue #151, opened 2025-07-12
Rusty RNGModel


ciaranra:
Create a Rust version of `RNGModel` as found in rng_model.py. This would allow for Rust side usage, for example for pecos-qasm as seen in the `refactor-n-memory` branch. Once created, expose the `RngModel` struct to Python via pecos-rslib using PyO3 and replace the Python `RNGModel`.

qciaran:
@Jonhas-qtm, I assume you would want to handle this sort of thing?

Jonhas-qtm:
> [@Jonhas-qtm](https://github.com/Jonhas-qtm), I assume you would want to handle this sort of thing?

Yes, I'll have this assigned to me

Jonhas-qtm:
Resolved as of https://github.com/PECOS-packages/PECOS/pull/167

PR #152, opened 2025-07-12
Feat/minimal pcg wrapper

ciaranra:
Use minimal Rust wrapper for PCG as well as use Rust for the build system. This avoids adding dependencies like cmake and additional wheel building and verification. It also starts the process of making PCG available for pure Rust QASM simulation as mentioned in Issue #151


qciaran:
Merging since it is dev merge time

PR #157, opened 2025-07-23
Fixing the artifact generation to use the right Python versions

ciaranra:


qartik:
@ciaranra would it be better to be using abi3 wheels targeting Python 3.10 and above instead of generating wheels for each Python version? That's what we are doing for several internal packages.

ciaranra:
Maybe, not sure what that is. Will look into it, thanks!

ciaranra:
I guess that is a particular binary interface for Python that maybe works for different versions? If so... didn't know you could generate one wheel for PyO3 for, e.g., Python 3.10+, nice

ciaranra:
Alright cool. That works. Thanks for pointing that out @qartik!

PR #166, opened 2025-07-29
Make PCG thread safe

ciaranra:
For the Rust wrapper of PCG, tests of determinism have rare failures that occur when multiple threads are running. This failure can be made much more likely when multiple threads are explicitly spawned. It appears that the C implementation and Rust wrapper are allowing a global state to make it so that threads are not independent/seed deterministic. This PR attempts to fix that. It also moves the C code into the Rust crate for simpler building and to be more idiomatic.


ciaranra:
This PR is effectively supplanted by PR  #167

PR #170, opened 2025-08-07
Handling of Job_shotnum in binary expressions + assignments 

Jonhas-qtm:
Small patch to have PECOS correlate JOB_shotnum variable to shot_id. Fixes issues when using the var for binary expressions and assignment expressions


ciaranra (approved):
LGTM

PR #175, opened 2025-08-14
Naive crosstalk

PabloAndresCQ:
Creating a PR just so that it's easier to track conversations.

### DIFF
--- a/crates/pecos-engines/src/noise/general.rs
+++ b/crates/pecos-engines/src/noise/general.rs
@@ -172,9 +172,9 @@ pub struct GeneralNoiseModel {
     /// qubit states after initialization. Ranges from 0 to 1.
     p_prep_leak_ratio: f64,
 
-    /// Probability of crosstalk during initialization operations
+    /// Probability of crosstalk during preparation operations
     ///
-    /// Models the probability that an initialization operation on one qubit affects nearby qubits.
+    /// Models the probability that a preparation operation on one qubit affects nearby qubits.
     /// In ion trap systems, this could represent scattered light during optical pumping affecting
     /// neighboring ions.
     p_prep_crosstalk: f64,
@@ -326,7 +326,7 @@ pub struct GeneralNoiseModel {
     /// Random number generator for stochastic noise processes
     rng: NoiseRng<ChaCha8Rng>,
 
-    /// Set of qubits that have been initialized at any point in the program.
+    /// Set of qubits that have been prepared at any point in the program.
     ///
     /// This is so that we know which qubits exists and we can apply crosstalk
     /// to them. Qubits that are measured / discarded are not removed from here, since
@@ -335,7 +335,7 @@ pub struct GeneralNoiseModel {
     /// and active qubits should always suffer errors under this naive crosstalk model.
     ///
     /// Using a BTreeSet because we will iterate over the qubits and we want determinism.
-    initialized_qubits: BTreeSet<usize>,
+    prepared_qubits: BTreeSet<usize>,
 
     /// Track which qubits are being measured in the current batch and their gate types
     /// This is needed to properly handle leakage during measurements as well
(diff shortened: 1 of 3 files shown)


crates/pecos-engines/src/noise/general.rs
```diff
@@ -783,6 +797,39 @@ impl GeneralNoiseModel {
...
+                if self.rng.occurs(probability) {
+                    affected_qubits.push(q);
+                }
+            }
+        }
+        // TODO potentially major issue: I expect these will be adding measurement
+        // outcomes to the OutputEngine, which should be ignored. How do we do this?
+        // Perhaps there should be another add_collapse() method that applies a
+        // measurement and discards the outcome.
+        builder.add_measurements(&affected_qubits);
```
PabloAndresCQ, line 839: Comment from Ciaran on the commit:

https://github.com/PECOS-packages/PECOS/commit/dcc5fb73e5bf9611e8981cd89d4e13d376515538#r164014518

> Hmmmm, let's see. Measurements end up kinda of being two rounds... First the noise model sends a measurement request to the simulator and then noise model gets a result from the simulator. Then then noise model can intercept/forward/etc. results to the "classical engine". This is done here:
> 
> [PECOS/crates/pecos-engines/src/noise/general.rs](https://github.com/PECOS-packages/PECOS/blob/55994b88935dab66f22f9221f8fc26f4af39c996/crates/pecos-engines/src/noise/general.rs#L362)
> 
> Line 362 in [55994b8](/PECOS-packages/PECOS/commit/55994b88935dab66f22f9221f8fc26f4af39c996)
> 
>      fn continue_processing( 
> I kinda of had to deal with a similar situation of distinguishing measurement results for leakage... and I ended up just storing the measurement type here:
> 
> [PECOS/crates/pecos-engines/src/noise/general.rs](https://github.com/PECOS-packages/PECOS/blob/55994b88935dab66f22f9221f8fc26f4af39c996/crates/pecos-engines/src/noise/general.rs#L332)
> 
> Line 332 in [55994b8](/PECOS-packages/PECOS/commit/55994b88935dab66f22f9221f8fc26f4af39c996)
> 
>      measured_qubits: Vec<(usize, bool)>, 
> But I feel like what needs to be done is to add some tagging mechanism in ByteMessage
>
> I might... try adding result tagging to ByteMessage...

crates/pecos-engines/src/noise/general/default.rs
```diff
@@ -100,6 +100,7 @@ impl Default for GeneralNoiseModel {
             p2_idle: 0.0,
             leaked_qubits: HashSet::new(),
             rng: NoiseRng::default(),
+            initialized_qubits: BTreeSet::new(),
```
PabloAndresCQ: Comment from Ciaran in the commit:

https://github.com/PECOS-packages/PECOS/commit/dcc5fb73e5bf9611e8981cd89d4e13d376515538#commitcomment-164014201

> Would prefer initialized -> prep or prepared or something
> 
> Note: crosstalk for measurement and prep/reset/init can look differently. But can change that later.

PabloAndresCQ: Sure, happy to change that.

crates/pecos-engines/src/noise/general.rs
PabloAndresCQ: Ok, I think I understand. Without tagging in ByteMessage, what I could do is add an extra bool to `measured_qubits`, so that's `Vec<(usize, bool, bool)>` where the second bool corresponds to `is_crosstalk` to indicate its origin.

For the naive crosstalk, I'd just throw away the measurement outcome at `apply_noise_on_continue_processing` [here](https://github.com/PECOS-packages/PECOS/blob/55994b88935dab66f22f9221f8fc26f4af39c996/crates/pecos-engines/src/noise/general.rs#L594) whenever `is_crosstalk` is True. In future versions, it could be used to make a decision on how to apply crosstalk.

That makes sense, I can do it tomorrow morning. I guess there's no need for tags for the first version, but if you get them in, I'll use them.

ciaranra: Yeah, sounds good. Later I can add a tag system and replace that.

ciaranra (approved):
LGTM

PR #182, opened 2025-08-24
convert QuantumCircuit/Stim to/from SLR

ciaranra:


ciaranra:
This is being handled by PR #204 now

PR #191, opened 2025-10-30
feat: Crosstalk Payload

PabloAndresCQ:
## Description

- Introduces new gatetypes `MeasCrosstalkLocalPayload` and `MeasCrosstalkGlobalPayload` to communicate details from runtime necessary to model crosstalk more realistically.
- Introduces noise channels for local and global crosstalk using the data from these payloads.
- The parameter `p_meas_crosstalk` has been replaced by `p_meas_crosstalk_global` and `p_meas_crosstalk_local`. The method `with_p_meas_crosstalk` from the builder now sets both of these parameters to the given value. There are also individual `with_*` methods for global and local.
- There is a new parameter `p_meas_crosstalk_model` for the transition probabilities after crosstalk-induced Z projection on the victim qubits.
- The specs for global and local crosstalk are defined here: https://github.com/quantinuum-dev/selene-pecos-plugin/issues/66 and https://github.com/quantinuum-dev/selene-pecos-plugin/issues/67.

## Discussion

- With `p_meas_crosstalk_model = {"0->0": 1.0, "1->1: 0.0"}` and using `with_p_meas_crosstalk`, the model reproduces the previous implementation of the channel... with one important caveat: PECOS does not itself introduce `MeasCrosstalk*Payload` instructions which are the ones that trigger the crosstalk channels.
  - A simple workaround for PECOS to be self-sufficient is to have an option `explicit_meas_crosstalk` (that Selene always sets to True) where, if False, causes PECOS to assume it won't receive `MeasCrosstalk*Payload` operations and instead needs to call `apply_crosstalk_faults_from_payload()` when encountering `Measure` or `MeasureLeaked`.

### DIFF
--- a/crates/pecos-engines/src/noise/general.rs
+++ b/crates/pecos-engines/src/noise/general.rs
@@ -309,7 +309,17 @@ pub struct GeneralNoiseModel {
     /// Models the probability that a measurement operation on one qubit affects nearby qubits. In
     /// ion trap systems, this could represent scattered light during fluorescence detection
     /// affecting neighboring ions.
-    p_meas_crosstalk: f64,
+    /// Further details on how crosstalk is modeled depends on information from
+    /// the device runtime, which are provided as MeasCrosstalkGlobalPayload instructions.
+    p_meas_crosstalk_global: f64,
+
+    /// Probability of crosstalk during measurement operations on local qubits
+    ///
+    /// See doc for p_meas_crosstalk_global. The intended distinction is that this
+    /// parameter applies to only qubits that are close to the measured qubit.
+    /// /// Further details on how crosstalk is modeled depends on information from
+    /// the device runtime, which are provided as MeasCrosstalkLocalPayload instructions.
+    p_meas_crosstalk_local: f64,
 
     // --- internally used variables --- //
     /// The maximum of `p_meas_0` and `p_meas_1`
@@ -520,7 +530,7 @@ impl GeneralNoiseModel {
                         self.prepared_qubits.insert(usize::from(q));
                     }
                     self.apply_prep_faults(&gate, &mut builder);
-                    self.apply_crosstalk_faults(&gate, self.p_prep_crosstalk, &mut builder);
+                    self.apply_simple_crosstalk_faults(&gate, self.p_prep_crosstalk, &mut builder);
                 }
                 GateType::Measure | GateType::MeasureLeaked => {
                     // Track which qubits are being measured for leakage handling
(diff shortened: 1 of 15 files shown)


crates/pecos-engines/src/noise/general.rs
```diff
@@ -340,9 +355,10 @@ pub struct GeneralNoiseModel {
     /// Track which qubits are being measured in the current batch and their gate types
     /// This is needed to properly handle leakage during measurements as well
     /// as crosstalk.
-    /// TODO: manage this via result tags.
-    /// Each entry is (`qubit_id`, `is_measure_leaked`, `is_crosstalk`)
-    measured_qubits: Vec<(usize, bool, bool)>,
+    measured_qubits: Vec<(usize, GateType)>,
+
+    /// Stored outcome builder
+    results_builder: ByteMessageBuilder,
```
PabloAndresCQ, line 361: Keeps track of the outcomes that need to be sent back once `continue_processing` returns `Complete(O)`. This is so that we can have multiple rounds of `NeedsProcessing(I)` and accummulate the outcomes until we are ready to return them.

crates/pecos-engines/src/noise/general.rs
```diff
@@ -495,6 +520,15 @@ impl GeneralNoiseModel {
...
 
         for gate in gates {
+            // Track which qubits are being measured for leakage handling
+            if matches!(gate.gate_type, GateType::Measure | GateType::MeasureLeaked) {
+                self.measured_qubits.extend(
+                    gate.qubits
+                        .iter()
+                        .map(|q| (usize::from(*q), gate.gate_type)),
+                );
+            }
```
PabloAndresCQ, line 530: Moved this out because it's necessary to keep track of `measured_qubits` even if `Measure` and/or `MeasureLeaked` are contained in `noiseless_gates`.

crates/pecos-engines/src/noise/general.rs
```diff
@@ -520,20 +554,41 @@ impl GeneralNoiseModel {
                         self.prepared_qubits.insert(usize::from(q));
                     }
                     self.apply_prep_faults(&gate, &mut builder);
-                    self.apply_crosstalk_faults(&gate, self.p_prep_crosstalk, &mut builder);
+                    self.apply_simple_crosstalk_faults(&gate, self.p_prep_crosstalk, &mut builder);
```
PabloAndresCQ, line 557: Prep still uses the old crosstalk channel. In Selene we currently set `p_prep_crosstalk = 0.0`.

crates/pecos-engines/src/noise/general.rs
```diff
@@ -596,69 +654,101 @@ impl GeneralNoiseModel {
...
-            results_builder.add_outcomes(&[val as usize]);
         }
-
-        // Clear the measured qubits for the next batch
         self.measured_qubits.clear();
-
-        // Build and return the biased measurement results
-        Ok(results_builder.build())
+        self.results_builder.add_outcomes(&outcomes);
+        Ok(ops_builder.build())
```
PabloAndresCQ, line 751: `apply_noise_on_continue_processing` now always returns a ByteMessage of quantum_ops and appends the outcomes to the new field `self.results_builder`. It's `continue_processing` the one that, if this ByteMessage is empty, flushes the `results_builder` and sends it via `Complete(O)`.

PabloAndresCQ (commented):
Ready for review. Some comments below to give context to the changes

ciaranra:
Nice! LGTM. Thanks for working on this!

PR #267, opened 2026-03-12
Changing error message for RNGIndex for more clarity

Jonhas-qtm:
Error message for RNGIndex doesn't provide enough information to the enduser on what caused the error. This patch fixes that.


ciaranra (approved):
Looks good

PR #330, opened 2026-06-24
fix: add RNGadvance support to legacy Python CVM

qartik:
## Summary
- add `RNGadvance(delta)` support to the legacy Python CVM RNG path, including backward replay through historical bound sequences
- preserve shot-local rewind correctness by resetting replay state at shot boundaries instead of mutating `count` directly
- keep existing `RNGseed`, `RNGindex`, and `RNGnum` behavior while tightening RNG argument validation and stream bookkeeping
- include the CI hardening and Julia formatter-only updates already present on this branch

## Reviewer Notes
- The behavior change is in `python/quantum-pecos/src/pecos/engines/cvm/rng_model.py` and the legacy caller in `python/quantum-pecos/src/pecos/engines/hybrid_engine_old.py`.
- Backward `RNGadvance` now replays from an explicit replay base: reseeds reset that base, and each shot snapshots the current RNG state so rewinds stay correct even when the stream continues across shots.
- Bound history is stored in run-length encoded form to reduce memory growth for long runs with stable bounds.
- The Rust binding adds `RngPcg.clone()` only to snapshot the current generator state for shot-local replay.
- Workflow and Julia file changes are CI fixes, not product behavior changes.

## Validation
- `uv run maturin develop --manifest-path python/pecos-rslib/Cargo.toml`
- `uv run --frozen pytest python/quantum-pecos/tests/pecos/unit/test_rng.py`
- `uv run --frozen ruff check python/quantum-pecos/src/pecos/engines/cvm/rng_model.py python/quantum-pecos/src/pecos/engines/hybrid_engine_old.py python/quantum-pecos/tests/pecos/unit/test_rng.py`

### DIFF
--- a/.github/workflows/julia-release.yml
+++ b/.github/workflows/julia-release.yml
@@ -16,6 +16,9 @@ permissions:
 
 env:
   TRIGGER_ON_PR_PUSH: true  # Set to true to enable triggers on PR pushes
+  CARGO_HTTP_MULTIPLEXING: "false"
+  CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+  CARGO_NET_RETRY: "10"
 
 on:
   push:
--- a/.github/workflows/python-release.yml
+++ b/.github/workflows/python-release.yml
@@ -5,6 +5,9 @@ permissions:
 
 env:
   TRIGGER_ON_PR_PUSH: true  # Set to true to enable triggers on PR pushes
+  CARGO_HTTP_MULTIPLEXING: "false"
+  CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+  CARGO_NET_RETRY: "10"
   # Must match pecos_build::cmake::CMAKE_VERSION. The MWPF decoder feature
   # uses highs-sys, which needs cmake; the wheel build installs this version
   # via `pecos install cmake` and points highs-sys's cmake-rs at it via $CMAKE.
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -9,6 +9,7 @@ env:
   RUST_BACKTRACE: 1
   PYTHONUTF8: 1
   CARGO_HTTP_MULTIPLEXING: "false"
+  CARGO_NET_GIT_FETCH_WITH_CLI: "true"
   CARGO_NET_RETRY: "10"
   LLVM_VERSION: "14.0.6"
   # Force the MWPF decoder feature on in CI so we exercise the cmake-dependent
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -8,6 +8,8 @@ env:
   RUSTFLAGS: -C debuginfo=0
   RUST_BACKTRACE: 1
   PYTHONUTF8: 1
+  CARGO_HTTP_MULTIPLEXING: "false"
+  CARGO_NET_RETRY: "10"
   LLVM_VERSION: "14.0.6"
   # Force the MWPF decoder feature on in CI so we exercise the cmake-dependent
   # build path and catch regressions. GitHub-hosted runners ship cmake.
@@ -109,11 +111,6 @@ jobs:
         rustup default stable
         "$HOME\.cargo\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
         $env:Path += ";$HOME\.cargo\bin"
-        # Cargo's default HTTP/2 multiplexing has been flaky on GitHub's
-        # Windows runners while talking to crates.io during the first
-        # dependency resolution in `just ci-env`.
-        "CARGO_HTTP_MULTIPLEXING=false" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-        "CARGO_NET_RETRY=10" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         rustup show
 
     - name: Set up Rust (Unix)
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -109,6 +109,11 @@ jobs:
         rustup default stable
         "$HOME\.cargo\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
         $env:Path += ";$HOME\.cargo\bin"
+        # Cargo's default HTTP/2 multiplexing has been flaky on GitHub's
+        # Windows runners while talking to crates.io during the first
+        # dependency resolution in `just ci-env`.
+        "CARGO_HTTP_MULTIPLEXING=false" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        "CARGO_NET_RETRY=10" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         rustup show
 
     - name: Set up Rust (Unix)
--- a/julia/PECOS.jl/src/Decoder.jl
+++ b/julia/PECOS.jl/src/Decoder.jl
@@ -44,11 +44,17 @@ export AbstractDecoder, DecodingResult, decode, check_count, bit_count
 Result of a decoding operation.
 """
 struct DecodingResult
-    """Decoded observable/correction vector."""
+    """
+    Decoded observable/correction vector.
+    """
     observable::Vector{UInt8}
-    """Weight/cost of the solution."""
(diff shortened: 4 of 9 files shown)


ciaranra:
Claude's review:

Reviewed the RNG changes. The Python `RNGModel` logic is sound — the reseed-and-replay rewind is correct precisely because `pcg32_boundedrand_r` uses rejection sampling (variable underlying steps), so replaying the historical bound sequence is necessary; tests pass; it also fixes a latent `self.seed = None` init bug.

One scoping note for follow-up: `RNGadvance` lands only on the legacy Python cvm `RNGModel` (reached via `hybrid_engine_old.py`). The newer workflows — `hybrid_engine.py` (RustPhirClassicalInterpreter), `sim()` (pecos-qasm / phir-json), and `sim_neo()` — don't route program RNG through it, and the Rust `RNGModel` has no advance. Filed #332 to track landing program RNG once in the canonical Rust `RNGModel` that the pecos-neo path will consume, rather than back-porting into the engines slated for retirement.

Minor: in `set_relative_index`, the rewind loop's `if remaining == 0: break` should be `<= 0` — when the target lands mid-run, `remaining` goes negative and the loop iterates the rest of the bound history doing empty draws (harmless today, just wasteful).

ciaranra:
## Consolidated review (Claude + Codex)

Two-model review of the RNG changes. Net: the core algorithm is sound, but the
PR introduces a real correctness bug on its one **live** path, and the scope is
narrower than the title. Recommend merge as a scoped legacy-path fix after two
fixes + a rename.

### What's correct
- The reseed-and-replay rewind is the right mechanism for this generator:
  `pcg32_boundedrand_r` uses rejection sampling (`crates/pecos-random/src/rng_pcg.rs:58-66`),
  so a bounded draw consumes a variable number of underlying PCG steps and there
  is no jump-ahead — replaying the exact historical bound sequence is necessary.
  The run-length `_draw_bound_runs` design holds up: adversarial sequences
  (bound changes, high-rejection bounds, `RNGindex`, mid-run rewinds,
  run-boundary rewinds, rewind-to-0, reseed, rewind-then-forward, plus 100
  randomized op sequences) all matched a fresh seeded PCG replay.
- Also fixes a latent init bug (`self.seed = self.set_seed(...)` previously
  stored `None`).

### Blocking: per-shot `count` reset desyncs the replay state (live default engine)
`hybrid_engine_old.py:144` resets `self.rng_model.count = 0` at each shot start
without clearing `_draw_bound_runs` or reseeding. Before this PR `count` was a
bare index counter; this PR **couples** it to `_draw_bound_runs` and the replay
base. So a multi-shot program that calls `RNGadvance(-k)` without an in-shot
`RNGseed` rewinds against stale shot-1 bound history and the seed-reset PCG, and
diverges.

This is not a dead path: `pecos/__init__.py:269` imports the top-level public
`pecos.HybridEngine` from `hybrid_engine_old`, so this is the default engine.

Repro — invariant: draw forward, `advance(-1)`, redraw should reproduce the
value just drawn. Verified independently by both reviewers:
```
# clean public API (no legacy reset): round-trips correctly
control: advance(-1) then redraw == last draw  -> True

# with the legacy per-shot `count = 0` (shot >= 2, no in-shot RNGseed):
shot2 forward draws : 706518008 257766219 1139787578   (count -> 3)
after advance(-1)   : 1868747362
matches last draw?  : False   # reconstructs from seed (PCG idx 2), not the
                              # continued stream the forward draws used (idx 7)
```
Fix: add a public shot-reset / replay-base API and have the engine call it
instead of mutating `count` directly — choose one explicit semantic (reseed +
clear history per shot, OR snapshot a per-shot replay base). Add a regression
test for multi-shot `RNGadvance` without an in-shot `RNGseed`.

### Minor: rewind loop early-exit
`set_relative_index` subtracts the full `run_length` (rng_model.py:109) while
breaking only on `remaining == 0` (line 104), so a mid-run target drives
`remaining` negative and the loop iterates the rest of the history doing empty
draws. Harmless today (empty `range`, `current_bound` restored), but should be
`to_replay = min(run_length, remaining); remaining -= to_replay; break on <= 0`.

### Scope / cross-engine
- The live `sim()` (QASM → Rust `pecos-qasm`, `PLATFORM_FUNCTIONS` = seed/bound/
  index/num, no advance) and `sim_neo()` paths do **not** route through this
  Python CVM `RNGModel`. So this is not general "RNGadvance support to PECOS" —
  it's the legacy Python CVM path only. Suggest renaming the PR accordingly.
- Cross-engine drift: this PR makes Python `set_seed` reset `count` + clear
  history (correct for a platform `RNGseed`), but Rust `RNGModel::set_seed`
  (`rng_pcg.rs:311-313`) does not reset `count`. Track parity + the canonical
  program-RNG home in #332.
- (FYI) HUGR has a separate `tket.qsystem.random` mechanism that already
  includes `RandomAdvance` (`crates/pecos-hugr/src/engine/handlers/qsystem.rs:288`);
  unrelated to this OpenQASM-style platform RNG.

### Recommendation
Merge as contained legacy Python CVM `RNGadvance` support after: (1) fixing the
per-shot reset invariant, (2) the `<= 0` loop fix, (3) renaming/rewording the PR
to the legacy scope. Track Rust `set_seed` parity + real neo program-RNG in #332.

### Corrections to an earlier review comment on this PR
The earlier comment said `hybrid_engine.py` uses `RustPhirClassicalInterpreter`
and framed the RNG path as a dead legacy one. Both are off: `hybrid_engine.py`
defaults to the Python `PhirClassicalInterpreter` (Rust is opt-in via
`cinterp="rust"`), and the RNGadvance path is the default public
`pecos.HybridEngine`, not dead — which is exactly why the Area 1 bug matters.

---
Reviewed by Claude Opus 4.8 and Codex (adversarial cross-review).

qartik:
@ciaranra could you review again?

ciaranra:
## Sign-off: fix verified, mergeable (Claude + Codex)

Re-reviewed `c1a612bc5` ("preserve shot-local RNG rewind state"). Two-model
adversarial pass; both converge: **mergeable as the scoped legacy Python-CVM
`RNGadvance` fix.** Thanks for the quick, well-targeted turnaround.

### The round-1 blocker is closed
The per-shot `count = 0` reset on the default public `pecos.HybridEngine` no
longer desyncs the rewind: `start_shot()` now resets `count` + `_draw_bound_runs`
and captures a `_replay_base_pcg` clone **at the current stream position**, and
backward `advance` replays from that clone instead of reseeding from `self.seed`.
Verified on the built tip — the exact failing case now round-trips:

```
start_shot(1); draw,draw,draw -> [..., 1139787578]
advance(-1); draw            -> 1139787578   (was 1868747362 pre-fix)
```

A bound-change-within-shot rewind also round-trips. The Area 2 loop fix
(`remaining <= 0` + `remaining -= replay_count`) is exactly right, and the PR
rename to the legacy-CVM scope matches reality.

### What was checked
- Full RNG unit suite: **15 passed** (incl. the new multi-shot test).
- Replay-base capture points (mid-shot reseed, bound-before-shot, `set_index`
  after `start_shot`), clone state-independence through PyO3, single-shot
  equivalence to the old `set_seed(self.seed)` path, loop arithmetic
  (exact-boundary / mid-run / rewind-to-zero / forward-after-rewind).
- A randomized oracle harness (200 sequences × 180 ops, bounds incl. 0, 65537,
  2147483649) comparing post-rewind state to an independent PCG replay — passed.
- Blast radius: the only other `rng_model.*` access is a `shot_id` read; the
  multiprocessing legacy path also routes through `eng.run()` → `start_shot()`.

### Non-blocking follow-ups
- Optional hardening tests worth adding: one combining `set_bound` +
  `start_shot` + `set_index` + backward `advance` in a single case, and a
  no-`start_shot` property test locking the single-shot non-regression. A
  one-line doc on the shot-local contract (rewind stays within the current
  shot/seed segment; `current_bound` carries across shots) would help too.
- Rust `RNGModel::set_seed` count-reset parity and real program-RNG for the
  surviving `sim()` / `sim_neo()` paths remain tracked in #332 — out of scope
  for this legacy fix.

LGTM.

---
Reviewed by Claude Opus 4.8 and Codex (gpt-5.5) — adversarial cross-review.

ciaranra (approved):
Looks good. Thanks!

qartik:
@ciaranra I can't merge, I am seeing:

> Commits must have verified signatures.

ciaranra:
Huh, weird... Maybe it is some added security I added... I'll try merging

PR #335, opened 2026-07-02
perf: remove coordinator result lock contention

qartik:
## Summary

- remove the shared `Arc<Mutex<Vec<_>>>` result collector and final global sort from `ParallelCoordinator::run()`
- collect each worker output locally with Rayon and flatten in worker order after the parallel phase
- add a regression test that locks in deterministic worker-order result aggregation

## Why this change

The previous implementation forced every worker to contend on one mutex while pushing results, then paid an extra global sort to restore deterministic ordering. This refactor keeps the ordering guarantee while removing both sources of overhead from the hot parallel coordinator path.

## Validation

- `cargo test -p pecos-neo coordinator --quiet`
- `cargo test -p pecos-neo --quiet`

## Performance metrics (before/after)

Benchmark command:

`cargo bench -p benchmarks --bench benchmarks -- 'ParallelCoordinator/'`

Before = `origin/dev` (`602ebd0`)  
After = PR head (`f300415`)

| Benchmark | Before (time) | After (time) | Delta (time) | Delta (throughput) |
|---|---:|---:|---:|---:|
| ParallelCoordinator / 100 shots | 163.56 µs | 158.60 µs | -4.84% | +5.08% |
| ParallelCoordinator / 1,000 shots | 1.0301 ms | 1.0097 ms | -5.28% | +5.57% |
| ParallelCoordinator / 10,000 shots | 10.647 ms | 10.289 ms | -3.36% | +3.48% |

Criterion reported "Performance has improved." for all three shot sizes in the after run.


ciaranra:
Looks good to me. (Straight-foward improvement.) Will wait for CI and do a quick review.

ciaranra (approved):
Looks good. Thanks!

ciaranra:
I am guessing I might still need to merge due to the "verified" thing which I probably need to look at turning off...

ciaranra:
I'll go ahead and merge :D

PR #338, opened 2026-07-07
Pecos dev merge

ciaranra:


ciaranra:
Merged origin/dev and fixed the pre-existing compile breakage properly. Root cause: the branch's decoder files (decoder-core, uf-decoder, fault_tolerance_bindings) were **stale pre-squash copies** — this lineage's work reached dev via the #322 squash and then evolved further there (wide-observables ObsMask API), but the earlier "Fix PECOS dev merge integration issues" resolution kept the branch's older vintage, reverting dev's API in those regions.

Reconciliation (content-verified, not guessed): every distinctive symbol of the branch's decoder work (`full_fallback`, `effective_windowing`, `parse_dem_metadata` hardening, windowed strategy, boundary-gate guards) already exists on dev at equal or newer form, so dev's versions were restored wholesale for the decoder lineage; the branch's genuinely original work (SZZ round-order gating, hugr-qis compiler, DEM-builder types, engines, examples, python tests) is untouched.

Verified: workspace compiles; 180 decoder-core + uf-decoder tests pass; clippy cold-clean on the three restored crates; fmt clean.

ciaranra:
Adversarial semantic-preservation review (Codex) of the four lint/merge-fix commits (`97e8862f9..660cfe21d`): **all high-risk items verified behavior-preserved** — the ZZ/RZZ match-arm merge (history commit `eda0cd8fb` confirms theta-ignoring was pre-existing intent), the deleted arm (provably unreachable, no side effects), the DEM-builder by-ref conversions, RenderPolicies grouping, the hoisted solver (guard still precedes the only call), the FFI vtable field renames (field order verified against Selene's actual `core_types.h`), the bitcode NUL strip (exactly one byte), and the exact-zero float asserts (negative zero/denormals not producible in those configurations).

**Three deliberate behavior changes, accepted and documented for this branch's eventual full review:**
1. `qec/surface/decode.py`: `szz_runtime_barriers` is now keyword-only in two functions (FBT002) — all 32 in-tree callers already use keywords; external positional callers (none known; unreleased branch API) would need `szz_runtime_barriers=`.
2. `examples/surface/dem_decomposition_diagnostics.py`: invalid cached-diagnostics JSON now raises `TypeError` (it is a type check) instead of `ValueError`; no caller catches either.
3. `selene_runtime.rs`: a >u32::MAX plugin argument count now fails loud instead of silently wrapping before the FFI call — the wrap was a latent bug.

PR #348, opened 2026-07-10
Default HybridEngine to the Rust PHIR interpreter

ciaranra:
Default `HybridEngine`'s classical interpreter to the Rust implementation (`RustPhirClassicalInterpreter`), and strengthen the Rust↔Python parity coverage that backs the switch.

## Change
`_make_classical_interpreter(None)` now returns the Rust interpreter instead of the Python one. The `cinterp` argument still lets callers choose: `cinterp="python"` for the pure-Python `PhirClassicalInterpreter`, `cinterp="rust"` (or the default) for the Rust-backed one, or a concrete instance. Rust is a core dependency (`pecos_rslib` always ships), so the default path is always available; no fallback added.

Motivation: the Rust interpreter is bit-for-bit equivalent to Python (parity suite) and materially faster.

## Verification
- Default resolves to `RustPhirClassicalInterpreter`.
- **quantum-pecos suite: 784 passed**, **slr_tests: 435 passed** (full SLR→PHIR→HybridEngine path) with the Rust default — no feature gaps.
- Parity coverage strengthened (second commit): fuzz iterations 200→2000 (classical) / 200→1000 (quantum); classical fuzz now nests expressions and includes unary `~`; added the `example1` WASM program; added a QASM→PHIR generated-differential over the curated converter fixtures. ~3,200 direct Rust≡Python comparisons total, all green, in <3s.

## Notes
- Users can opt back to Python via `cinterp="python"`.
- This is a default-behavior change affecting all `HybridEngine` users.

### DIFF
--- a/python/quantum-pecos/src/pecos/engines/hybrid_engine.py
+++ b/python/quantum-pecos/src/pecos/engines/hybrid_engine.py
@@ -36,16 +36,16 @@ def _make_classical_interpreter(
     """Create a classical interpreter from a string specifier.
 
     Args:
-        spec: One of "python", "rust", or None (defaults to "python").
+        spec: One of "python", "rust", or None (defaults to "rust").
 
     Returns:
         A classical interpreter instance.
     """
-    if spec == "rust":
+    if spec is None or spec == "rust":
         from pecos_rslib import RustPhirClassicalInterpreter  # noqa: PLC0415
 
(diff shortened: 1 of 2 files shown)


ciaranra:
Consolidated into #347 (the default-to-Rust switch + parity strengthening now live there); avoiding a separate PR.

PR #351, opened 2026-07-11
Bump Python packages to 0.9.0.dev0

ciaranra:
Coordinated Python version bump `0.8.0.dev8` → `0.9.0.dev0`, prepping the release that flips `HybridEngine`'s default classical interpreter to Rust (#347).

**Why a minor bump:** the default-interpreter switch is a user-visible default-behavior change; in the 0.x convention the minor number is the signal for that (additive-only would be a patch/dev increment). It gives the release note — and the `cinterp="python"` escape hatch — a clear version boundary.

**What's bumped (every occurrence, verified zero stragglers):**
- `version =` in all 11 workspace `pyproject.toml`s (root meta-package, pecos-rslib, -llvm, -cuda, -exp, quantum-pecos, 5 selene plugins)
- Exact-version internal pins: root `quantum-pecos[cuda12/13]==...` (4) and quantum-pecos's `pecos-rslib==...` / `pecos-rslib-llvm==...`
- `uv.lock` + `exp/zluppy/uv.lock` regenerated — diffs are version-lines-only (11 and 3 respectively, no dependency churn)
- `docs/user-guide/cli.md` example output

**Verification:** `git grep 0.8.0.dev8` → no matches; `uv lock --check` passes; CI-standard sync rebuilds cleanly and `pecos.__version__` reports `0.9.0.dev0` with the Rust default interpreter intact.

After merge: tag `py-0.9.0.dev0` (or at cut time) to build the tagged artifact bundle; publishing remains manual from the collected artifacts.

### DIFF
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -132,7 +132,7 @@ LLVM 21.1:
 
 Python:
   [OK] uv: uv 0.11.3
-  [OK] import pecos: v0.8.0.dev8
+  [OK] import pecos: v0.9.0.dev0
   [OK] pecos_rslib: v0.2.0-dev.0
 
 CUDA (optional):
--- a/exp/zluppy/uv.lock
+++ b/exp/zluppy/uv.lock
@@ -539,7 +539,7 @@ wheels = [
 
 [[package]]
 name = "pecos-rslib"
-version = "0.8.0.dev8"
+version = "0.9.0.dev0"
 source = { editable = "../../python/pecos-rslib" }
 
 [package.metadata]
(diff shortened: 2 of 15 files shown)


ciaranra:
Also updated `scripts/publish-wheels.sh` to publish `pecos-rslib-llvm` (before `quantum-pecos`, which pins it exactly). The script predated the LLVM split — it would have silently skipped the llvm wheels, leaving `quantum-pecos==0.9.0.dev0` uninstallable from PyPI (`pecos-rslib-llvm` has never been published; note the wheel bundle dir name is `dist/pecos-rslib-llvm/`, which the script already resolves).

PR #352, opened 2026-07-11
Fix silent failures in publish-wheels.sh (masked twine errors, skipped confirmations)

ciaranra:
Two robustness bugs in `scripts/publish-wheels.sh`, both of which bit during the 0.9.0.dev0 publish:

1. **Error-masked twine check.** The check piped twine's output through `grep -v license-file`, so the `if` tested *grep's* exit status — a failed or even *missing* twine ("Failed to spawn: twine") still printed "Distribution checks passed". Now the output is captured once and twine's own exit status decides (twine exits nonzero on real errors, zero on the benign maturin license-file warning, which is still filtered from the display).

2. **Piped confirmation skipped a package.** The per-package prompt used `read -n 1`, so piping `yes` fed `y` to the first prompt and the leftover `\n` to the second — during the 0.9.0.dev0 publish this silently skipped `pecos-rslib-llvm`, briefly leaving `quantum-pecos` on PyPI with an unresolvable exact pin (caught and uploaded directly). Fixed twofold: the prompt now reads a full line, and a `-y/--yes` flag supports non-interactive use without piping tricks.

**Verification:** `bash -n` clean; side-by-side read semantics demonstrate the old bug and the fix (`old read 2: []` vs `new read 2: [y]` under piped `yes`); `--dry-run` against the real 0.9.0.dev0 distribution bundle passes all three package checks with the new logic.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -66,8 +66,13 @@ installed, e.g. `uv tool install twine`).
 ```
 
 Confirm each package at its prompt. Upload order matters and the script
-handles it: `pecos-rslib` -> `pecos-rslib-llvm` -> `quantum-pecos`
-(quantum-pecos pins both at exact versions, so dependencies go first).
+enforces it: all packages are preflighted (complete set, consistent version,
+only expected files) before anything uploads, dependencies go first
+(`pecos-rslib` -> `pecos-rslib-llvm` -> `quantum-pecos`, which pins both at
+exact versions), and any failure or declined prompt aborts the remaining
+uploads rather than continuing. Publishing `quantum-pecos` alone (`-p`) is
+blocked unless its pinned dependencies already exist on PyPI. Checks run
+`twine check --strict`.
 Credentials come from `~/.pypirc`; a **new** package's first upload needs an
 account-scoped token (project-scoped tokens cannot create projects).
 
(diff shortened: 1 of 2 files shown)


ciaranra:
Added `RELEASING.md` to this PR: the full manual release sequence as exercised for 0.9.0.dev0 (coordinated version bump -> post-merge green -> tag -> tag-run bundle -> dry-run -> manual per-package publish -> GitHub release -> fresh-venv verification from PyPI). The publish step stays deliberately manual with a human confirming each upload.

ciaranra:
Codex adversarial review (single-arm, right-sized for the diff) found the warn-and-continue paths could recreate the partial-publish incident: missing/empty package dirs and declined prompts returned success, and `-p quantum-pecos` bypassed dependency ordering. Reworked and verified with a stub-twine harness:

- **All-packages mode preflights everything before uploading anything**: complete package set, only expected `<pkg>-*.whl/.tar.gz` files, one consistent version across the set — any violation is fatal (exit 1) before the first upload.
- **Any upload failure or declined prompt aborts the remaining packages** (verified: rslib uploads, llvm upload fails → quantum-pecos never attempted).
- **`-p quantum-pecos` is guarded**: its exact `pecos-rslib`/`-llvm` pins must already exist on PyPI.
- `twine check --strict` (current wheels pass strict cleanly, so the license-file exception is gone — any warning is now a real signal); quoted `trap 'rm -rf -- "${TEMP_DIR:?}"'`; TEMP_DIR only created for zip input; non-TTY stdin without `--yes` refused upfront; EOF at a prompt is a loud fatal, not a skip.
- RELEASING.md updated to describe the enforced semantics.

PR #378, opened 2026-07-27
`SeedReport`: Saving & Replaying RNG seed capability.

drewconnelly-qntm:
## Changes:
In `crates/pecos-engines/src/monte_carlo/engine.rs`:
- Introduces `SeedReport` which stores the RNG seed information for a monte carlo run
- Introduces `run_with_workers_report_seeds` which runs a monte carlo job with workers, while managing the RNG seeds with a seed report. It returns a tuple with `(ShotVec, SeedReport)`.
- Introduces `run_with_workers_from_seed_report` which replays a run from the given seed report which stores all the seed and worker information required to replay the shot. This returns a `ShotVec`.
- Introduces `WorkerSeedRecord` which stores the worker index, num shots, and seed for that worker, which is what is held in a vector inside the overall job's `SeedReport` for each monte carlo job.
- Introduces methods for saving a `SeedReport` to JSON as well as loading them from JSON. (`SeedReport::to_json_file` and `SeedReport::from_json_file`)

## Potential Future Work:
- Lets work start up on Error Pruning, Visualizing now that we can save monte carlo runs and rerun them.
- Opens the possibility for single-shot replaying, so the user can identify a particular shot (or maybe all shots with logical failures) and rerun that exact shot. Right now only rerunning the entire monte carlo job is permitted.

## Testing:
`cargo test -p pecos-engines` was all good
`just build; just test` ran through the rust tests fine but had issues with the python pytests due to some bugs with my environment locally.


qciaran:
I haven't looked closely but just skimming things: You might want to see if you can do this without using an Arc Mutex. They are slightly expensive.

drewconnelly-qntm:
> I haven't looked closely but just skimming things: You might want to see if you can do this without using an Arc Mutex. They are slightly expensive.

Thanks, I'll check on how necessary that is.

ciaranra:
## Review

Nice feature and the right general shape: the report is the single source of truth for worker seeds, the JSON round-trip is tested, and error paths map to `PecosError::Input`. `run_with_workers()` delegating to the new method (rather than keeping a third copy of the run loop) is also good. I ran the PR's tests (7/7 pass) and clippy (clean) locally. That said, there are two blocking issues and a few smaller ones.

### 1. The replay tests are vacuous (blocking)

The two headline tests — `rerun_from_seed_report_reproduces_original_results` and `rerun_from_seed_report_loaded_from_json_reproduces_original_results` — currently prove nothing. Verified by experiment: running the fixture with seed 42 vs seed 999999 produces **identical results**. `ExternalClassicalEngine` always returns `result: 0` and `new_with_defaults` uses pass-through noise, so shot output does not depend on seeds at all. The reproduce assertions would still pass if all seeding logic were deleted.

Shot-level determinism is untested anywhere: `run_with_seed_report_is_deterministic_for_same_seed_workers_and_shots` only compares report metadata, and its different-seed engine (`c`) only asserts the *seeds* differ, not the results.

Fix: use a fixture whose output actually depends on the RNG (e.g. `new_with_depolarizing_noise` with a circuit that measures), then mutation-check the tests: a different seed must produce different shots, and replaying the recorded seeds must produce equal shots.

### 2. ~110 lines duplicated verbatim (blocking)

`rerun_from_seed_report` is a byte-for-byte copy of the execution body of `run_with_workers_seed_report` — worker-engine setup, dedicated thread pool, panic-catching shot loop, result sorting — differing only in where the seeds come from. Any future fix to the pool/panic/ordering logic now has to be made twice, and the copies will drift. Suggest factoring one private helper, e.g.:

```rust
fn run_workers_with_seeds(
    &self,
    worker_seeds: &[u64],
    shots_per_worker: &[usize],
) -> Result<ShotVec, PecosError>
```

and having both paths call it.

### 3. Replay half-ignores the loaded report

A `SeedReport` loaded from JSON is a system boundary, and replay handles it loosely:

- `seed_report.workers[worker_idx]` panics if `workers.len() < num_workers`. A truncated or hand-edited file should produce `PecosError::Input`, not a panic.
- The recorded `WorkerSeedRecord.shots` and `worker_idx` fields are never used: replay recomputes `distribute_shots(num_shots, num_workers)` and indexes positionally. A report with reordered workers or edited shot counts replays silently wrong. Either use the recorded values or validate that they match — otherwise they're dead data.

### 4. `self.seed = seed_report.root_seed` leaves the engine inconsistent

`rerun_from_seed_report` assigns the `seed` field directly, but `set_seed()` also reseeds `self.rng` and the `hybrid_engine_template`. After a replay, the engine claims `seed == root_seed` while its RNG stream and template are elsewhere — a subsequent `run_with_workers` call produces a report whose `root_seed` doesn't correspond to its `base_seed`. Either call `self.set_seed(seed_report.root_seed)` or don't mutate engine state at all (replay doesn't need it — the worker seeds come from the report). The assignment also sits before the zero-shot/zero-worker asserts.

### 5. API shape

The `save_seed_report: bool` flag plus the hardcoded `"seed_report.json"` path is a footgun: it silently overwrites in the current working directory, concurrent runs clobber each other, and the caller can't choose a path. Simpler API: drop the bool, always return the report, and give `SeedReport` a `to_json_file(path)` method symmetric with `from_json_file`. That also removes `MonteCarloEngine::save_seed_report_json`, which doesn't touch the engine — the behavior belongs on `SeedReport`.

### Nits

- `crates/pecos-engines/Cargo.toml`: use `tempfile.workspace = true` — the workspace root already defines `tempfile = "3"` and the other crates use the workspace entry.
- `save_seed_report_json` calls `std::fs::write` even though `use std::fs;` is imported.
- Worth a doc note on `SeedReport` that `root_seed` alone cannot reproduce `base_seed` if the engine ran jobs before this one (the RNG has advanced) — the per-worker seeds are the actual replay contract.

ciaranra:
## Review: `SeedReport` — saving & replaying RNG seeds

One blocking issue and several boundary-validation and packaging problems below. Line numbers refer to the PR head.

### Blocking

**1. Every normal run now rewinds the engine RNG (`crates/pecos-engines/src/monte_carlo/engine.rs:388`)**

`run_with_workers` and `run` both funnel through `run_with_workers_report_seeds` → `run_with_workers_from_seed_report`, which unconditionally calls `self.set_seed(seed_report.root_seed)`. Before this PR, `base_seed` advanced via `self.rng.next_u64()` on each run; now it is re-derived from the same root seed every time.

Reproduced against this branch:

```rust
engine.set_seed(42);
let a = engine.run_with_workers(8, 2)?;   // base_seed = 15898102487349570925
let b = engine.run_with_workers(16, 2)?;  // base_seed = 15898102487349570925
// worker-0 shots of `b` are byte-identical to worker-0 shots of `a`
```

`engine.run(20)` twice likewise returns identical `ShotVec`s. The multi-batch usage documented in `crates/pecos/examples/sim_api_final.rs` (`run(100)`, then `run(500)`, then `run_with_workers(1000, 8)` on one engine) therefore accumulates duplicated samples: correlated statistics and underestimated variance, with no error or warning.

Suggested shape: keep the recording path drawing a fresh `base_seed` from `self.rng`, and let the replay path take the base seed from the report without touching engine state.

**2. Replay mutates the caller's engine (`engine.rs:369`)**

`run_with_workers_from_seed_report(&mut self, ...)` overwrites `self.seed`, reseeds `self.rng`, and reseeds `hybrid_engine_template` with the report's `root_seed`, and never restores them. Verified: an engine at `set_seed(999)` that replays a report recorded with `root_seed = 42` has `engine.seed == 42` afterwards.

The realistic path to trouble: a user loads an archived report to inspect a suspicious shot, then continues their sweep on the same engine — every later run reproduces the archived data. `reset()` restores `self.seed`, so it cannot recover the original stream either. Replay should either be `&self` (deriving worker seeds without mutating engine state) or save/restore around the call.

### Boundary validation

A `SeedReport` comes from a JSON file on disk, so it is untrusted input; the replay path currently validates it with panics inside a `Result`-returning API.

- **`engine.rs:414`** — `assert!(seed_report.workers[i].worker_idx == item.0, "..")` and the matching `shots` assert. A schema-valid report `{num_shots: 10, num_workers: 2, workers: [{worker_idx: 0, shots: 7, ...}, {worker_idx: 1, shots: 3, ...}]}` parses fine and then panics here with the message `..`. Across an FFI or server boundary this aborts instead of surfacing a recoverable error. These should return `PecosError::Input` with a message naming the mismatch.
- **`engine.rs:378`** — `assert!(num_shots > 0)` / `assert!(num_workers > 0)` on deserialized values. The too-few-workers check five lines below already has the right shape (`PecosError::Input`); these should match it.
- **`engine.rs:380`** — validation is asymmetric: `workers.len() < num_workers` errors cleanly, but a report with *more* worker records than `num_workers` passes this guard and panics later at line 414. `workers.len() != num_workers` would route both shapes through the same error path.
- **`engine.rs:393`, `:422`** — `num_shots` flows straight into `Vec::<(usize, usize, Shot)>::with_capacity(num_shots)` and `num_workers` into `ThreadPoolBuilder::num_threads`. A corrupted report with `"num_shots": 10000000000000` requests a multi-terabyte allocation before any shot runs; a large `num_workers` tries to spawn that many OS threads. Worth a sanity bound at load time.

### Design

**Recorded per-worker shots are ignored (`engine.rs:386-387`)** — replay recomputes `distribute_shots(num_shots, num_workers)` and drives execution from the re-derived split; `workers[i].shots` is only asserted against, never used. That makes the report reproducible only while the distribution policy is unchanged: a report written by another build or policy, or hand-edited to reproduce one worker's shots, either trips the assert or runs a different split than the report documents. Driving replay from the recorded `(worker_idx, shots, seed)` records would make the report self-describing and remove the need for the assert entirely.

### Packaging

- **`crates/pecos-engines/Cargo.toml:28`** — `tempfile` is in `[dependencies]` but used only by `tests/seed_report.rs`. Every other crate in the workspace lists it under `[dev-dependencies]`. As written, `pecos-engines` and all its dependents ship `tempfile` and its `rustix`/`getrandom`/`windows-sys` chain as runtime dependencies.
- **`Cargo.lock`** — beyond the `tempfile` edge, the lockfile moves unrelated transitive dependencies backwards: `unicode-width` 0.2.2 → 0.1.14, `windows-sys` 0.61.2 → 0.52.0 in five places, `getrandom` 0.4.3 → 0.3.4. A clean re-resolve does not produce these, so it looks like regeneration under an older toolchain. Worth restoring the lockfile and re-adding only the `tempfile` edge.

### Tests

`to_json_file` is the only new public write path and has no coverage — the tests exercise `from_json_str`, `from_json_file`, and the missing-file error, but never write a report and read it back. A save/load round-trip test would close that gap (and would justify the `tempfile` dependency, once it moves to `[dev-dependencies]`).

ciaranra (commented):
Anchoring three of the items from my [earlier comment](https://github.com/PECOS-packages/PECOS/pull/378#issuecomment-5260087647) to the lines they apply to — same findings, no new ones.

crates/pecos-engines/src/monte_carlo/engine.rs
```diff
@@ -274,8 +286,106 @@ impl MonteCarloEngine {
...
         assert!(num_workers > 0, "num_workers cannot be zero");
+        if seed_report.workers.len() < num_workers {
+            return Err(PecosError::Input(format!(
+                "Seed report contains {} worker records, but num_workers is {num_workers}",
+                seed_report.workers.len()
+            )));
+        }
+
+        let shots_per_worker = distribute_shots(num_shots, num_workers);
+        self.set_seed(seed_report.root_seed); // make sure to update root seed.
```
ciaranra, line 388: This is the blocking item. Because `run` and `run_with_workers` both funnel through `run_with_workers_from_seed_report`, this line re-seeds the engine on *every* run, so consecutive runs on one engine return byte-identical shots and replaying an archived report overwrites the caller's seed.

Replay does not need it: each worker is seeded from the report at line 404, and nothing downstream reads `self.rng` or the template seed. Removing it restores `base_seed = self.rng.next_u64()` advancing per run, and makes replay non-mutating.

```suggestion
```

With this gone, the only remaining use of `&mut self` in this method is the template clone, so the signature can become `&self` — which turns "replay does not mutate the engine" into a compiler-checked property. Existing callers are unaffected.

crates/pecos-engines/src/monte_carlo/engine.rs
```diff
@@ -284,23 +394,26 @@ impl MonteCarloEngine {
...
         // This avoids potential deadlocks when worker threads try to clone engines
         // simultaneously, which can trigger concurrent library loading operations
         // that contend with each other or the dynamic linker.
         let worker_engines: Vec<_> = (0..num_workers)
             .map(|worker_idx| {
                 let mut engine = self.hybrid_engine_template.clone();
-                let worker_seed = derive_seed(base_seed, &format!("worker_{worker_idx}"));
-                engine.set_seed(worker_seed);
+                engine.set_seed(seed_report.workers[worker_idx].seed);
                 (worker_idx, shots_per_worker[worker_idx], engine)
```
ciaranra, line 405: Related to the same comment: the worker tuples are built from `shots_per_worker` (re-derived by `distribute_shots` on line 387) rather than from the records in the report, which is what forces the two `assert!(.., "..")`s below to police the redundancy — and those abort the process on a hand-edited or older-format report instead of returning `PecosError::Input`.

Driving replay from the records instead makes the report self-describing and lets the asserts, the `distribute_shots` call, and the one-sided `workers.len() < num_workers` check all go away:

```rust
let worker_engines: Vec<_> = seed_report
    .workers
    .iter()
    .map(|record| {
        let mut engine = self.hybrid_engine_template.clone();
        engine.set_seed(record.seed);
        (record.worker_idx, record.shots, engine)
    })
    .collect();
```

Then `num_threads` takes `seed_report.workers.len()`, and the `Vec::with_capacity(num_shots)` on line 393 can be a plain `Vec::new()` — a few reallocations are nothing next to running the shots, and a corrupted `"num_shots": 10000000000000` no longer requests a multi-terabyte allocation before the first shot.

If the scalar fields should stay authoritative, one consistency check at the top of the method replaces all three panics:

```rust
if seed_report.workers.is_empty() {
    return Err(PecosError::Input(
        "Seed report contains no worker records".to_string(),
    ));
}
if seed_report.workers.len() != seed_report.num_workers {
    return Err(PecosError::Input(format!(
        "Seed report contains {} worker records, but num_workers is {}",
        seed_report.workers.len(),
        seed_report.num_workers
    )));
}
```

Either way the `# Panics` section on the doc comment can go.

crates/pecos-engines/Cargo.toml
```diff
@@ -25,6 +25,7 @@ bitflags.workspace = true
 dyn-clone.workspace = true
 num-bigint.workspace = true
 bitvec.workspace = true
+tempfile.workspace = true
```
ciaranra, line 28: `tempfile` is used only by `tests/seed_report.rs`, so as written it ships as a runtime dependency of `pecos-engines` and everything downstream (`pecos`, `pecos-hugr`, `pecos-qasm`, `pecos-foreign`, `pecos-phir-json`, benchmarks), pulling in its `rustix`/`getrandom`/`windows-sys` chain. Every other crate in the workspace lists it under `[dev-dependencies]`.

```suggestion
```

and add at the end of the file:

```toml
[dev-dependencies]
tempfile.workspace = true
```

ciaranra:
Follow-up: I applied the suggestions from the inline comments in a scratch worktree at `c8a9c721c` and ran them, so the numbers below are measured rather than predicted.

**The regression reproduces on the current head.** Two tests, both failing before any change:

```
consecutive_runs_on_one_engine_are_not_identical
  base_seed did not advance between runs: 15898102487349570925

replay_does_not_mutate_caller_engine_stream
  replay perturbed the caller's RNG stream
  left: 1289665474627176275   right: 15898102487349570925
```

The second compares two engines seeded at 999: one runs twice, the other runs, replays an unrelated report, then runs again. The report's `root_seed` leaks into the second engine's stream.

**With the fixes applied** — drop the `self.set_seed(...)` line, drive `worker_engines` from `seed_report.workers`, replace the asserts with one `PecosError::Input` consistency check, `Vec::new()` instead of `with_capacity(num_shots)`, `num_threads(seed_report.workers.len())`, `&self` instead of `&mut self`, and `tempfile` moved to `[dev-dependencies]`:

- `cargo test -p pecos-engines`: 17 test binaries, all pass, 0 failures. That includes the seven existing `tests/seed_report.rs` tests unchanged — both replay-reproduces-original tests still pass, so the replay guarantee is preserved.
- `cargo clippy -p pecos-engines --all-targets` (run cold): clean, no warnings.
- `cargo check --all-targets` on `pecos`, `pecos-hugr`, `pecos-qasm`, `pecos-phir-json`: all OK, so nothing downstream depended on the `&mut self` signature or on `tempfile` being a normal dependency.

Net diff: 28 insertions, 36 deletions across `engine.rs`, `Cargo.toml`, and two lines of `tests/seed_report.rs`.

Two things the change turns up that are worth knowing before you apply it:

1. Removing the whole `# Panics` doc section trips clippy's `missing_panics_doc`, because `results_vec.lock().expect("results mutex poisoned")` in the worker loop still panics. Keeping a one-line `# Panics` that names only the poisoned mutex clears it.
2. The `&mut self` → `&self` change makes the two `let mut replay_engine = ...` bindings in `tests/seed_report.rs` warn `variable does not need to be mutable`. Dropping `mut` in both is the whole fix.

Not verified on my side: the `Cargo.lock` re-resolve (re-resolving here would only reflect my toolchain), and a full `cargo check --workspace`, which fails in my environment on `llvm-sys` for want of a local LLVM — unrelated to this PR.

ciaranra:
## Review, round 3

All five findings from the earlier rounds are fixed and verified at `ebf5cfe17`:

- Ordinary runs draw a fresh `base_seed` from the engine RNG again; replay takes `&self` and leaves `seed`, `rng`, and the template untouched.
- Replay is driven from the recorded `(worker_idx, shots, seed)` records; `SeedReport::validate` returns `PecosError::Input` for every malformed shape I tried and there are no asserts left on deserialized data.
- `tempfile` is under `[dev-dependencies]` and the lockfile change is the single `tempfile` edge.
- `to_json_file` has round-trip and error-path coverage.
- The replay tests are no longer vacuous. I mutation-tested `engine.rs` four ways (drop `engine.set_seed(record.seed)`, re-derive worker seeds from `base_seed` instead of the record, skip `validate()` in replay, drop the result sort) and each mutation fails between 1 and 11 of the 19 tests in `tests/seed_report.rs`.

Locally: `cargo test -p pecos-engines` passes across all test binaries, `cargo clippy -p pecos-engines --all-targets -- -D warnings` is clean on a cold run, and CI is green. I also confirmed that on `dev` `new_with_depolarizing_noise` panics at `hybrid/builder.rs:198` for want of a quantum engine, so installing a `StateVecEngine` sized from the classical engine is a fix, not just a convenience for the tests.

Four small items remain, none of them structural.

### 1. Restore the shot and worker counts in the debug logs

Commit `90f321115` (the Copilot autofix for CodeQL alert 222) removed `{num_shots}` and `{num_workers}` from the two `debug!` calls. The alert is a false positive: shot and worker counts are not sensitive, and the taint CodeQL reports comes from PyO3 argument extraction. The autofix also left every ordinary run logging `Running Monte Carlo simulation` followed by `Replaying Monte Carlo simulation`, because `run_with_workers_report_seeds` delegates to the replay method.

Please restore the counts and give the shared executor neutral wording, for example `Executing {} shots across {} workers` at the top of `run_with_workers_from_seed_report`. I will dismiss the alert as a false positive.

### 2. `ExternalClassicalEngine` never resets on the Monte Carlo path

`HybridEngine::reset` calls `ClassicalEngine::reset` (`crates/pecos-engines/src/hybrid/engine.rs:111`), and that trait method has a default no-op body (`crates/pecos-engines/src/classical.rs:58`). `ExternalClassicalEngine` implements only `Engine::reset`, so the result-clearing code never runs during a shot loop. The new doc on `new_with_circuit` ("Calling `reset` retains the existing result fields and sets every value back to zero before the next shot") is therefore not true through `HybridEngine`.

This is masked today because a fixed circuit overwrites `result` on every shot, and the defect predates this PR, but the PR is what documents the promise. Implementing `ClassicalEngine::reset` on the fixture by delegating to `Engine::reset`, plus a test that goes through `HybridEngine::reset` with two measurement fields, closes it.

### 3. Noise installation is unpinned

Changing the fixture probability from `0.1` to `0.0` in `tests/seed_report.rs` leaves all 19 tests passing (verified). So the mutation `.with_depolarizing_noise(p)` to `.with_depolarizing_noise(0.0)` in `new_with_depolarizing_noise` survives the suite. A short test would pin it: `PZ; MZ` on qubit 0 with `p = 0.0` yields all zeros, with `p = 0.5` yields some ones.

### 4. PR description is stale

The description still refers to `run_with_workers_seed_report` with a `bool` flag and a hardcoded `seed_report.json`. Please update it to the current surface: `run_with_workers_report_seeds` returning `(ShotVec, SeedReport)`, `run_with_workers_from_seed_report`, and `SeedReport::to_json_file` / `from_json_file`.

### Nit

The `# Panics` section on `run_with_workers_from_seed_report` names only the poisoned mutex; engine clone, `set_seed`, and `reset` run outside `catch_unwind`, so a component panic there also propagates. Fine to leave, or add one sentence.

### Considered and not requested

A resource bound on loaded reports (very large `num_shots`, or many zero-shot worker records each cloning an engine). A report is a workload specification with the same trust as the `num_shots` / `num_workers` arguments, and `run_with_workers(usize::MAX, 1)` is equally unbounded on `dev`. Not needed for this PR.

Verdict: approve once items 1 and 4 land. Items 2 and 3 are small enough to ride in the same push.

ciaranra:
Addendum to the round-3 review: a reuse pass over the workspace.

### 5. Reuse `shot_from_byte_message` in the fixture

`crates/pecos-engines/src/shot_results.rs:34` already provides `shot_from_byte_message(message, result_id_to_name)`, which calls `message.outcomes()`, names each outcome from the map (falling back to `result_{id}`), and builds a `Shot`. That is what the new `ExternalClassicalEngine::handle_measurements` plus the existing `get_results` do together by hand, via an intermediate `BTreeMap<String, i64>` and a sign-conversion loop. The helper currently has no callers.

Suggested shape: the fixture stores a `Shot` instead of the `i64` map, `handle_measurements` becomes a single call to `shot_from_byte_message` with `{0 => "result"}` as the name map (which reproduces the current `result`, `result_1`, `result_2` naming and keeps the `result = 0` default for empty circuits), `get_results` returns a clone, and the `u32` / `u64` / `i64` conversion in `get_results` goes away. This also gives item 2 (the `ClassicalEngine::reset` gap) a single obvious state to zero.

### Checked and fine

- JSON file round trips: there is no shared serde-to-path helper anywhere in `crates/`. `PhirJsonEngine::new`, the pecos-qis trace writers, and the pecos-qec DEM metadata code all hand-roll `read_to_string` + `from_str` / `to_string_pretty` the same way this PR does, so the `from_json_str` / `from_json_file` / `to_json_file` trio matches precedent. Not asking for a shared helper.
- Parallel shot loop: this PR collapsed the two copies into one. The only other copies live in `exp/pecos-neo` and predate this work.
- `RecordingRng` / `ReplayingRng` in `pecos-core` capture RNG values, not seeds. Seed-level replay is the right granularity for Monte Carlo, so there is no overlap.
- Manual `Debug` for `ExternalClassicalEngine`: `ByteMessage` has no `Debug` anywhere in the workspace, so the impl is justified. Deriving `Debug` on `ByteMessage` would be a simpler alternative, but it is a nit.

Verdict unchanged.

drewconnelly-qntm:
Comments 1,3 resolved in code.

Comment 2 seems to be a longer-standing bug but should be fixed in the suggested way.

Resolved comment 4 by updating the PR description.

crates/pecos-engines/src/monte_carlo/engine.rs
```diff
@@ -274,30 +286,115 @@
...
+    /// Panics if the results mutex is poisoned.
+    pub fn run_with_workers_from_seed_report(
+        &self,
+        seed_report: &SeedReport,
+    ) -> Result<ShotVec, PecosError> {
+        seed_report.validate()?;
+
+        debug!(
+            "Executing {} shots across {} workers",
+            seed_report.num_shots, seed_report.num_workers
```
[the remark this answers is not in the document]
drewconnelly-qntm: This is the false-positive security concern referenced above. I suggest we dismiss it. It is logging # shots and # workers, not sensitive information. I'll leave it here for @ciaranra to dismiss.

ciaranra (approved):
Round 4: all four items from round 3 landed in `58844e50c` and I verified each one rather than the reply.

- `ClassicalEngine::reset` on `ExternalClassicalEngine` now delegates to `Engine::reset`, and the new hybrid-reset test goes through `HybridEngine::reset` with two fields. Replacing the delegation with `Ok(())` fails exactly that test.
- The depolarizing constructor is pinned: replacing `.with_depolarizing_noise(p)` with `0.0` fails exactly the new noise test (`p=0.5: measured 0 ones`).
- Both `debug!` lines carry the counts again with neutral wording.
- The description matches the current surface.

Locally at `58844e50c`: cold `cargo clippy -p pecos-engines --all-targets -- -D warnings` is clean and `cargo test -p pecos-engines --no-fail-fast` passes across every binary. CodeQL alerts 224 and 225 are dismissed as false positives (they flag shot and worker counts tainted through PyO3 argument extraction); the remaining CodeQL neutral status is the known `go` configuration residue and is non-blocking.

One cosmetic leftover that does not need another push: an ordinary run logs the identical `Executing {} shots across {} workers` line twice, once in `run_with_workers_report_seeds` and once in the replay method it delegates to. Dropping the first would be enough if anyone touches this later.

Approving.

Issue #382, opened 2026-07-28
Batched MZ nodes: InfluenceBuilder records one measurement, DagFaultAnalyzer records one per qubit


ciaranra:
**Impact: batched measurement nodes silently lose measurements in one of the two analyzers that `from_circuit` composes.**

A `DagCircuit` node may carry several gate instances; `DagCircuit::gate_count` counts them individually (`dag_circuit.rs:557`), and `Gate::mz(&[0, 1])` builds exactly such a node. The two analyzers that `build_dem_from_circuit` runs together disagree about what that means:

| | batched `MZ` node handling |
| --- | --- |
| `DagFaultAnalyzer` (`propagator/dag.rs:1946`) | emits one measurement entry **per qubit** |
| `InfluenceBuilder::run_symbolic_simulation` (`influence_builder.rs`) | `sim.mz(&[qubits[0]])`, increments `meas_idx` **once**, stores **one** index for the node |

So for a node measuring two qubits, one analyzer reports two measurements and the other reports one. `node_to_meas_idx` is `Vec<Option<usize>>` -- one slot per node -- so it cannot express more than a single measurement per node even in principle.

`build_dem_from_circuit` composes both (`dem_builder/builder.rs:3208`), and the exact-branch replay path consumes those histories (`dem_builder/builder.rs:1202`), so mismatched measurement counts and node→index mappings can propagate into detector extraction with no error raised.

## Not introduced by recent work

This predates #376. That PR fixed the analogous divergence for *unitary* gates (batched nodes now replay every instance) and, in follow-up, for batched two-qubit propagation in `apply_gate`. The measurement path was deliberately left alone because fixing it needs a data-structure change, not a one-line change: `node_to_meas_idx` must be able to map one node to several measurement indices, distinguishing which qubit each belongs to.

## Suggested directions

1. Support every measurement instance -- replace `node_to_meas_idx: Vec<Option<usize>>` with a mapping that can carry one index per (node, qubit), and have the symbolic simulation measure every qubit in the node.
2. Or reject batched measurement nodes explicitly at the `DagCircuit` API boundary, so the ambiguity cannot be constructed. That is a smaller change but narrows what `DagCircuit` can represent, and `gate_count` already documents batched nodes as legitimate.

Option 1 matches the semantics the rest of the stack has now settled on. Option 2 is only reasonable if batched measurement nodes are considered a mistake to begin with -- worth deciding explicitly rather than leaving the two analyzers disagreeing.

[...]

PR #384, opened 2026-07-28
Make observable metadata win over observable annotations (fixes #377)

### LINKED ISSUE #377: from_circuit and generate_dem_from_tick_circuit produce different DEMs for the same circuit
**Impact: silent wrong results from a public API.** `DetectorErrorModel.from_circuit` emits spurious error mechanisms for any circuit carrying both annotations and detector/observable metadata -- which is what the surface generators produce. An independent Stim oracle agrees with the other path and disagrees with `from_circuit` (86 vs 76 mechanisms at d=3). No exception is raised and the detector space is unchanged, so a caller gets a plausible-looking DEM that is wrong. See the correction comment below for the oracle result.

The two public DEM construction paths return different DEMs for the same `TickCircuit` under identical noise. The cause is isolated: one path merges circuit annotations into the influence map and the other does not.

Found while auditing the `from_circuit` surface for #325 (see #371, #376). It is unrelated to `F`/`SY` propagation — see the control below.

## Reproduction

d=3 surface patch, 1 round, p2-only noise, `decompose_errors=False`:

```python
from pecos.qec import DetectorErrorModel
from pecos.qec.surface import SurfacePatch
from pecos.qec.surface.circuit_builder import (
    generate_tick_circuit_from_patch,
    generate_dem_from_tick_circuit,
)

tc = generate_tick_circuit_from_patch(
    SurfacePatch.create(distance=3), num_rounds=1, basis="Z", interaction_basis="cx"
)
noise = dict(p1=0.0, p2=0.01, p_meas=0.0, p_prep=0.0)

a = DetectorErrorModel.from_circuit(tc, **noise).to_string()
b = generate_dem_from_tick_circuit(tc, **noise, decompose_errors=False)
assert a == b  # fails: 99 vs 89 lines, 26 unique to `a`, 16 unique to `b`
```

The differences are mostly observable assignment, e.g. `from_circuit` emits `error(0.00067) D0 D1 D7` where the other path emits `error(0.00067) D0 D1 D7 L0`.

## Not an F/SY issue

Identical divergence (26 / 16 unique lines) on a plain CX circuit containing no `F` or `SY` at all:

| circuit | from_circuit only | influence-path only |
| --- | --- | --- |
| `interaction_basis="cx"` | 26 | 16 |
| `interaction_basis="szz"`, no prefixes | 26 | 16 |
| `interaction_basis="szz"`, with prefixes | 26 | 16 |

## Cause

`build_dem_from_circuit` (`dem_builder/builder.rs:3288`) unconditionally merges annotation-derived DEM outputs:

```rust
let annotation_map = InfluenceBuilder::new(circuit)
    .with_circuit_annotations(circuit)
    .build();
influence_map.merge_dem_outputs_from(&annotation_map);
```

`generate_dem_from_tick_circuit` calls `_build_canonical_dem_influence_map(dag)` (`circuit_builder.py:3903`), whose `include_circuit_annotations` defaults to `False`, so that merge never happens.

Rebuilding the Python path with `include_circuit_annotations=True` reproduces `from_circuit` byte-for-byte, in every configuration tested:

| circuit | basis | annotations merged | annotations not merged |
| --- | --- | --- | --- |
| cx plain | Z / X | matches | differs |
| szz no prefixes | Z / X | matches | differs |
| szz with prefixes | Z / X | matches | differs |

This is the whole explanation. `with_measurement_order` was the other candidate and is ruled out: toggling it changes nothing about whether the two agree, in either direction.

## Open question

Which path is right? These surface circuits carry observable annotations, so `from_circuit` picks up annotation-derived observable outputs that `generate_dem_from_tick_circuit` drops. Whether the Python default is a deliberate choice (metadata JSON is meant to be the sole source of observables) or an oversight determines the fix:

- if annotations should always be merged, `_build_canonical_dem_influence_map` should default to `True`;
- if the metadata JSON is authoritative, `build_dem_from_circuit` is over-merging;
- if both are legitimate for different callers, the difference should be documented and the two paths should not be assumed interchangeable.

No fix proposed here — the intent needs deciding first.

ciaranra:
Fixes #377.

`DetectorErrorModel::from_circuit` returned DEMs with spurious error mechanisms for any circuit carrying both circuit annotations and detector/observable metadata, which is what the surface-code generators produce. An independent Stim oracle agreed with the other DEM path and disagreed with `from_circuit` (86 mechanisms versus 76 at d=3).

## Root cause

An observable **annotation** silently overrode the observable declared in **metadata JSON**.

`build_dem_from_circuit` installed annotation-derived DEM outputs into the influence map (`merge_dem_outputs_from`) *before* it read the metadata attributes. The annotation's definition therefore won during mechanism generation even though `with_observables_json` was called with the caller's explicit declaration.

Minimal reproduction -- metadata covers both measurements, the annotation covers only the first, so the probability identifies which definition was used:

```
WITHOUT annotation:  error(0.375) L0     <- 2 * 0.25 * 0.75, observable over BOTH measurements (metadata)
WITH annotation:     error(0.25)  L0     <- flips only on the first measurement (annotation)
```

The asymmetry is essential. With one measurement per qubit and a symmetric observable the two definitions produce identical DEMs, and an earlier version of this test proved nothing for exactly that reason.

## The fix

Metadata is parsed **before** annotations are collected, and observable annotations are excluded when the circuit supplies observables JSON:

```rust
let annotation_map = if obs_json.is_some() {
    annotations.with_circuit_annotations_excluding_observables(circuit)
} else {
    annotations.with_circuit_annotations(circuit)
}
.build();
```

`with_circuit_annotations_excluding_observables` is a new `pub(crate)` sibling sharing one inner implementation, so the ~25 existing callers of the public method are untouched. Tracked-Pauli annotations are still collected either way -- metadata has no equivalent for them.

This matches the precedence the very next block already applied to observable *records*: metadata first, annotations only as fallback.

## Coupled fix: observable labels from metadata

Excluding annotations initially broke `sampler_paths_preserve_output_split_for_noiseless_and_forced_faults`, which expected a DEM-output label of `"obs0"` and got `None`.

That was not the test encoding the bug. **`parse_observables_json` discarded the `label` field.** `ParsedObservable` had no label, and DEM-output labels were populated in exactly one place (`influence_builder.rs:441`) from annotation-derived outputs -- so annotations were the only possible source of an observable label, even though the metadata format already carries one and callers already write it.

Editing that test to expect `None` would have been a real regression. Instead the metadata format is now self-sufficient: `ParsedObservable` carries `label`, `parse_single_observable` reads it, and `try_build` attaches it when emitting the observable. The previously-failing test now passes on its own merits, with the label coming from the metadata that always declared it.

## Testing

Four tests, each verified to FAIL when the corresponding change is reverted -- not merely to pass:

| test | pins |
| --- | --- |
| `test_observable_metadata_wins_over_observable_annotation` | metadata records win over a disagreeing annotation |
| `test_observable_annotation_used_when_metadata_absent` | the fallback path still works (passes either way by design -- guards against over-fixing) |
| `test_observable_label_comes_from_metadata_without_annotations` | a metadata label reaches the DEM with no annotation present |
| `sampler_paths_preserve_output_split_for_noiseless_and_forced_faults` (existing) | label survives when both sources are present |

- `cargo test -p pecos-qec` green: 598 lib tests and 58 doctests, 0 failed.
- `cargo check --workspace --all-targets`, `cargo clippy -p pecos-qec --all-targets`, `cargo fmt --all --check` all clean.

## End-to-end verification (draft criterion met)

Built the `pecos-rslib` extension from this branch into an isolated venv and re-ran the original Stim comparison. Same script, same circuits, only the binary differs:

| | `from_circuit` mechanisms | agrees with Stim? |
| --- | --- | --- |
| `dev` (pre-fix) | 86 (Z basis) / 80 (X basis) | no -- all four configs |
| this branch | **76** in all four | **yes -- all four configs** |

Covered: `interaction_basis="cx"` and `interaction_basis="szz", szz_physical_prefixes=True`, each in Z and X basis, d=3, 1 round, p2-only. `influence == stim` in both runs, confirming that path was unaffected and that nothing else shifted.

The spurious mechanisms reported in #377 are gone.

One methodology note: a `pecos_rslib.pth` initially shadowed the built wheel with the repo source tree, which would have silently tested the pre-fix binary. The run above loads the wheel's own `pecos_rslib.abi3.so` from the isolated venv.

### DIFF
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -52,6 +52,12 @@ struct ParsedObservable {
     id: u32,
     records: Vec<i32>,
     meas_ids: Vec<usize>,
+    /// Human-readable name from the metadata JSON's `label` field.
+    ///
+    /// The metadata format has always carried this and callers already write
+    /// it, but it used to be parsed and dropped, leaving circuit annotations as
+    /// the only way to get a label onto an observable (issue #377).
+    label: Option<String>,
 }
 
 // ============================================================================
@@ -537,6 +543,7 @@ impl<'a> DemBuilder<'a> {
                 id: id as u32,
                 records,
                 meas_ids: Vec::new(),
+                label: None,
             })
             .collect();
         self.clear_exact_branch_cache();
@@ -810,7 +817,10 @@ impl<'a> DemBuilder<'a> {
         // Observable IDs are not shifted by tracked Paulis.
         for obs in &self.observables {
             let records = self.effective_record_offsets(&obs.records, &obs.meas_ids);
-            let def = DemOutput::new(obs.id).with_records(records.iter().copied());
+            let mut def = DemOutput::new(obs.id).with_records(records.iter().copied());
+            if let Some(label) = &obs.label {
+                def = def.with_label(label.clone());
+            }
             dem.add_observable(def);
         }
 
@@ -2654,11 +2664,16 @@ fn parse_single_observable(value: &serde_json::Value) -> Result<ParsedObservable
     )?;
 
     let (records, meas_ids) = extract_measurement_refs(object, "observable")?;
+    let label = object
+        .get("label")
+        .and_then(serde_json::Value::as_str)
+        .map(std::string::ToString::to_string);
 
     Ok(ParsedObservable {
         id,
         records,
         meas_ids,
+        label,
     })
 }
 
@@ -5268,6 +5283,36 @@ mod tests {
             "with no observables metadata the annotation defines L0 over m0 \
              alone, so it flips at p_meas; got:\n{}",
             dem.to_string()
+        );
+    }
+
+    /// Issue #377: an observable label declared in metadata JSON must reach the
+    /// DEM without an annotation supplying it.
+    ///
+    /// The metadata format has always carried `label` and callers already write
+    /// it, but the parser dropped it, so annotations were the only route. That
+    /// coupling is what made excluding annotations lose labels.
+    #[test]
+    fn test_observable_label_comes_from_metadata_without_annotations() {
+        use pecos_num::graph::Attribute;
+        use pecos_quantum::DagCircuit;
+
+        let mut circuit = DagCircuit::new();
+        circuit.pz(&[0]);
+        let _meas = circuit.mz(&[0]);
+        circuit.set_attr("num_measurements", Attribute::String("1".to_string()));
+        circuit.set_attr(
+            "observables",
+            Attribute::String(r#"[{"id":0,"records":[-1],"label":"from_metadata"}]"#.to_string()),
+        );
+        circuit.set_attr("detectors", Attribute::String("[]".to_string()));
+
+        let dem = DemBuilder::from_circuit(&circuit, 0.0, 0.0, 0.5, 0.0);
+        assert_eq!(dem.num_observables(), 1);
+        assert_eq!(
+            dem.observables().next().unwrap().label.as_deref(),
+            Some("from_metadata"),
+            "the label declared in observables metadata must reach the DEM"
         );
     }
 }
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -3205,14 +3205,10 @@ fn build_dem_from_circuit(
     use crate::fault_tolerance::propagator::DagFaultAnalyzer;
     use pecos_num::graph::Attribute;
 
-    let mut influence_map = DagFaultAnalyzer::new(circuit).build_influence_map();
-    let annotated_observable_records = observable_records_from_annotations(circuit, &influence_map);
-    let annotation_map = InfluenceBuilder::new(circuit)
-        .with_circuit_annotations(circuit)
-        .build();
-    influence_map.merge_dem_outputs_from(&annotation_map);
-
-    // Extract metadata before building (to avoid borrow issues)
+    // Metadata is read BEFORE annotations are collected: when the caller
+    // supplies observables JSON, that declaration wins, and an observable
+    // annotation describing the same output must not be installed into the
+    // influence map where it would silently override it (issue #377).
     let det_json = circuit.get_attr("detectors").and_then(|a| {
         if let Attribute::String(s) = a {
             Some(s.clone())
@@ -3234,6 +3230,19 @@ fn build_dem_from_circuit(
             None
         }
     });
+
+    let mut influence_map = DagFaultAnalyzer::new(circuit).build_influence_map();
+    let annotated_observable_records = observable_records_from_annotations(circuit, &influence_map);
+    let annotations = InfluenceBuilder::new(circuit);
+    // Tracked-Pauli annotations have no metadata equivalent, so they are
+    // collected either way.
+    let annotation_map = if obs_json.is_some() {
+        annotations.with_circuit_annotations_excluding_observables(circuit)
+    } else {
+        annotations.with_circuit_annotations(circuit)
+    }
+    .build();
+    influence_map.merge_dem_outputs_from(&annotation_map);
 
     let builder = DemBuilder::new(&influence_map)
         .with_noise_config(noise)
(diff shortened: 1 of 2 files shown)


ciaranra:
Independent adversarial review returned **REVISE**. I verified the load-bearing findings against the code rather than taking them; all three confirmed. Converting this back to draft.

## Confirmed: the fix covers one of two entry points

`DemSampler::from_circuit` (`sampler.rs:424`) carries the identical defect this PR fixes in `build_dem_from_circuit`:

```rust
let annotation_map = InfluenceBuilder::new(circuit)
    .with_circuit_annotations(circuit)
    .build();
influence_map.merge_dem_outputs_from(&annotation_map);
// ... metadata read afterwards
```

So #377 still reproduces through `DemSampler::from_circuit` / the Python `DemSampler.from_circuit`. I fixed `DemBuilder` and did not check its sibling. The two need shared preprocessing rather than parallel copies -- the same "two hand-maintained copies of one thing" shape as #376.

## Confirmed: this PR regresses the observable label on the real surface path

My end-to-end verification compared error mechanisms only, so it could not see this.

The surface generator writes observables JSON **without** a label (`circuit_builder.py:2681`):

```python
observables = [{"id": 0, "records": logical_rec_offsets}]
```

and puts the label solely on the annotation (`circuit_builder.py:2797`):

```python
circuit.observable(obs_refs, label=f"logical_{basis.upper()}")
```

Excluding observable annotations therefore removes `logical_Z` / `logical_X` from the generated DEM -- and also the observable's Pauli metadata, which metadata JSON has no field for at all. My label plumbing only helps callers who put a label in the JSON; the default generator does not.

## Confirmed: the label conflict path is fail-silent in release

`merge_observable_definition` (`types.rs:5104`) resolves conflicting labels with `debug_assert_eq!` -- panics in debug, silently keeps the existing label in release. Parsing metadata labels makes that path newly reachable, so this PR increases exposure to it.

## On failing fast and loudly

Reviewing this against the project's fail-loud rule, the current answer is **no**, in three places:

1. `debug_assert_eq!` above -- the worst of both worlds: a debug-only panic with no message a user could act on, and silent divergence in release. Should be a real error naming both labels and the observable.
2. `parse_single_observable` (`builder.rs:2667`) silently discards a malformed label: `"label": 42` is treated identically to no label. The richer DEM parser correctly rejects non-string/non-null labels (`types.rs:3351`). Should be a `ParseError`.
3. This PR's own precedence is silent. Rejecting a genuine conflict is the right end state, but only after #382, since today the "conflict" is PECOS misresolving the annotation (see below).

## Revised plan

1. **#382 first, and broader than filed.** The review found the one-index-per-node map in *three* places (`builder.rs:3331`, `sampler.rs:1109`) plus a related batched-node defect in `InfluenceBuilder` (`influence_builder.rs:229`) where a reference to a batched measurement node becomes a propagation term over *every* qubit that node measures. Fixing `.or_insert` alone is not sufficient -- it needs per-measurement identity (node + qubit, or a stable `MeasId`).
2. **Then decide the authoritative representation for observables** -- records, label and Pauli -- so excluding annotations cannot silently drop fields metadata is unable to express. Either extend the metadata format or validate equivalence and merge the annotation-only fields.
3. **Then this PR's precedence**, applied to `DemBuilder` and `DemSampler` through one shared path, with loud rejection of genuine conflicts.
4. Independently of all of the above: make the two silent failures in (1) and (2) above loud.

The review confirmed the three tests are sound (`test_observable_annotation_used_when_metadata_absent` passes against pre-PR code by design -- it guards against over-fixing rather than detecting the bug), and that tracked-Pauli ID allocation and ordering are unaffected. The defect is scope, not the mechanism.

ciaranra:
Parking this and splitting out the part that stands alone.

**Split out: #388** -- carrying observable labels from metadata JSON, plus rejecting a malformed `label` rather than silently dropping it. That work was originally needed here (excluding annotations otherwise lost labels) but is not contingent on the precedence decision: the metadata format documents a `label`, callers write one, and the parser discarded it. It is worth landing on its own merit.

**Parked: the precedence change itself.** Three reasons.

1. **Its premise weakened.** This PR made metadata authoritative on the grounds that the annotation-derived observable was untrustworthy. #386 established *why* it was untrustworthy -- the Python binding fabricated `record_idx: 0` for every reference, so an observable over three measurements collapsed to one. That is a corruption, not a disagreement. Deciding precedence against corrupted data decides the wrong question.

2. **It is incomplete.** Review found `DemSampler::from_circuit` (`sampler.rs:424`) carries the byte-identical defect this PR fixes in `build_dem_from_circuit`, so #377 would still reproduce through the Python `DemSampler.from_circuit`.

3. **It would regress the generated surface path.** The surface builder writes observables JSON *without* a label and puts it solely on the annotation (`circuit_builder.py:2681`, `:2797`). Excluding annotations therefore strips `logical_Z`/`logical_X` from generated DEMs, along with the observable's Pauli, which the JSON format cannot express at all. My end-to-end verification compared error mechanisms only and could not see this.

## What should happen instead

After #387 fixes measurement-record resolution, a disagreement between metadata and annotations will mean the two sources genuinely differ. At that point the right response is probably to **reject it loudly** rather than silently pick a winner -- the conflict check prototyped on #377 already works and is what surfaced the `[-9, -9, -9]` corruption in the first place. It was not shipped only because, today, it fires on the standard surface path for a defect that is PECOS's fault rather than the caller's.

Leaving this open as a placeholder for that decision rather than closing it, since the underlying question -- what happens when both sources describe the same output -- is still unanswered and worth answering deliberately.

ciaranra:
Closing as superseded, on two grounds.

**The policy this PR implements was rejected.** It makes observable metadata silently win over observable annotations. The direction actually taken -- decided while this sat as a draft -- is that conflicting definitions **fail loudly** in every build: #385 (merged) enforces that for label and Pauli conflicts, and the fail-fast principle has since been applied across the DEM ingest paths. Silent precedence in either direction is the behaviour we removed.

**Its target is fixed.** #377 is closed; the `from_circuit` divergence was resolved through #386 and the measurement-identity work (#391, #397, #401) rather than through output-priority reshuffling.

The reproduction in this PR's description -- metadata covering both measurements while the annotation covers one, with the probability identifying which definition won -- remains a good de-aliased fixture shape, and the #387 step-3/4 pivot (which reworks annotation resolution onto `MeasId`) should pick it up as a regression case. Recorded there.

Branch left in place; nothing here merges.

PR #391, opened 2026-07-29
Give DagCircuit measurements a MeasId of their own

ciaranra:
Every measurement in a `DagCircuit` now carries a unique `MeasId`. Step 1 of the plan in #387.

`meas_ids` appeared zero times in `dag_circuit.rs` before this: `mz`/`mz_free`/`mz_labeled` built measurements with no identity at all, so PECOS's own native DAG builder was the largest producer of id-less measurements. Anything downstream that needed to name one measurement had to fall back on node ids, which cannot distinguish the measurements inside a single batched gate.

Ids are assigned in `try_add_gate`, the one path every gate reaches the circuit through. Gates arriving with ids keep them -- they came from a `TickCircuit` conversion or an external source that owns the numbering.

## This changes measurement ordering, deliberately

The first revision of this PR claimed no downstream behaviour change. That was wrong, and two independent reviews refuted it by running.

Both `DagFaultAnalyzer::extract_measurements` and `InfluenceBuilder::extract_measurements` pick their sort key based on whether ids are present: topological position when absent, `MeasId` order when present. Minting ids everywhere therefore flips every `DagCircuit`-built circuit onto the id key. Existing tests did not catch it because minted ids coincide with topological order whenever a circuit is built in causal order, which every in-repo test does.

The new ordering is `MeasId` order. For minted ids that comes out as the order the program writes its measurements, which is what a measurement record means and the convention Stim's record offsets use. It is not the same thing as insertion order in general -- `mz_with_ids` accepts external Guppy ids that need not ascend, and there the supplier owns the numbering. The topological alternative is an artifact of how the DAG happens to be walked.

`measurements_extract_in_insertion_order_not_topological_order` pins it. On `dev` that test fails with `[1, 0]`; here it passes with `[0, 1]`.

The end-to-end DEM pipeline is unaffected: sampling a detector-bearing circuit over 50k shots gives identical detector rates and an identical mechanism count on both branches, because offset resolution lands on the same physical measurement either way.

## Uniqueness is enforced, and it found a real bug

A supplied id below the counter could previously duplicate a minted one. Ids are now tracked in a set and a repeat is rejected with a message naming the id, as is `usize::MAX`, which leaves no room for a successor.

That check failed five `pecos-zx` tests on its first run, all genuine: `build_unrolled_circuit` and `build_zx_body_circuit` replay a body segment N times by cloning its gates, so every round inherited the same `MeasId`. A replayed measurement is a new measurement -- round 2's ancilla readout is not round 1's -- so the replays now clear the source id and let the target mint its own numbering. Harmless before this PR only because the ids did not exist.

## MeasureLeaked is not minted for

The first revision included it, to avoid perpetuating the `MZ | MeasureFree` asymmetry. That was the wrong call, and the reasoning given for reversing it was also wrong: I claimed `TickCircuit` allocates `MeasureLeaked` no record. Review showed `DagCircuit -> TickCircuit` did mint for it, so an id-less `MeasureLeaked` took the id a real measurement already held, and `pauli_frame.rs` numbered it positionally while real measurements were numbered by `MeasId`.

Both now exclude it, so every site that numbers measurement records agrees: both conversions, `meas_ref`, `AssignMissingMeasIds`, the fault extractors and the Pauli frame lookup. Admitting `MeasureLeaked` across the remaining classification sites is a repo-wide sweep, not a decision for one function.

Ids *supplied* on a `MeasureLeaked` are still reserved, so the sweep only has to change which types are minted for.

## Verification

- Full workspace `cargo test`, exit 0. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings`, exit 0, on a fresh target directory. `cargo fmt --check` clean.
- Nine mutations run against the new tests, all killed: readmitting `MeasureLeaked` to minting on either side, dropping duplicate rejection, dropping overflow rejection, dropping id reservation, ignoring `MeasureLeaked` entirely, reserving ids before validating the whole gate, dropping the mint-fits precheck, and making the counter follow the latest supplied id instead of the highest.
- Every dev-versus-branch comparison used a separate `CARGO_TARGET_DIR`. A shared warm one silently links stale rlibs across worktrees and makes such comparisons meaningless.

## Not in this PR

The dual sort key in both extractors is now dead on one side but still present. Removing it changes `extract_measurements`'s public contract (`meas_ids` becomes always-populated rather than empty-for-legacy) and touches the DEM consumers, so it is its own change with its own review.

## Correctness of the ordering change, verified end to end

Independent review sampled an ordering-adversarial circuit over 50k shots: the exact DEM emits `error(0.124397) D0`, matching the analytic 2q(1-q), with sampled D0 = 0.1215 and D1 exactly 0.0000. A record-order swap would have moved that mass to D1.

Three consumers that still order measurements topologically -- `hugr_executor.rs`, `noisy_symbolic.rs`, and `pecos-neo`'s DAG command conversion -- were checked and are inert: they keep their own histories in execution-index space and never join influence-map indices, and `pecos-neo` drops `meas_ids` entirely. Inside `pecos-qec` every cross-space join is content-based.

### DIFF
--- a/crates/pecos-core/src/gate_type.rs
+++ b/crates/pecos-core/src/gate_type.rs
@@ -204,6 +204,22 @@ impl GateType {
         matches!(self, GateType::TrackedPauliMeta)
     }
 
+    /// Returns true if this gate consumes a measurement record.
+    ///
+    /// A measurement record is a slot in the classical result stream, named by a
+    /// [`MeasId`](crate::MeasId). Every site that allocates, counts, or resolves
+    /// records must agree on which gates get one, or two different measurements
+    /// end up sharing a record. Use this predicate rather than spelling the gate
+    /// types out, so there is one place to change.
+    ///
+    /// `MeasureLeaked` is a measurement to `Gate::validate` and collapses the
+    /// qubit, but it produces no classical result and so consumes no record.
+    /// Deciding otherwise means changing this function and nothing else.
+    #[must_use]
+    pub const fn consumes_measurement_record(self) -> bool {
+        matches!(self, GateType::MZ | GateType::MeasureFree)
+    }
+
     /// Returns the number of angle parameters this gate type requires
     ///
     /// # Returns
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -1690,10 +1690,11 @@ fn remap_history_onto_records(
             if let Some(&dependency_column) = column_of.get(&dependency) {
                 outcome.insert(dependency_column);
             } else {
-                let (tick, qubits) = leaked_sites
-                    .get(&dependency)
-                    .cloned()
-                    .unwrap_or((usize::MAX, Vec::new()));
+                let (tick, qubits) = leaked_sites.get(&dependency).cloned().expect(
+                    "every simulator measurement index is either a record column or a \
+                     leaked site, so a dependency outside both means the two maps were \
+                     not built from the same walk",
+                );
                 return Err(MeasurementHistoryError::LeakedMeasurementNotRepresentable {
                     tick,
                     qubits,
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -271,7 +271,7 @@ pub(crate) fn flatten_tick_circuit(tc: &TickCircuit) -> (Vec<GateLoc>, HashMap<u
                 .iter()
                 .map(pecos_core::QubitId::index)
                 .collect();
-            if is_supported_measurement_gate(gate.gate_type()) {
+            if gate.gate_type().consumes_measurement_record() {
                 meas_positions.insert(gates.len(), meas_count);
                 meas_count += 1;
             }
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -39,7 +39,7 @@ use pecos_random::{PecosRng, RngExt};
 use pecos_simulators::measurement_sampler::{MeasurementKind, SampleResult};
 use pecos_simulators::symbolic_sparse_stab::MeasurementHistory;
 use pecos_simulators::{BitmaskPauliProp, CliffordGateable};
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 
 /// Error returned when `build_fault_table` encounters an unsupported gate.
@@ -1511,13 +1511,17 @@ impl From<UnsupportedGateError> for MeasurementHistoryError {
 /// Iterates tick-by-tick to match the `TickCircuit`'s measurement numbering
 /// (which detector/DEM-output record indices reference).
 ///
-/// Errors on unsupported gates with tick/gate/qubit context (same gate set
-/// as [`build_fault_table`]).
+/// One column per measurement *record*. `MeasureLeaked` collapses its qubit but
+/// consumes no record, so it gets no column and the remaining columns are
+/// renumbered around it.
 ///
 /// # Errors
 ///
-/// Returns [`UnsupportedGateError`] when the circuit contains a gate outside
-/// the supported Clifford/prep/measurement/metadata set.
+/// Returns [`MeasurementHistoryError::UnsupportedGate`] for a gate outside the
+/// supported Clifford/prep/measurement/metadata set, and
+/// [`MeasurementHistoryError::LeakedMeasurementNotRepresentable`] when a
+/// recorded outcome depends on a `MeasureLeaked` result -- that dependency
+/// names a column this history has no way to express.
 pub fn symbolic_measurement_history(
     tc: &TickCircuit,
 ) -> Result<MeasurementHistory, MeasurementHistoryError> {
@@ -1531,6 +1535,10 @@ pub fn symbolic_measurement_history(
         .unwrap_or(0);
 
     let mut sim = SymbolicSparseStab::new(num_qubits);
+    // Simulator measurement index -> record column, for the measurements that
+    // consume a record; and the leaked ones, which consume none.
+    let mut column_of: BTreeMap<usize, usize> = BTreeMap::new();
+    let mut leaked_sites: BTreeMap<usize, (usize, Vec<usize>)> = BTreeMap::new();
 
     for (tick_idx, tick) in tc.iter_ticks() {
         for gate in tick.iter_gate_batches() {
@@ -1620,14 +1628,18 @@ pub fn symbolic_measurement_history(
                 // every reference to it. Representing that needs a notion of a
                 // hidden random source this history does not have, so say so
                 // rather than return a history whose dependencies dangle.
-                GateType::MeasureLeaked => {
-                    return Err(MeasurementHistoryError::LeakedMeasurementNotRepresentable {
-                        tick: tick_idx,
-                        qubits: qs,
-                    });
-                }
-                GateType::MZ | GateType::MeasureFree => {
-                    sim.mz(&qs);
+                GateType::MZ | GateType::MeasureFree | GateType::MeasureLeaked => {
+                    // Every measurement collapses its qubit, so all three run.
+                    // Only the record-bearing ones earn a column.
+                    let records_a_result = gate.gate_type.consumes_measurement_record();
+                    for result in sim.mz(&qs) {
+                        if records_a_result {
+                            let column = column_of.len();
+                            column_of.insert(result.index, column);
+                        } else {
+                            leaked_sites.insert(result.index, (tick_idx, qs.clone()));
+                        }
+                    }
                 }
                 GateType::I
                 | GateType::Idle
@@ -1649,7 +1661,55 @@ pub fn symbolic_measurement_history(
         }
     }
 
-    Ok(sim.measurement_history().clone())
+    if leaked_sites.is_empty() {
+        return Ok(sim.measurement_history().clone());
+    }
+    remap_history_onto_records(sim.measurement_history(), &column_of, &leaked_sites)
+}
+
+/// Renumber a measurement history onto record columns, dropping the leaked
+/// measurements that hold no record.
+///
+/// A symbolic result names its dependencies by the simulator's own measurement
+/// index, so dropping a column means every surviving reference has to be
+/// rewritten. A dependency on a *leaked* measurement cannot be rewritten at all
+/// -- it names a random source with no column -- so that is where this gives up
+/// rather than emit a history whose dependencies dangle.
+fn remap_history_onto_records(
+    history: &MeasurementHistory,
+    column_of: &BTreeMap<usize, usize>,
+    leaked_sites: &BTreeMap<usize, (usize, Vec<usize>)>,
+) -> Result<MeasurementHistory, MeasurementHistoryError> {
+    let mut remapped = MeasurementHistory::new();
+    for result in history.iter() {
+        let Some(&column) = column_of.get(&result.index) else {
+            continue;
+        };
+        let mut outcome = pecos_core::BitSet::new();
+        for dependency in &result.outcome {
+            if let Some(&dependency_column) = column_of.get(&dependency) {
+                outcome.insert(dependency_column);
+            } else {
+                let (tick, qubits) = leaked_sites
+                    .get(&dependency)
+                    .cloned()
+                    .unwrap_or((usize::MAX, Vec::new()));
+                return Err(MeasurementHistoryError::LeakedMeasurementNotRepresentable {
+                    tick,
+                    qubits,
+                });
+            }
+        }
+        remapped.push(
+            pecos_simulators::symbolic_sparse_stab::SymbolicMeasurementResult {
+                outcome,
+                flip: result.flip,
+                is_deterministic: result.is_deterministic,
+                index: column,
+            },
+        );
+    }
+    Ok(remapped)
 }
 
 fn symbolic_pairs(qs: &[usize]) -> Vec<(usize, usize)> {
@@ -2479,13 +2539,32 @@ mod tests {
         );
     }
 
-    /// A record-aligned history cannot represent a `MeasureLeaked`: it consumes
-    /// no record, so it has no column, but later outcomes can depend on its
-    /// result and those dependencies are carried as indices into the
-    /// simulator's own numbering. Dropping the column strands them, so the
-    /// circuit is refused instead.
+    /// A leaked measurement nothing depends on is simply renumbered around: it
+    /// holds no record, so the measurements that do keep contiguous columns.
     #[test]
-    fn measure_leaked_is_refused_rather_than_dropped_from_the_history() {
+    fn a_leaked_measurement_nothing_depends_on_is_renumbered_around() {
+        let mut tc = TickCircuit::new();
+        tc.tick().pz(&[0, 1]);
+        tc.tick()
+            .try_add_gate(pecos_core::Gate::measure_leaked(&[0usize]))
+            .expect("gate is valid");
+        tc.tick().mz(&[1]);
+
+        let history = symbolic_measurement_history(&tc)
+            .expect("nothing records a dependency on the leaked result");
+        assert_eq!(history.len(), 1, "only the MZ holds a record");
+        assert_eq!(
+            history.get(0).expect("one column").index,
+            0,
+            "the surviving measurement takes column 0, not the simulator's index 1"
+        );
+    }
+
+    /// When a recorded outcome *does* depend on the leaked result, there is no
+    /// column to point at, so the circuit is refused rather than handed back
+    /// with a dependency naming a column that does not exist.
+    #[test]
+    fn a_recorded_outcome_depending_on_a_leaked_result_is_refused() {
         let mut tc = TickCircuit::new();
         tc.tick().pz(&[0, 1]);
         tc.tick().h(&[0]);
@@ -2495,29 +2574,36 @@ mod tests {
             .expect("gate is valid");
         tc.tick().mz(&[1]);
 
-        let err = symbolic_measurement_history(&tc)
-            .expect_err("a leaked measurement has no place in a record-aligned history");
-        assert!(
-            matches!(
-                err,
-                MeasurementHistoryError::LeakedMeasurementNotRepresentable { .. }
-            ),
-            "expected the leaked-measurement refusal, got: {err}"
-        );
+        match symbolic_measurement_history(&tc)
+            .expect_err("the MZ outcome depends on the leaked measurement")
+        {
+            MeasurementHistoryError::LeakedMeasurementNotRepresentable { tick, qubits } => {
+                assert_eq!(tick, 3, "the error must name the leaked measurement's tick");
+                assert_eq!(qubits, vec![0], "and its qubit");
+            }
+            other @ MeasurementHistoryError::UnsupportedGate(_) => {
+                panic!("expected the leaked-dependency refusal, got: {other}")
+            }
+        }
     }
 
-    /// The same circuit without the leaked measurement is fine, so the refusal
-    /// is about the leaked measurement and not the surrounding gates.
+    /// `MeasureFree` consumes a record, so it keeps a column of its own. Nothing
+    /// else pinned that arm, so dropping it went unnoticed.
     #[test]
-    fn the_same_circuit_without_a_leaked_measurement_is_accepted() {
+    fn measure_free_keeps_its_record_column() {
         let mut tc = TickCircuit::new();
         tc.tick().pz(&[0, 1]);
         tc.tick().h(&[0]);
         tc.tick().cx(&[(0, 1)]);
+        tc.tick().mz_free(&[0]);
         tc.tick().mz(&[1]);
 
         let history = symbolic_measurement_history(&tc).expect("supported gates");
-        assert_eq!(history.len(), 1);
+        assert_eq!(
+            history.len(),
+            2,
+            "both the MeasureFree and the MZ consume a record"
+        );
     }
 
     // ---- symbolic_measurement_history tests ----
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -1464,6 +1464,42 @@ fn toggle_sorted(values: &mut Vec<usize>, value: usize) {
     }
 }
 
+/// Why a circuit cannot be turned into a record-aligned measurement history.
+#[derive(Debug, Clone)]
+pub enum MeasurementHistoryError {
+    /// A gate outside the supported Clifford/prep/measurement/metadata set.
+    UnsupportedGate(UnsupportedGateError),
+    /// The circuit contains a `MeasureLeaked`.
+    ///
+    /// It collapses its qubit and later outcomes can depend on the result, but
+    /// it consumes no measurement record, so it has no column in a
+    /// record-aligned history and its dependencies cannot be expressed without
+    /// one.
+    LeakedMeasurementNotRepresentable { tick: usize, qubits: Vec<usize> },
+}
+
+impl fmt::Display for MeasurementHistoryError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::UnsupportedGate(err) => write!(f, "{err}"),
+            Self::LeakedMeasurementNotRepresentable { tick, qubits } => write!(
+                f,
+                "MeasureLeaked at tick {tick} on qubits {qubits:?} consumes no measurement \
+                 record, so it cannot appear in a record-aligned measurement history; \
+                 remove it or use a path that does not require record alignment"
+            ),
+        }
+    }
+}
+
+impl std::error::Error for MeasurementHistoryError {}
+
+impl From<UnsupportedGateError> for MeasurementHistoryError {
+    fn from(err: UnsupportedGateError) -> Self {
+        Self::UnsupportedGate(err)
+    }
+}
+
 // ============================================================================
 // Shared symbolic simulation helper
 // ============================================================================
@@ -1484,7 +1520,7 @@ fn toggle_sorted(values: &mut Vec<usize>, value: usize) {
 /// the supported Clifford/prep/measurement/metadata set.
 pub fn symbolic_measurement_history(
     tc: &TickCircuit,
-) -> Result<MeasurementHistory, UnsupportedGateError> {
+) -> Result<MeasurementHistory, MeasurementHistoryError> {
     use pecos_simulators::SymbolicSparseStab;
 
     let num_qubits = tc
@@ -1495,7 +1531,6 @@ pub fn symbolic_measurement_history(
         .unwrap_or(0);
 
     let mut sim = SymbolicSparseStab::new(num_qubits);
-    let mut recorded = pecos_simulators::MeasurementHistory::new();
 
     for (tick_idx, tick) in tc.iter_ticks() {
         for gate in tick.iter_gate_batches() {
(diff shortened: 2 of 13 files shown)


ciaranra:
Two independent deep reviews (Fable and Sol). **Both REVISE**, and both refute this PR's central claim by running.

## The claim "no downstream behaviour change" is false

Both arms demonstrated it independently. The mechanism, confirmed by reading `propagator/dag.rs:1946` and `influence_builder.rs:494`:

```rust
if gate.meas_ids.is_empty() {
    let topo_pos = ...topo_position(node);              // topological order
} else {
    let sort_key = mr.map_or(usize::MAX, MeasId::index); // id order
}
entries.sort_by_key(...)
```

Minting ids on every DAG measurement flips that sort key for every `DagCircuit`-built circuit. Existing tests pass only because minted ids coincide with topological order whenever a circuit is built in causal order -- which every in-repo test does. `add_gate` plus a later `connect` breaks that coincidence, and Sol ran exactly that case.

Fable measured the same at the `DagCircuit -> TickCircuit` boundary: `h(&[0]); mz(&[0]); mz(&[1])` gives q1->MeasId(0), q0->MeasId(1) on `dev` (layer order) and the reverse on this branch (insertion order). `pauli_frame.rs:431` shifts the same way.

**Important mitigation, and Fable is the only one who checked:** the end-to-end DEM pipeline is empirically *unchanged*. It built a skewed circuit with detectors addressed by record offsets, sampled 50k shots on both branches, and got identical `detector_rates` and identical `num_mechanisms`. Offset resolution lands on the same physical measurement either way. So this is a real semantic change at two boundaries, not a corruption of the DEM path.

## A mutation survived: MeasureLeaked is completely untested

Fable ran five mutations against the four tests. Four are killed. **Dropping `MeasureLeaked` from the minting match passes all four.**

That is the single most contested decision in the PR and nothing pins it. Worse, Fable then found the decision is actively harmful as things stand: `TickCircuit::from(&DagCircuit)`, the tick builders, and `AssignMissingMeasIds` all treat only `MZ | MeasureFree` as record-bearing, so minting for `MeasureLeaked` here **stamps duplicate ids at the Tick->Dag boundary** (verified by running: an id-less `MeasureLeaked` gets `MeasId(1)`, colliding with an `MZ` that already holds `MeasId(1)`).

I included `MeasureLeaked` to avoid perpetuating the `MZ | MeasureFree` asymmetry. It creates a different one instead, and until the Tick side counts `MeasureLeaked` this converts a latent inconsistency into stamped duplicates.

## The duplicate-id hole I suspected

Confirmed reachable by both, and both add what I had not: **this PR does not create the downstream corruption** -- `dev` produces an identically corrupted `TickCircuit` from the same input, because its conversion has the same keep-supplied/mint-if-empty hole. What is new is that the duplicate now lives *inside* the DagCircuit, so DAG-side consumers see it where they previously did not.

Also found: `MeasId(usize::MAX)` panics on the `+1` inside an advertised fallible path, and `gate_mut` is a live escape from the choke point (no in-repo caller does it, but it bypasses minting and the counter).

## A methodology warning that touches my own earlier work

Fable discovered that the shared warm `CARGO_TARGET_DIR` **silently linked stale `pecos-quantum` rlibs across worktrees** -- cargo judged another worktree's build "fresh". It re-verified every dev-versus-branch comparison with forced rebuilds and a sentinel print confirming which code was actually linked.

I used that same shared target directory for cross-worktree comparisons throughout this issue's investigation. My probes did show differing results between branches, which argues they were not stale -- but I cannot claim they were immune, and anyone repeating them should force the rebuild.

## Where this leaves the PR

Must-fix before it can land:

1. Add a test covering `MeasureLeaked` minting, killing the surviving mutation.
2. Either exclude `MeasureLeaked` until the Tick side counts it, or align the Tick side first -- currently it stamps duplicates.
3. Amend the description to state the actual behaviour change (Dag->Tick record indices and `pauli_frame` numbering move from layer/topological order to insertion order) and why that direction is intended.
4. Reject overflowing supplied ids; document the `gate_mut` escape.

ciaranra:
CI caught something the Rust-only verification could not: `test_dem_sampler_builder_rejects_duplicate_stamped_meas_ids` built its duplicate through `mz_with_ids([0, 1], [7, 7])`, and the new uniqueness check turned that into a panic during `to_dag_circuit()`.

The check was right to fire -- those ids really are duplicates -- but a Rust panic reaches Python as `PanicException`, which derives from `BaseException` and so escapes ordinary `except Exception` handling. Loud, but not catchable, which is the wrong shape for input a caller controls.

Fixed at the two boundaries where the ids arrive:

- `mz_with_ids` rejects a repeated id within one call, naming the id.
- `to_dag_circuit` scans for duplicates before converting and raises `ValueError`. The conversion itself goes through an infallible `From` impl and cannot report, so the check sits in front of it.

The `DagCircuit` panic stays as the internal backstop for callers that bypass the bindings.

Two tests moved as a result. `test_dem_sampler_builder_rejects_duplicate_stamped_meas_ids` became two tests covering the within-call and across-call routes at their new failure points, and `test_result_tag_remap_validation_rejects_duplicate_traced_meas_ids` now supplies its duplicate directly rather than through the binding, matching the fake-circuit pattern already used by the test beside it. The sampler's own `reject_duplicate_stamped_meas_ids` guard is unchanged and still covered by `test_validate_measurement_count_rejects_duplicate_stamped_meas_id`, which builds the influence map directly.

Also stripped issue-tracking prose out of the code comments in this branch, and out of three comments left by earlier PRs in the files it touches. Comments should describe the code, not the backlog.

Verified: fast Python lane 4120 passed / 56 skipped, exit 0. Full workspace `cargo test` exit 0. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `pre-commit run --all-files` exit 0.

ciaranra:
Both reviews returned REVISE. Every finding below was checked against the code before being acted on; fixes are in `2c1cda6cd`.

## Confirmed and fixed

**Id assignment was not transactional.** Ids were reserved as they were checked, so a rejected gate burned the ones ahead of the failure -- a `[3, 3]` batch left `3` taken by a gate that never entered the circuit, and a rejected `usize::MAX` stayed reserved too. `assign_measurement_ids` now validates the whole gate before touching any state.

**Overflow was reachable, and not where I first fixed it.** A supplied `usize::MAX - 1` left the counter at the ceiling and the next mint did an unchecked `+= 1`. I initially patched only the `DagCircuit` side and removed a `usize::MAX` guard from `to_dag_circuit`, arguing no producer could reach it. That argument was wrong: `mz_with_ids([0], [2**64-2])` followed by a plain `mz` panics, and the overflow is in the `TickCircuit` record counter, before `DagCircuit` is involved.

`TickCircuit::try_advance_meas_counter` now reserves records with checked arithmetic, and all five allocation sites route through it -- `TickHandle::mz`, `TickHandle::mz_free`, and the three Python measurement paths. The last representable id is never handed out, so the counter always has a valid successor.

**`MeasureLeaked` was inconsistent in two more places than I found.** `DagCircuit -> TickCircuit` minted for it while `DagCircuit` did not, so an id-less `MeasureLeaked` took the id a real measurement already held and converting back reported a duplicate. Separately, `pauli_frame.rs:415` treated it as record-consuming, numbering it positionally while real measurements were numbered by `MeasId` -- two measurements mapping to record 0. Both now exclude it, matching the other numbering sites. The comment claiming they already agreed was false and has been corrected.

**The "insertion order" claim was overstated.** Extraction sorts by `MeasId` value, which equals insertion order only for minted ids; `mz_with_ids` accepts external ids that need not ascend. The test is renamed to say what it checks, and a second test pins that supplied ids drive the order even when they run counter to the order of addition.

**A mutation survived.** `next_meas_id.max(successor)` -> `successor` left the whole suite green. `a_lower_supplied_id_does_not_rewind_the_counter` kills it.

## Judgments made against a recommendation

**`used_meas_ids` memory -- not optimising now.** The two measurements disagree: one reports ~19.6 bytes per entry and ~35 MB for a distance-31, 1000-round circuit; the other timed the actual path at 1.110s vs 1.054s for a million insertions, ~5%, with a realistic DEM circuit at ~10^4 measurements costing microseconds and well under 1 MB. The second measures this code rather than a standalone probe, at a scale the repo actually builds, so the lazy-allocation scheme is deferred rather than added to a function that just had three defects fixed in it. Both numbers are recorded on #387.

**`gate_mut` -- documented, not closed.** One review rated this HIGH and asked for it to be made crate-private. All four in-repo callers were verified not to touch measurement ids, and `dag_circuit.rs:599` shows `gate_mut` could already desync `max_qubit` before this PR, so it is a pre-existing unsound API rather than something this change introduced. `gate_mut` and `remove_gate` now document the hazard; closing it is an API change affecting a `pecos-qec` caller and belongs on its own.

## Not done

Both reviews want a fallible `TryFrom<&TickCircuit> for DagCircuit` so the duplicated uniqueness check in `to_dag_circuit` can be deleted rather than kept in step with the core. That is step 3 of #387's plan and is not in this PR. The pre-scan remains a second implementation of the invariant until then, and its comment says so.

## Verification

Full workspace `cargo test` exit 0 (304 suites). Fast Python lane 4122 passed / 56 skipped, exit 0. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory -- it caught two missing `# Panics` sections and a redundant closure that `pre-commit` passed over. `cargo fmt --check` and `pre-commit run --all-files` exit 0.

Nine mutations run against the new tests, all killed: readmitting `MeasureLeaked` to minting on either side, dropping duplicate rejection, dropping overflow rejection, dropping id reservation, ignoring `MeasureLeaked` entirely, reserving before validating, dropping the mint-fits precheck, and the counter-follows-latest mutation above.

One process note: clippy also caught that a misplaced `#[test]` attribute had made a new test run twice while silently disabling `test_basic_tick_circuit`. The doubled test name was visible in my own earlier output and I did not question it.

ciaranra:
Third round, both arms REVISE. Ten findings between them; every one verified against the code before acting, and all fixed in `fc8e19563` except one deliberately scoped out below.

My prediction that a third round would be low-value was wrong. It found real defects, including two I introduced while fixing the second round.

## Record allocation had three more holes

`AssignMissingMeasIds` (`pass.rs`) still did an unchecked `next_id += 1`, reachable from Python. `TickHandle::mz` and `mz_free` reserved one record *per qubit*, so a two-qubit measurement near the ceiling reserved the first and panicked on the second. And the Python bindings still reserved *before* the fallible gate insertion, so `tick.mz([0, 0])` raised `ValueError` and left the counter advanced by 2 -- which is item 6 on #387, documented and then walked past twice while editing those exact lines.

All three now reserve the whole batch through `try_advance_meas_counter`, and the bindings commit only after the gate is in. Empty `mz_with_ids` no longer reserves a phantom record for a measurement that does not exist.

## Two surviving mutants, in the fixes for surviving mutants

`highest_successor.max(successor)` -> `= successor` passed everything: a single batch supplying `[10, 3]` left the counter at 4, so the seventh later mint silently re-issued `MeasId(10)`. The round-two test used two separate gates, never one batch with descending ids.

Both `MeasureFree` legs of the rewritten predicates were also unkilled -- nothing exercised `MeasureFree` through record mapping or the conversion.

## A regression I introduced, and the reason it happened

`pauli_frame.rs` had its `clear_qubit` call *inside* the record lookup. Excluding `MeasureLeaked` from the record map therefore also stopped it collapsing the qubit, so a Pauli propagated straight through a leaked measurement -- disagreeing with `fault_sampler`, which does clear. Changing record numbering silently changed propagation physics. Clearing is now unconditional, which is what a measurement does whether or not it produces a record.

And a fourth `MeasureLeaked` site: `fault_sampler.rs` numbered it positionally while detector records number by `MeasId`. That is the second time I claimed "every site excludes it" and was wrong.

## The fix for that class, rather than a fifth site

`GateType::consumes_measurement_record()` now lives in `pecos-core`, and all seven record-numbering sites call it instead of spelling out `MZ | MeasureFree`. Mutating that one function fails tests in both `pecos-quantum` and `pecos-qec`. The invariant is greppable rather than remembered, which is what #387's "suggested shape" asked for.

## The ordering test was vacuous

`extraction_follows_supplied_ids_not_the_order_they_were_added` passed with the production sort key mutated to topological position -- the fixture's topological order coincided with its id order, so it never tested its own mechanism. Rebuilt with three measurements whose insertion, topological and id orders are pairwise distinct; it now fails under that mutation. Two stale doc blocks corrected, including `extract_measurements` still documenting the topological rule it no longer uses.

## Scoped out, not fixed

`add_gate("MZ", [0])` creates a measurement without reserving a record, so an id-less measurement and a stamped one collide on an entirely mundane circuit, and `to_dag_circuit`'s prescan cannot see it because it only inspects *supplied* ids. The real fix is the fallible `TryFrom<&TickCircuit>` both reviewers have now asked for three rounds running. I am not adding a fourth partial guard in front of a conversion that cannot report failure -- that is the pattern that has already failed here repeatedly. Recorded on #387 as a reachable panic with its reproduction, alongside the note that legal sparse ids meet dense `vec![...; num_measurements]` allocations.

## Verification

Full workspace `cargo test` exit 0 (304 suites). Python lane 4125 passed / 56 skipped, exit 0. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory -- it caught three missing `# Panics` sections that `pre-commit` passed over, the second time in this branch that cold clippy has failed where `pre-commit` did not.

Seven mutants killed this round, each confirmed by running: the descending-batch counter, both `MeasureFree` predicate legs, readmitting `MeasureLeaked`, the per-qubit reservation, the empty-input reservation, and reserving before insertion.

ciaranra:
Fourth round, scoped to `2c1cda6cd..fc8e19563` only. Both arms REVISE. Fixes in `a6dbdb8a9`.

This round existed to close one specific gap: the end-to-end DEM verification from round three was run at `2c1cda6cd`, and the delta had since changed the two things that bear on DEM correctness -- which gates receive positional measurement indices, and whether a measurement clears a propagating Pauli. That evidence was stale for exactly the files that changed.

## The DEM pipeline is correct at this head

Independently re-verified by running: a circuit with detectors addressed by record offsets, sampled 200k shots, gives D0 = 0.2194 against analytic 0.2200 (0.6 sigma) and D1 = 0.2759 against 0.2760 (0.1 sigma). The exact DEM mechanism table is **bit-identical with and without a `MeasureLeaked` in the circuit**, which is what the record-numbering fix was supposed to achieve. Pauli-frame records follow `MeasId` rather than insertion order on a swapped-id adversarial circuit.

## Both arms found the same bug, and it was mine

Round three fixed `fault_sampler`'s record numbering to exclude `MeasureLeaked` but left `symbolic_measurement_history` still giving it a column. So the history had one more column than the fault mechanisms' record space. One arm measured the detector marginals effectively swapped; the other measured the leaked column reading 0.2759 while the real measurement's column read 0.0000. Fixing the external mismatch created an internal one.

All three measurement types still run on the simulator, because every measurement collapses its qubit. Only record-bearing ones take a history column, so the columns line up with the record indices detectors reference.

`try_mz` and `try_mz_free` were also not batch-atomic: with one id slot left, a two-qubit batch inserted the first measurement, wired it, then returned `Err`. They now check the whole batch fits before inserting any of it -- the same standard the `TickHandle` fix set in the previous round.

## Both fixes from the previous round were untested

Reverting the `pauli_frame` clear to its old position, and reverting `fault_sampler`'s predicate, each left the entire `pecos-qec` suite green. Two behaviour fixes shipped with nothing pinning either. Both now have tests, and both mutations now fail.

## The arms disagreed on physics

One held that clearing the propagating Pauli at a non-destructive measurement is wrong -- collapse projects onto an eigenstate, it does not reset to the ideal state, so an earlier X still flips a later repeated measurement. The other held the collapse claim was right.

Resolved in favour of the first, on evidence rather than argument: `StateVec` gives `{0: 1}` then `{0: 1}` for `X`, `MZ`, `MZ` -- the error persists through the measurement. And `quantum.rs:205` executes `MeasureLeaked` as a plain `MZ` with no reset, so whatever holds for `MZ` holds for it.

**That clearing is pre-existing, not introduced here.** `origin/dev` already had `MeasureLeaked` in the record map, so `clear_qubit` already ran for all three types; this branch preserves that behaviour exactly rather than changing it. Correcting it changes fault propagation for every circuit with a repeated measurement on one qubit and needs validating against Stim, so it is filed separately rather than slipped into a PR about id minting. It has stayed latent because surface-code ancillas are re-prepared each round and `PZ` clears legitimately.

The new test pins `MeasureLeaked` and `MZ` against **each other** rather than against an absolute, so it stays valid whichever way that question is settled.

## Also confirmed clean by the reviews

Six of the seven sites routed through the shared predicate are exactly equivalent to their old spelled-out conditions; the seventh is `fault_sampler`, deliberately non-equivalent and correct against detector metadata. `try_add_gate_auto_wire`'s error path leaves no partial wiring and does not change `qubit_heads` or `last_node`. No path exists where a Python binding inserts a gate and then fails the counter commit. `AssignMissingMeasIds` pre-counts and assigns over the same gates in the same order, so it reserves exactly what it hands out.

## Verification

Full workspace `cargo test` exit 0 (304 suites). Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `cargo fmt --check` clean. Python lane 4125 passed / 56 skipped, exit 0. `pre-commit run --all-files` exit 0.

One process note against myself: an earlier run of this verification exited 1 and my grep filter did not match the error, so it printed nothing. The cause was mine -- I had a background workspace test and foreground mutation runs sharing one `CARGO_TARGET_DIR`, and they raced over an rlib. The result above is from a run with nothing else touching that directory, and I read its exit status directly rather than inferring it from a filter.

ciaranra:
A final narrow check of `fc8e19563..a6dbdb8a9` -- 171 lines, mostly tests -- returned REVISE and found a runtime panic. Fixed in `50f6278c7`.

I had recommended merging on the grounds that this delta was "small, mechanical, and test-covered". That was wrong, and wrong for a reason worth recording: I judged it by line count rather than by what it touched.

## The history fix was broken

`a6dbdb8a9` made `symbolic_measurement_history` build a fresh `MeasurementHistory` containing only record-bearing measurements, so its columns would line up with the record indices detectors reference.

That cannot work. `SymbolicMeasurementResult` carries `index` and an `outcome` bitset of *other measurements' indices*, both in the simulator's own numbering. Filtering the results leaves those references pointing at columns that no longer exist. Reproduced: `PZ(0,1); H(0); CX(0,1); MeasureLeaked(0); MZ(1)` produced a one-column history whose only entry had `index == 1` and a dependency on column 0, and `RawMeasurementPlan::sample` panicked.

Renumbering would not fix it either. A filtered-out `MeasureLeaked` can be the hidden random source a retained measurement depends on, and a record-aligned history has no way to express a source with no column.

So the function now refuses. `MeasurementHistoryError::LeakedMeasurementNotRepresentable` says why, rather than returning a history whose dependencies dangle.

I checked the alternatives before choosing that. Reverting the filter reinstates the misalignment both arms found in round four. Reverting `fault_sampler.rs:274` as well makes that file self-consistent again but reinstates a mismatch with the record space that *this branch* causes, since `MeasureLeaked` no longer consumes a record. Refusing is the only option that yields no wrong answers anywhere. `build_fault_table` is unaffected -- a leaked measurement is still a valid fault location there, it just holds no record.

## Two more of my tests were vacuous

`measure_leaked_collapses_its_qubit_without_taking_a_history_column` did not test collapse: its fixture measured a qubit immediately after `PZ`, so removing `MeasureLeaked` from the simulator arm entirely still passed. And the batch-atomicity test exercised `try_mz` but never `try_mz_free`, so deleting that function's pre-check went unnoticed.

Both replaced. The first is superseded by a refusal test plus a control showing the same circuit without the leaked measurement is accepted; the second now covers both entry points.

I mutation-checked both of these when I wrote them and still missed it, because I mutated the guard each test *names* rather than every guard it *claims* to cover.

## Confirmed sound by the check

`reserve_room_for_mints` is exactly the condition the per-gate mint path enforces, nothing between the check and the inserts can alter the counter, and `count == 0` is handled. The `MeasureLeaked`-versus-`MZ` comparator is sound -- its meta node, topological position and record lookup are what the test assumes -- and it correctly keeps passing when clearing is removed from both compared types, so it will survive #394 rather than pinning behaviour that is known wrong.

## Verification

Full workspace `cargo test` exit 0 (304 suites), run with nothing else touching the target directory. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory -- it caught a wildcard match on the new error enum that would have silently swallowed any variant added later. `cargo fmt --check` clean. Python lane 4125 passed / 56 skipped exit 0. `pre-commit run --all-files` exit 0.

Both new guards mutation-killed: dropping the `try_mz_free` pre-check, and accepting `MeasureLeaked` into the history again.

The round-four end-to-end DEM verification still stands -- that path contains no `MeasureLeaked`, and the refusal only fires on circuits that do.

ciaranra:
Both arms reviewed `a6dbdb8a9..50f6278c7`. Both REVISE, and they **disagreed on the central question**. Fixes in `2d7527699`.

## The disagreement, and why both were right

I asked both whether refusing was the right call or whether I had traded a panic for a worse regression.

One arm said the blanket refusal regresses circuits that worked: `PZ(0,1); MeasureLeaked(0); MZ(1)` has a retained measurement that is independently deterministic, so the previous commit's filtering sampled it correctly, and refusing now returns `None` through `dem_simulator.rs`'s `.ok()?`.

The other said refusing is correct and no caller loses a working path: through these consumers a leaked circuit was **silently wrong before this branch** -- the leaked column shifted every fault flip -- so a typed error beats both that and the panic.

They used different baselines. The first compared against `a6dbdb8a9`, my own broken intermediate. The second compared against `dev`. Against `dev`, refusing is strictly better. But the first arm's case is still real: that circuit *is* representable, and refusing it blocks leakage-aware sampling for no reason.

So I did neither. `symbolic_measurement_history` now renumbers the surviving measurements onto record columns and refuses **only** when a recorded outcome actually depends on a leaked result -- the one case with no column to point at. The representable subset works; the unrepresentable one fails loudly and names the leaked measurement's tick and qubits.

## Two mutants survived the review

**`MeasureFree` had no coverage at all through this path.** Deleting it from the arm passed the entire `pecos-qec` and `pecos-eeg` suites. That is the arm I had rewritten twice, and a regression there would silently reintroduce the record-misalignment this branch exists to fix.

**The refusal test ignored its own error payload.** It matched `{ .. }`, so constructing the error with `tick: 999_999, qubits: vec![]` passed.

Both dead now, along with two more I added for the new paths: refusing whenever any leaked measurement exists, and skipping the renumbering.

## Confirmed clean by the reviews

The non-leaked path was verified textually identical to the pre-`a6dbdb8a9` implementation -- no partial version of the broken filtering left behind. The refusal arm cannot be bypassed by batched gates, mixed-type ticks, zero- or multi-qubit leaked gates, or `MeasureLeaked` reached through custom-gate decomposition, which parses to the concrete enum and lands on the same arm. `From<UnsupportedGateError>` is unused in-repo but reasonable public conversion; dropping `PartialEq`/`Eq` regresses nothing since `UnsupportedGateError` never had them.

## Verification

Full workspace `cargo test` exit 0 (304 suites), nothing else touching the target directory. Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory -- it caught two lints the suite passed over, a wildcard match that would have swallowed future enum variants and an `explicit_iter_loop` in the new remap. `cargo fmt --check` clean. Python lane 4125 passed / 56 skipped exit 0. `pre-commit run --all-files` exit 0.

## Two things I got wrong again

I repeated an index-splice edit that duplicates a block by anchoring on a marker positioned before the insertion point. Same mistake as earlier in this branch; caught by the compiler both times, which is luck rather than method. I have stopped using that pattern for these edits.

And cold clippy has now failed four times in this branch where the full suite passed, every time in code written in response to a review finding. Those edits go in fast on the assumption the reviewer already established what correct looks like, and they skip the lint pass I would otherwise do while writing.

## Note on the end-to-end evidence

One arm re-ran the DEM verification at `50f6278c7` and it passed analytically (0.09488 against 0.095 at 200k shots, plus a zero-noise XOR of exactly 0). I then changed the fault-sampling path underneath it, so that number does not describe the current head and I am not presenting it as though it does. The new logic is mutation-covered on both branches, but an end-to-end figure for what actually merges would need a run at the merge commit.

ciaranra:
## End-to-end verification at `2d7527699`

Run against the intended merge commit, with nothing further queued, so this result describes what would actually land. Previous end-to-end runs were invalidated three times by my own subsequent fixes; this one is not.

**Verdict: the DEM pipeline is correct at `2d7527699`. No correctness defect in the code this PR introduced.**

### Ordering-adversarial end-to-end

Three qubits with records deliberately out of qubit order (`rec0=q2, rec1=q0, rec2=q1`), entangling gates placed so faults spread across records, detectors addressed by record offsets (`D0=rec[-3]`, `D1=rec[-2]^rec[-1]`), noise `p1=0.06 p2=0.12 p_meas=0.10 p_prep=0.05`.

- Exact DEM: the three analytically predicted footprints, matching hand-derived values to full print precision -- `{D0}` 0.163825 vs 0.1638245, `{D0,D1}` 0.079781 vs 0.0797807, `{D1}` 0.23942 vs 0.2394204. Per-channel isolation exact as well.
- Sampled at 200k shots against exact analytics (D0 0.217472, D1 0.281009): seed 7 gives D0 -0.41 sigma, D1 -2.05 sigma; a second seed gives +0.52 and +1.03 sigma.
- Zero-noise control: record columns exactly `(0, 200000, 200000)`, D1 XOR exactly 0.

### The three paths this commit created

- **Leaked measurement nothing depends on** -- now succeeds, which is precisely what the previous commit wrongly refused. History length 2, columns contiguous at 0 and 1, sampled rates within 1.0 sigma. A dependency crossing the leaked measurement is correctly renumbered from simulator index to record column, with zero-noise XOR exactly 0.
- **Recorded outcome depending on a leaked result** -- refused with `LeakedMeasurementNotRepresentable { tick: 3, qubits: [0] }`, naming the leaked site rather than producing wrong numbers.
- **Mixed `MZ` / `mz_free` / `MeasureLeaked`** -- two leaked sites interleaved among three surviving records, contiguous columns 0..2, correct deterministic flips, zero-noise columns exactly `(0,1,1)`.

Structurally confirmed that `flatten_tick_circuit` assigns fault-flip indices only to record-consuming gates, so `build_fault_table` mechanism indices and the remapped history agree on record columns.

`cargo test -p pecos-qec` 766 passed, 0 failed, run cold in a private target directory.

### Observations, neither a defect in new code

1. A pre-existing convention split: `DemBuilder` models depolarizing channels as n independent channels with `q = 1-(1-p)^(1/n)`, while `build_fault_table`/`RawMeasurementPlan` uses mutually-exclusive 1-of-k alternatives. DEM entry probabilities therefore differ from sampler-side group sums at O(p^2/n), while implied detector marginals still agree to ~2e-6 even at the large probabilities used here. Not introduced by this PR. Filed separately.

2. `remap_history_onto_records` uses `unwrap_or((usize::MAX, Vec::new()))` for a dependency present in neither map -- unreachable by construction, since every simulator measurement index lands in exactly one of them. Cosmetic, but it is a silent fallback where an assertion would state the invariant.

ciaranra:
Observation 2 from the verification run is closed in `d946b9c63`.

`remap_history_onto_records` used `unwrap_or((usize::MAX, Vec::new()))` for a dependency present in neither the record-column map nor the leaked-site map. That branch is unreachable by construction -- every simulator measurement index lands in exactly one of the two -- so it could only ever have shaped an error message on an impossible path. It now states the invariant with an `expect` instead:

> every simulator measurement index is either a record column or a leaked site, so a dependency outside both means the two maps were not built from the same walk

Worth recording why it was there: this project treats fallbacks as a smell and asks that they be discussed before being added, and I added one without doing that. It took an external reviewer to notice. The code was written quickly in response to a review finding, which is the same context in which four cold-clippy failures on this branch originated -- edits made on the assumption that the reviewer had already established what correct looks like, skipping the pass I would otherwise do while writing.

The change cannot affect any reachable path, so the end-to-end verification posted above still describes this code. I re-ran the full set regardless rather than relying on that argument: cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory, `cargo fmt --check` clean, full workspace `cargo test` exit 0 across 304 suites, `pecos-qec` 609 lib tests green.

I have made exactly this "provably unreachable" argument before on this branch -- to justify removing a `usize::MAX` guard from `to_dag_circuit` -- and a reviewer then produced a two-call sequence that reached it. Hence re-running rather than reasoning.

ciaranra:
`origin/dev` merged in (`24b846366`), bringing #392 (RNG seeding). Clean merge, no conflicts. Two corrections to the record while I am here.

## The end-to-end verification describes an earlier commit

The verification posted above was run at `2d7527699` and says it describes "this code". The head has since moved twice -- `d946b9c63` (the invariant assertion) and this merge. Neither touches the DEM path, and the verification's conclusion still holds for it, but the commit under test is no longer the head and I should not leave that claim standing unqualified. The whole point of that run was that it described the exact merge commit.

## My workspace test command was not the one this project uses

`cargo test --workspace` is not what CI runs. `pecos-cli/src/cli/rust_cmd.rs` excludes six crates from workspace tests -- `pecos-rslib`, `pecos-rslib-cuda`, `pecos-julia-ffi`, `pecos-go-ffi`, plus `pecos-rslib-exp` and `pecos-rslib-llvm` -- and the comment states exactly why: they are pyo3 cdylibs whose `extension-module` feature is opt-in, so a workspace test run tries to link the test binary against libpython and fails.

Every "full workspace `cargo test` exit 0" I have reported on this PR came from that non-CI invocation. It passed only because `pecos-rslib`'s test target was cached from a maturin build and never relinked. This merge changed `pecos_random_bindings.rs`, forced a rebuild, and it failed immediately on `undefined symbol: PyExc_OverflowError` -- nothing to do with correctness.

Re-run with the exclusion set the CLI actually applies: **297 suites, exit 0**. The 297-vs-304 difference is the excluded crates' targets, and those pyo3 cdylibs have no Rust unit tests of their own, which is the stated reason for excluding them. So no coverage was lost, but the results were produced by a command this repo does not use and were one cache invalidation away from failing for unrelated reasons. I should have checked how CI runs tests before assuming `--workspace` was the obvious thing.

## Verification on the merged tree

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `cargo fmt --check` clean. `cargo test` with the CI exclusions exit 0, 297 suites. Python lane 4129 passed / 56 skipped exit 0 -- four more than before, matching the tests #392 adds to `test_rng.py`. `pre-commit run --all-files` exit 0.

One discarded result worth noting: a Python run that was in flight when I performed the merge reported green, but its extension build predated the merge while its test collection picked up post-merge files. That is a torn read describing neither state, so it is not evidence. The numbers above come from a run started after the merge completed.

Issue #332, opened 2026-06-25
Program RNG platform functions (incl. RNGadvance) need a canonical home in the pecos-neo path, not the retiring engines


ciaranra:
### Context

PR #330 adds `RNGadvance(delta)` (relative stream movement / rewind) to the Python `RNGModel` (`python/quantum-pecos/src/pecos/engines/cvm/rng_model.py`). The logic is correct and well-tested, but that `RNGModel` is reached only by the legacy `hybrid_engine_old.py`. We're treating #330 as a **contained fix for that legacy workflow**.

The broader picture: **program-level RNG platform functions** (`RNGseed`, `RNGbound`, `RNGindex`, `RNGnum`, `RNGadvance`) are fragmented and mostly absent from the newer workflows.

| Workflow | Classical interpreter | RNG platform fns | `RNGadvance` |
|---|---|---|---|
| `hybrid_engine_old.py` (legacy) | Python cvm `RNGModel` | seed/bound/index/num | yes (#330) |
| `hybrid_engine.py` | `RustPhirClassicalInterpreter` (`pecos-phir-json`) | none | no |
| `sim()` — QASM | `pecos-qasm` engine | seed/bound/index/num | no |
| `sim()` — PHIR/HUGR/Guppy | `pecos-phir-json` / `pecos-hugr` | none | no |
| `sim_neo()` | `pecos_rslib_exp` / pecos-neo | none | no |

(`.seed()` on `sim()`/`sim_neo()` is the Monte-Carlo *noise sampling* seed — a separate mechanism from program RNG.)

### Strategic direction

Per the pecos-neo default transition (strangler-fig: route `sim()` to pecos-neo, then collapse the other workflows onto it; see #151 for the canonical Rust `RNGModel`), we should **not** scatter RNG implementations across `pecos-qasm` / `pecos-phir-json` / `pecos-hugr` — those classical paths are on the retirement track. Investing there is throwaway work.

Instead: program RNG (including `RNGadvance`) should land **once**, in the canonical Rust `RNGModel` (#151) consumed by the pecos-neo classical path, so it's there when the other workflows migrate.

### Proposed work

1. Establish program RNG as a tracked item in the neo transition (sibling to the existing G-series gaps), with the Rust `RNGModel` as the single source of truth.
[...]

PR #392, opened 2026-07-29
Random seeding all 64 bit values

pschow-qtm:
See:  https://cqc.atlassian.net/browse/QTM-2277

Fixed so that underlying support for `___random_seed() `accepts all 64-bit values, including the top (signed) bit set.

### DIFF
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -52,6 +52,12 @@ struct ParsedObservable {
     id: u32,
     records: Vec<i32>,
     meas_ids: Vec<usize>,
+    /// Human-readable name from the metadata JSON's `label` field.
+    ///
+    /// The metadata format has always carried this and callers already write
+    /// it, but it used to be parsed and dropped, leaving circuit annotations as
+    /// the only way to get a label onto an observable.
+    label: Option<String>,
 }
 
 // ============================================================================
@@ -194,6 +200,15 @@ impl<'a> DemBuilder<'a> {
     /// # Errors
     ///
     /// Returns an error if detector or observable metadata is malformed.
+    ///
+    /// # Panics
+    ///
+    /// Malformed *metadata* is reported as an error, but a circuit whose
+    /// metadata and annotations **contradict each other** is not a parse
+    /// failure and is not reported that way: two definitions of one observable
+    /// disagreeing about the label or the Pauli panic inside
+    /// [`DetectorErrorModel::add_observable`] rather than returning here. Both
+    /// sources must agree, or only one should supply the field.
     pub fn try_from_circuit(
         circuit: &pecos_quantum::DagCircuit,
         p1: f64,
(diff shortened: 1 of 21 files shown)


ciaranra:
Versions should be left at "0.9.0.dev2"and then will do a separate bump PR.

Looks like maybe there are still linting issues picked up by the CI.

I am happy to fight with the linter, if you want. But no worries if you want to push ahead.

pschow-qtm:
> Versions should be left at "0.9.0.dev2"and then will do a separate bump PR.
> 
> Looks like maybe there are still linting issues picked up by the CI.
> 
> I am happy to fight with the linter, if you want. But no worries if you want to push ahead.

Ok, I will revert the versions.  Yes, I changed them to make the lint run happy

ciaranra:
Thanks — I checked out the branch and verified the fix end-to-end (debug-mode Rust tests, plus the Python path: `srandom(-1)` produces a stream byte-identical to `srandom(2**64 - 1)`, and the QASM engine seed path is covered by the shared-core change). The mechanism looks right. Three asks before merge:

1. **The CI failure is a single clippy lint.** Both rust-lint and python-lint fail on `cast_sign_loss` at `pecos_random_bindings.rs:66` (`signed as u64`). Clippy's own suggestion is the proper fix — replace the cast with `signed.cast_unsigned()` — no `#[allow]` needed.

2. **Please add a Python-level test for the negative-seed path.** The original bug was pyo3 rejecting negative Python ints at the `u64` conversion boundary, but the new Rust unit tests call `srandom(i128)` directly from Rust, so the Python-int conversion layer that was actually broken is still untested. Something like: assert `RNGModel.set_seed(-1)` (or `RngPcg.srandom(-1)`) succeeds and yields the same stream as seed `2**64 - 1`. I verified this equivalence holds on the branch.

3. **Minor:** `test_negative_values_are_rejected_for_non_advance_rng_funcs` in `test_rng.py` now has a stale name and docstring ("Verifies only RNGadvance accepts negative numeric arguments") — RNGseed accepts negatives too now.

ciaranra:
The points are minor. I'll work on them and then if all the CI passes, I'll merge.

ciaranra:
I pushed the three items above directly to this branch to move things along:

- `0b94131b5` — replaced the `signed as u64` cast with `i64::cast_unsigned` to clear the `cast_sign_loss` clippy error that was failing rust-lint and python-lint.
- `c1a8db01c` — added Python-level tests covering the pyo3 conversion layer (negative seed equals its two's-complement unsigned seed, the 64-bit extremes are accepted, out-of-range seeds raise `OverflowError`), and updated the stale name/docstring of the negative-args test. The new tests were mutation-tested: re-introducing the old negative-seed rejection fails the acceptance tests, and a truncating accept-everything variant fails the range test.

CI should go green now. Thanks for the fix!

ciaranra:
Ran an independent adversarial review of the full diff before merge. Outcome:

- **Fixed** (`3d9dcff71`): `RNGModel.set_seed` updated `self.seed` before calling `srandom`, so a failed (out-of-range) reseed left a stale `seed` attribute while the stream stayed on the old seed. The assignment now happens only after `srandom` succeeds, and a test pins the full state (seed, count, stream position) across a failed reseed.
- **Fixed** (`3d9dcff71`): the new seed tests now assert against golden vectors computed from an independent Python implementation of the PCG32 (setseq/XSH-RR) reference — this pins both the two's-complement mapping for negative seeds and that existing positive-seed streams are unchanged, rather than only checking self-consistency.
- **Out of scope, noting for the record**: the QASM engine's `RNGseed` only accepts integer literal expressions, so a negative literal seed in a QASM program is still rejected there (negative literals parse as unary expressions). That path cannot currently receive two's-complement seeds and is part of the known RNG platform-function fragmentation tracked in #332; it should be resolved when the canonical Rust RNGModel lands rather than patched here. Similarly, `pecos.random.seed()` (numpy drop-in) and the sim builder `.seed()` keep their unsigned u64 contracts deliberately.

ciaranra (approved):
Looks good. Thanks!

PR #393, opened 2026-07-29
Expose public program-to-QIS tracing APIs

qciaran:
## Summary
- add a public `pecos.tracing` module and top-level APIs for capturing QIS operation traces, replaying saved traces into `TickCircuit`, and tracing supported programs directly to `TickCircuit`
- use input-generic naming so Guppy, HUGR, direct QIS, and future QIS-lowering frontends share the same API
- move trace framing, validation, provenance extraction, hosted-operation checks, and `TickCircuit` replay into a generic `pecos._qis_trace_replay` module with no QEC or surface-code dependency
- move runtime-traced circuit normalization and measurement-ID traversal out of `pecos.qec.surface` into the generic private `pecos._traced_circuit` module; DEM no longer imports these helpers through the surface package
- preserve runtime-emitted scheduled `Idle` operations as integer nanosecond `TimeUnits`, without synthesizing idles for runtimes that do not emit them
- validate source/runtime measurement provenance and parameterized-gate arity when replaying stored or custom-processed traces
- keep the former surface-code tracing entry-point names as compatibility aliases and route DEM construction through the generic namespace
- add a dedicated Runtime QIS Tracing tutorial covering capture, inspection, JSON persistence, replay, runtime selection, explicit idles, and single-path limitations

## Public API
- `pecos.capture_qis_operation_trace(...)`
- `pecos.qis_operation_trace_to_tick_circuit(...)`
- `pecos.trace_program_to_tick_circuit(...)`

The extracted normalization and measurement-ID helpers remain private implementation details for now.

## Validation
- `ruff check` on changed Python sources and tests
- public tracing API tests: 12 passed, including provenance and gate-arity regressions
- generic tracing and hosted-operation tests: 20 passed
- focused replay, framing, idle, and DEM integration tests: 28 passed
- extracted-helper tests: 8 passed (Clifford normalization, non-sequential measurement IDs, and result-tag remapping)
- direct-QIS hands-on walkthrough covered capture, JSON round-trip, replay, convenience conversion, and measurement provenance
- documentation test generation recognizes the new tutorial and produces one executable test
- MkDocs build completed successfully with the tutorial in navigation and cross-links
- the Guppy tutorial example reaches capture locally but hits the LLVM mismatch described below; the equivalent Guppy documentation path passes in CI's supported LLVM environment

## Local environment note
The local `llvm-as` rejects opaque-pointer IR emitted by current Guppy. Direct-QIS capture passes locally, and Guppy end-to-end documentation tests run in CI's supported LLVM 21 environment.


ciaranra:
## Review

Solid PR overall. The move-and-harden of the trace replay machinery into generic modules is well executed, the new validation is the right kind of fail-loud, and the public API surface is sensible. One real breakage must be fixed before merge; the rest is minor.

### Must fix: three example scripts break with ImportError (verified by execution)

`normalize_traced_qis_tick_circuit` is deleted from `pecos/qec/surface/circuit_builder.py` with no compatibility alias, but three example scripts still import it from there:

- `examples/surface/dem_decomposition_diagnostics.py:537`
- `examples/surface/graphlike_dem_projection_benchmark.py:108`
- `examples/surface/szz_circuit_quality_report.py:30`

Confirmed on this branch:

```
ImportError: cannot import name 'normalize_traced_qis_tick_circuit' from 'pecos.qec.surface.circuit_builder'
```

Examples are not exercised in CI, so the green test suite does not catch this. Fix: update the three scripts to use `from pecos._traced_circuit import normalize_traced_tick_circuit` (matching how the tests were updated). A grep for the other removed names (`_extract_measurement_meas_ids`, `assert_traced_qis_tick_circuit_dem_ready`, the private replay helpers) found no other stragglers outside `src/` and `tests/`.

### Minor: `# noqa: F401` on compatibility re-exports

`pecos/qec/surface/decode.py` re-exports `capture_guppy_operation_trace` and `trace_guppy_into_tick_circuit` via `# noqa: F401`, and the module has no `__all__`. Project convention is to handle re-export F401 with `__all__` (or the `import x as x` idiom) rather than suppressions.

### Behavior notes (deliberate per the PR body, listed for the record)

- Duplicate runtime measurement-ID rejection and source-vs-runtime measurement-identity set equality are now enforced at the trace-replay boundary for all `trace_program_to_tick_circuit` callers, where previously only the DEM boundary checked identities. This is a contract tightening, not a pure refactor: a runtime emitting internal extra measurements (RUS-style) is now rejected at replay.
- Metadata key renamed `guppy_source_measurement_ids` -> `qis_source_measurement_ids`; `dem.py` reads new-then-old. A grep confirms no other reader of the old key exists, so the fallback covers everything.
- Idle conversion now catches seconds-to-nanoseconds overflow to infinity. A huge-but-finite duration still fails inside `TimeUnits()` with a less descriptive overflow error; not worth blocking on.

### Verification run locally (this branch, LLVM 21.1 environment)

- `tests/pecos/test_tracing.py` + `tests/pecos/test_hosted_operations.py`: 34 passed
- `tests/qec/test_guppy_dem_build.py` + `tests/qec/test_traced_qis_clifford_pipeline.py`: 51 passed
- `tests/qec/test_from_guppy_dem.py` (real Guppy capture end-to-end): 95 passed, 1 xfailed -- the local LLVM mismatch noted in the PR description does not reproduce in a managed LLVM 21.1 environment
- `tests/pecos/test_selene_sim_parity.py`: 15 passed
- Renamed-call tests in `test_surface_decoder.py` / `test_szz_interaction_basis.py`: 17 passed
- The tutorial quick-start example, run verbatim as a script: passes end-to-end including the JSON round trip
- `ruff check` on the new files: clean

### Test quality

The new guard tests mutate one field of a known-good trace and assert the specific error, so they pin the mechanism rather than the intent. Not every new guard has a direct negative test (negative `engine_trace_id`, boolean `chunk_index`, the duplicate-runtime-ID path); fine as-is, cheap to add if desired.

With the three example imports fixed, this is good to merge.

qciaran:
Addressed in 53edaf2b: updated all three example scripts to import and call the generic normalize_traced_tick_circuit helper; retained normalize_traced_qis_tick_circuit as a compatibility alias in the former surface module; and replaced the F401 suppressions with explicit compatibility assignments. Verified all three scripts start through --help, focused tracing/hosted-operation tests pass 34/34, and Ruff is clean on every changed file.

ciaranra:
## Re-review of 53edaf2b3

All findings from the previous review are addressed; verified by execution at this commit.

- The three example scripts now import `normalize_traced_tick_circuit` from `pecos._traced_circuit`; all three modules import cleanly (verified with the examples directory on `sys.path`).
- A compatibility alias `normalize_traced_qis_tick_circuit` was additionally added in `circuit_builder.py` and is identity-equal to the generic helper (pinned by the extended `test_surface_module_legacy_names_remain_compatible`).
- The `# noqa: F401` re-exports in `decode.py` were replaced with module-attribute assignments; no suppressions remain.

Verification at 53edaf2b3: `test_tracing.py` 19 passed; `test_from_guppy_dem.py` + `test_guppy_dem_build.py` 130 passed, 1 xfailed; `ruff check` clean on all files touched by the fix. Good to merge.

ciaranra (approved):
Approving per the review record: all prior findings addressed and verified by execution at 53edaf2b3; re-verified at head 86af47e50 after the dev MeasId merges (focused suites 217 passed; test_from_guppy_dem.py 95 passed, 1 xfailed; example scripts run; all CI checks green).

PR #401, opened 2026-07-31
Make the TickCircuit to DagCircuit conversion fallible

ciaranra:
Step 2 of #387's annotation-identity work (design: #400; step 1: #397). Both reviewers asked for this in three consecutive rounds of #391.

`impl From<&TickCircuit> for DagCircuit` becomes `TryFrom`, returning a `TickToDagError` that names the rejected tick and the reason. The owned `From<TickCircuit>` is removed too -- it delegated to the borrowed impl, so leaving it would have preserved exactly the panic being removed behind one `.unwrap()`.

## What this deletes

The Python binding's `to_dag_circuit` pre-scan: a second implementation of the measurement-id uniqueness invariant, kept "in step with `assign_measurement_ids`" by comment alone. It had already drifted once (disagreeing about `usize::MAX`), and it was structurally incomplete -- it saw only *supplied* ids, so a collision involving an id minted during conversion was invisible to it:

```python
tc = TickCircuit()
tc.tick().add_gate("MZ", [0])   # generic add: reserves no record, id-less
tc.tick().mz([1])               # mints record 0
tc.to_dag_circuit()             # was: PanicException (uncatchable)
```

That case now raises `ValueError: cannot convert TickCircuit to DagCircuit: gate at tick 1 was rejected: Measurement gate MZ reuses MeasId(0)...` -- verified by running. The invariant now lives in one place, and the conversion reports its own failures.

## Migration cost

Smaller than planned for. All three production callers (`DemSampler::from_tick_circuit`, `DetectorErrorModel::try_from_tick_circuit`, `..._with_noise_config`) already return `Result`, so the error maps into their existing enums and **no public signature changed**. The remaining call sites were test round-trips, now `try_from(...).expect(...)`.

One pre-existing test updated: it pinned the deleted pre-scan's error message ("share MeasId(7)"); the conversion's own message ("reuses MeasId(7)") is what survives. Same behaviour, same exception type.

## What this deliberately does not claim

The reverse direction, `From<&DagCircuit> for TickCircuit`, stays infallible-by-signature but is **not proven total**. Two panic sites are reachable from public API: `tick_circuit.rs:3349` unwraps on an invalid gate (reachable if `gate_mut` desyncs a gate's qubits), and `:3432` panics on an annotation referencing a non-measurement node (the public annotation API accepts arbitrary node numbers). Recorded on #387 rather than asserted away; fixing it is its own change, and this PR's scope is the direction with the live defect.

## Verification

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `cargo fmt --check` clean. `cargo test` with the exclusion set `pecos-cli` applies exit 0, 297 suites. Full Python lane 4371 passed / 68 skipped exit 0 (one more than before -- the new conversion test). `pre-commit run --all-files` exit 0.

New tests: the Rust conversion error carries the offending tick and reason (`converting_a_circuit_with_duplicate_ids_reports_the_tick`), and the Python case above that the pre-scan structurally could not catch (`test_an_idless_and_a_stamped_measurement_collide_loudly`).

### DIFF
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -271,7 +271,9 @@ impl<'a> DemBuilder<'a> {
     ///
     /// # Errors
     ///
-    /// Returns an error if detector or observable metadata is malformed.
+    /// Returns an error if detector or observable metadata is malformed, or
+    /// if the `TickCircuit` cannot be converted to a `DagCircuit` (two
+    /// measurements sharing a `MeasId`).
     ///
     /// # Panics
     ///
@@ -301,7 +303,9 @@ impl<'a> DemBuilder<'a> {
     ///
     /// # Errors
     ///
-    /// Returns an error if detector or observable metadata is malformed.
+    /// Returns an error if detector or observable metadata is malformed, or
+    /// if the `TickCircuit` cannot be converted to a `DagCircuit` (two
+    /// measurements sharing a `MeasId`).
     ///
     /// # Panics
     ///
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -288,7 +288,8 @@ impl<'a> DemBuilder<'a> {
         p_meas: f64,
         p_prep: f64,
     ) -> Result<DetectorErrorModel, DemBuilderError> {
-        let dag = pecos_quantum::DagCircuit::from(circuit);
+        let dag = pecos_quantum::DagCircuit::try_from(circuit)
+            .map_err(|err| DemBuilderError::ConfigurationError(err.to_string()))?;
         build_dem_from_circuit(&dag, NoiseConfig::new(p1, p2, p_meas, p_prep))
     }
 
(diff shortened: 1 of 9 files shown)


ciaranra:
One narrow adversarial review at the merged head `7ecb7bf3b` returned REVISE with two findings, both verified and fixed in `847d71097`.

## What it found

**The committed Rust doc-test crate still called the removed API** (`user_guide_circuit_representation.rs:226`, generated from `docs/user-guide/circuit-representation.md`), so it failed to compile -- caught by running. My local verification missed it: the fast Python lane deselects the doc-test wrapper, so matching CI's Rust exclusions was not the same as matching its coverage. Fixed through the framework's proper path: the markdown source updated (the snippet now shows `try_from(...).expect(...)` with a comment on why the conversion is fallible), the fixture regenerated with `scripts/docs/generate_doc_tests.py`, and the whole doc-test crate run -- exit 0, 21 suites. The stale `symbolic_replay.rs` comment naming `DagCircuit::from(&TickCircuit)` is corrected too.

**The three qec entry points' `# Errors` docs** (`try_from_tick_circuit`, `..._with_noise_config`, `DemSampler::from_tick_circuit`) did not mention the new conversion-failure case. All three now do.

## What it cleared, by running

No production defect in the conversion, the pre-scan deletion, the error-enum mappings, the interaction with the freshly merged #397, or the Python entry points. The Rust test's tick assertion fails under a wrong-tick mutation. The merge commit matches git's automatic three-way merge exactly.

One point worth keeping: the reviewer checked whether the new Python tests could distinguish the conversion's error from the deleted pre-scan's. The id-less test alone could not, but the duplicate-id test matches `reuses MeasId(7)` -- the pre-scan said `share MeasId(7)` -- so the pair together proves the old mechanism is gone, not merely bypassed.

## Verification at `847d71097`

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `cargo fmt --check` clean. `cargo test` with the `pecos-cli` exclusion set exit 0. The doc-test crate exit 0 (21 suites). `pre-commit run --all-files` exit 0.

PR #403, opened 2026-08-01
Pre-land the measurement-id resolution helpers

ciaranra:
Pre-lands the id-resolution helpers from `design/measurement-id-annotations.md` (step 3 of the revised migration order), so the eventual field pivot is a mechanical swap onto tested helpers rather than the judgment-laden commit where this subsystem's defects have historically bred.

This also settles the design's open error-taxonomy question against real code rather than on paper.

## The three helpers

**`DagCircuit::find_measurement(MeasId) -> Result<MeasRef, MeasResolveError>`.** The error enum distinguishes the three failure meanings the design required, which an `Option` cannot:

- `Unknown` -- never minted or supplied here (a caller bug, or a reference from another circuit);
- `Removed` -- reserved in `used_meas_ids` but no gate holds it. Removed ids stay reserved for the life of the circuit, so this cannot be confused with `Unknown`;
- `RecordLess { id, node }` -- a `MeasureLeaked` carrying a supplied id: a real measurement, but nothing that reads records can resolve through it.

Deliberately an O(gates) scan, not a maintained index -- an index would silently desync under `gate_mut` edits, and consumers resolving many ids build their own map from one pass.

**`DagFaultInfluenceMap::meas_index_of(MeasId) -> Option<usize>`** -- id to the map's own ordinal (id-rank order), documented as private to that boundary. `None` means absent; the map cannot know why, and the doc points callers at `find_measurement` when the reason matters.

**`ExpandedCircuit::meas_id_rank: BTreeMap<MeasId, usize>`** -- eeg's private ordinal (expansion order), built in the same walk that populates `measurement_qubit`. Empty for id-less legacy circuits. This is the map that will replace the silent `record_idx < num_meas` skip in `builder.rs` when consumers migrate.

Per the design's ordering rule: each boundary computes its own ordinal from an explicit map, ordinals are never interchanged, and nothing is `Vec`-indexed by a raw `MeasId`.

## Tests, all de-aliased

Every fixture uses scrambled, non-contiguous supplied ids so node index, id value, and ordinal cannot coincide -- in nearly every pre-existing test in this subsystem the three are the same number, which is why conflation bugs stayed invisible.

Five mutations run and killed: resolving by first qubit instead of the id's position, collapsing `Removed` into `Unknown`, dropping the `RecordLess` check, ranking eeg's map by id value, and indexing the influence map by id value. The last two are this bug class's signature shape.

## Observed, not fixed here

eeg's expansion matches `GateType::MZ` only; a `MeasureFree` passes through unchanged, silently dropping its measurement from the expansion. Pre-existing and independent of this change -- recorded on #387's ledger.

## Verification

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory. `cargo fmt --check` clean. `cargo test` with the `pecos-cli` exclusion set exit 0. The Rust doc-test crate exit 0. `pre-commit run --all-files` exit 0.

### DIFF
--- a/crates/pecos-core/src/meas_id.rs
+++ b/crates/pecos-core/src/meas_id.rs
@@ -21,7 +21,10 @@ use std::fmt;
 
 /// Unique identity of a measurement result.
 ///
-/// Lightweight (pointer-sized), `Copy`, directly usable as an array index.
+/// Lightweight (pointer-sized), `Copy`. An identity, **not an ordinal**:
+/// externally supplied ids may be sparse and out of order, so the numeric
+/// value must never be used to index an array. Boundaries that need a dense
+/// ordinal build their own `MeasId -> index` map.
 /// Analogous to [`QubitId`](crate::QubitId) but for measurement outcomes.
 ///
 /// # Example
@@ -29,13 +32,17 @@ use std::fmt;
 /// ```
 /// use pecos_core::MeasId;
 ///
+/// use std::collections::BTreeMap;
+///
 /// let m0 = MeasId(0);
 /// let m1 = MeasId(1);
 /// assert_ne!(m0, m1);
 ///
-/// // Direct array indexing
-/// let mut outcomes = vec![false; 10];
-/// outcomes[m0.0] = true;
+/// // Resolve through a map, never by the numeric value: externally supplied
+/// // ids (e.g. MeasId(9000)) are legal and would index out of bounds.
+/// let ordinal: BTreeMap<MeasId, usize> = [(m0, 0), (m1, 1)].into();
+/// let mut outcomes = vec![false; ordinal.len()];
+/// outcomes[ordinal[&m0]] = true;
 /// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct MeasId(pub usize);
--- a/crates/pecos-qec/src/fault_tolerance/propagator/dag.rs
+++ b/crates/pecos-qec/src/fault_tolerance/propagator/dag.rs
@@ -648,6 +648,13 @@ impl DagFaultInfluenceMap {
     /// once and build their own map.
     #[must_use]
     pub fn meas_index_of(&self, id: pecos_core::MeasId) -> Option<usize> {
+        // `extract_measurements` fills id-less entries of a mixed map with
+        // `MeasId(usize::MAX)` as a sentinel. No real gate can hold that id --
+        // insertion rejects it -- so a lookup for it must not "succeed" against
+        // the sentinel.
+        if id == pecos_core::MeasId(usize::MAX) {
+            return None;
+        }
         self.meas_ids.iter().position(|&held| held == id)
     }
 
@@ -2661,9 +2668,12 @@ mod tests {
         use pecos_core::MeasId;
         use pecos_quantum::Gate;
 
+        // Ids (7, 9, 4) on measurement nodes (3, 4, 5): no id equals its node,
+        // and no id equals its rank (sorted [4, 7, 9] -> ranks 0, 1, 2), so all
+        // three spaces are pairwise distinct for every entry.
         let mut dag = DagCircuit::new();
         dag.pz(&[0, 1, 2]);
-        for (qubit, id) in [(0usize, 9usize), (1, 1), (2, 5)] {
+        for (qubit, id) in [(0usize, 7usize), (1, 9), (2, 4)] {
             let mut gate = Gate::mz(&[qubit]);
             gate.meas_ids = smallvec::smallvec![MeasId(id)];
             dag.add_gate_auto_wire(gate);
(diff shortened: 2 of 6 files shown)


ciaranra:
One adversarial round on the helpers returned REVISE with six findings, all verified against the code and fixed in `64cd94705`.

## What the round caught

- **`MeasResolveError` was not exported.** Public in the return type, missing from `lib.rs`'s re-export list -- downstream code could not match the taxonomy this PR exists to provide. Caught with a failing import; invisible in-crate.
- **The taxonomy laundered `gate_mut` desync.** A forged id written into a live gate resolved *successfully*, and a duplicated id resolved to whichever gate scans first. The resolver now scans the whole circuit, detects duplicate holders, and checks the found id was reserved; all three states return a new `Inconsistent` variant whose documented reaction is "stop trusting the circuit". Each guard is mutation-killed.
- **My doc claimed the impossible**: that a reference from a different circuit lands in `Unknown`. Ids are circuit-local numbers; a colliding foreign id resolves to the local measurement, and no bare-id resolver can detect that. The doc now states the limit plainly -- rejecting foreign references needs validation where a reference *enters* a circuit, which is pivot work.
- **The mixed-stamping sentinel resolved as a held id.** `extract_measurements` fills id-less entries of a mixed map with `MeasId(usize::MAX)`; `meas_index_of` would find it. Guarded -- and my first test for the guard was vacuous (its fixture contained no sentinel), caught only when the mutation survived. The fixture now contains one.
- **The eeg test could not kill `.get(position)` -> `.first()`** -- two unary gates cannot distinguish per-position from first-id. Rebuilt around one batched MZ with scrambled ids. Two fixtures also had id/node coincidences (id 5 on node 5); all fixtures are now pairwise de-aliased across node, id, and ordinal.
- **`MeasId`'s own doc taught the bug** -- "directly usable as an array index", with a `.0`-indexing example. It now teaches map-resolution with a sparse-id warning.

## Design decisions this round settled (recorded in `design/measurement-id-annotations.md`)

**The forgery seam gets closed, as a pivot prerequisite.** Judged against the helpers rather than on paper: they make forgery *more* dangerous, because a forged id that collides with a real one now resolves to a real `MeasRef`. Private tuple field, constructor-controlled creation, opaque Python references -- its own change, since it touches every `MeasId(x)` constructor in the tree.

The reviewer's pivot-API consequences are also in the doc as a checklist: fallible `with_circuit_annotations` on sampler and `InfluenceBuilder` (plus dropping the latter's separate circuit argument), eeg's unsupported-measurement error, and making the sentinel representation explicit when the pivot touches those types.

## Verification at `64cd94705`

Cold clippy exit 0 on a fresh target directory; fmt clean; `cargo test` with the `pecos-cli` exclusion set exit 0; `pre-commit run --all-files` exit 0. Nine mutations killed across the branch, including the reviewer's surviving mutant and my own vacuous-test discovery.

ciaranra:
A narrow follow-up check of the hardening commit (`64cd94705` alone -- review-response code, the category with this effort's worst defect rate) returned REVISE. It was right to run: it found a HIGH in the rewritten resolver, verified by executing it.

## The laundering hole

`used_meas_ids` proves an id was reserved *once*, not that its current holder owns it. Remove a gate, then write its id onto another live gate through `gate_mut`, and `find_measurement` returned:

```
Ok(MeasRef { node: 3, qubit: QubitId(1), meas_id: MeasId(0) })
```

A stale annotation silently binding to a different measurement -- the exact failure class this subsystem's work exists to end, reintroduced by the code meant to end it.

Fixed in `a9d2fbe35` with tombstones: `remove_gate` records the removed gate's ids, and the resolver's precedence becomes fully honest --

- no holder, tombstoned -> `Removed` (which now *genuinely* means `remove_gate` ran);
- no holder, reserved, no tombstone -> `Inconsistent` (an id erased by `gate_mut` is not a removal);
- holder present but tombstoned or unreserved -> `Inconsistent` (reassigned or forged).

This also made the docs true instead of hedged: `Removed`'s old "unambiguous" claim, which the previous round forced me to weaken, is now accurate again because the tombstone provides the missing information.

## Also from the check

- A duplicate id *within one batched gate* was detected by the code but pinned by no test -- the reviewer showed a per-gate `.position()` regression would have passed the whole suite. Now pinned.
- The enum-level docs said "three cases" (there are four) and overstated in both directions; corrected.
- The reviewer confirmed clean: variant precedence for mixed leaked/MZ duplicates, forged-on-record-less ordering, the sentinel's unreachability as a legitimate id through every insertion route, the None-slot indexing, and the lib.rs export.

## Process note, on my own verification

My first mutation run reported one guard's mutation as surviving. It had never applied: `cargo fmt` had reformatted the target expression, my pattern missed, and the script carried no assert -- so it silently tested unmutated code and reported the result as meaningful. Re-run with an asserted anchor, the mutation is killed. Every mutation probe now asserts its anchor matched.

All three tombstone guards are mutation-killed with asserted anchors. Full verification at `a9d2fbe35`: cold clippy exit 0, fmt clean, CI-shaped Rust suite exit 0, pre-commit exit 0.

Issue #398, opened 2026-07-31
InfluenceBuilder simulates only the first qubit of a batched measurement gate


ciaranra:
**Impact: a batched measurement gate is simulated as if only its first qubit were measured, so faults on the other qubits are invisible to the influence map.** Found during design review of #387 step 3; independent of that change and present today.

`InfluenceBuilder::run_symbolic_simulation` (`influence_builder.rs:293`):

```rust
GateType::MZ | GateType::MeasureFree => {
    sim.mz(&[qubits[0]]);
    node_to_meas_idx[node] = Some(meas_idx);
    meas_idx += 1;
}
```

For `Gate::mz(&[0, 1])` -- a single node measuring two qubits -- only qubit 0 is measured symbolically, and the node maps to exactly **one** measurement index. Lookup then takes that first index (`:374`), while detector propagation separately includes *every* qubit on the node (`:553`).

So a batched node is under-measured in the simulation and over-included in propagation, and the two disagree.

## Why it is not fixed by the identity work

#387 step 3 replaces annotation references with `MeasId`, which lets an annotation name one measurement inside a batched node. That is a different problem. This is a separate `node -> one measurement index` model inside the symbolic simulation, and swapping the annotation's reference type does not touch it. The design originally claimed the batched-node fix came for free; for this path it does not.

## What a fix needs

A per-measurement mapping rather than per-node -- `MeasId -> {history index, node, qubit}` -- with `sim.mz` run for every measured qubit on the gate.

## Reachability

`DagCircuit::mz` creates one gate per qubit, so the batched shape does not arise from the ordinary builder. It arises from `add_gate(Gate::mz(&[0, 1]))` directly, and from `TickCircuit` conversion where batches are split per instance. Worth confirming which in-repo paths actually produce a multi-qubit measurement node before sizing the fix.

## Test that would catch it

A batched `Gate::mz(&[0, 1])` added via `add_gate`, with a fault on qubit 1 only, asserting the influence map records it. Today qubit 1 is never measured, so it cannot.

ciaranra:
[...]

Issue #395, opened 2026-07-31
symbolic_measurement_history cannot represent MeasureLeaked, so it refuses those circuits


ciaranra:
Split out of #391. `symbolic_measurement_history` now refuses circuits containing `MeasureLeaked` rather than returning a history whose dependencies dangle. That is correct but it is a refusal, not support.

## Why it refuses

A record-aligned history has one column per measurement *record*. `MeasureLeaked` consumes no record, so it has no column -- but it does collapse its qubit, and later outcomes can depend on its result.

`SymbolicMeasurementResult` carries `index` and an `outcome` bitset of other measurements' indices, both in the simulator's own numbering. Simply omitting the leaked measurement's result leaves those references pointing at columns that do not exist:

```
PZ(0,1); H(0); CX(0,1); MeasureLeaked(0); MZ(1)
```

produced a one-column history whose only entry had `index == 1` and a dependency on column 0, and `RawMeasurementPlan::sample` panicked.

Renumbering does not fix it. The leaked measurement can be the hidden random source the retained measurement depends on, and there is no way to express a source with no column.

## What support would need

A representation that distinguishes *record columns* from *random sources*, so a retained measurement can depend on a source that never becomes an output column. That is a change to `MeasurementHistory` and to `RawMeasurementPlan`'s sampling, not to the walk in `symbolic_measurement_history`.

Worth doing when leakage-aware sampling is actually wanted. Until then the refusal is the honest behaviour: the alternative shapes all produce wrong sample columns rather than an error.

## Related

- #394 -- whether a non-destructive measurement should clear a propagating Pauli at all. Independent of this, but touches the same gate.
- #387 -- the wider measurement-record consistency work.

ciaranra:
**Updated:** the behaviour described above is no longer what the code does. `symbolic_measurement_history` no longer refuses every circuit containing a `MeasureLeaked`.

It now renumbers the surviving measurements onto record columns and refuses **only** when a recorded outcome actually depends on a leaked result. So:

- `PZ(0,1); MeasureLeaked(0); MZ(1)` -- **works**. The leaked measurement holds no record, the `MZ` takes column 0, and nothing references the leaked outcome.
- `PZ(0,1); H(0); CX(0,1); MeasureLeaked(0); MZ(1)` -- **refused**. The `MZ` outcome XORs against the leaked measurement's random result, which has no column to name.

[...]

PR #408, opened 2026-08-02
Make annotation ingestion fallible across the three DEM consumers

ciaranra:
The fallible consumer APIs from `design/measurement-id-annotations.md`'s pivot checklist (recorded there by #403's adversarial round). With this, every annotation-ingestion path reports failure instead of silently thinning its outputs, and the pivot can swap node-based resolution for id-based resolution inside error channels that already exist.

## InfluenceBuilder

`with_circuit_annotations()` no longer takes a circuit: it ingests `self.dag`, the circuit the builder was constructed with. The parameter let annotations come from a *different* circuit than the one under analysis, silently producing nonsense; every in-repo caller passed the same circuit, verified before removal.

It returns `Result<Self, AnnotationIngestError>`: an observable referencing a missing node, an observable referencing a gate that consumes no measurement record, or a tracked Pauli with no `TrackedPauliMeta` gate. All three were silent skips. Ingestion is transactional and order-preserving -- a rejected ingest commits nothing, and interleaved observables and tracked Paulis keep their annotation order (the first draft of this change deferred one kind and committed the other, which would have reordered outputs consumers index positionally; caught before it compiled).

`build` stays infallible, deliberately deviating from the review checklist: post-pivot, resolution failures all surface at ingestion, so `build` has nothing honest to fail on, and an always-`Ok` `Result` is a fallback-shaped lie. The review round for this PR should attack that reasoning.

## DemSamplerBuilder

`with_circuit_annotations` returns `Result`, with a new `DetectorValidationError::UnresolvableAnnotationRef { output_kind, annotation_index, node }`. Four silent shapes closed: a detector node absent from the influence map, a detector measurement outside every auto-detector, an observable node absent from the map, and the `i32`-overflow branch that returned silently *empty* records.

## eeg

`EegBuildError`, and `build`/`build_dem_string`/`summary` all return `Result`:

- `UnsupportedMeasurement` -- a `MeasureFree` used to pass through the MZ-only expansion *unchanged*, silently deleting its measurement record from the deferred circuit. This closes the ledger item filed on #387 when the expansion map was built.
- `UnresolvableAnnotationRecord` -- the `record_idx < num_meas` silent skip is dead; the error names both sides of the mismatch.

`EegDemBuilder` has no callers outside its own crate, so the signature changes ripple nowhere.

## Behavioral evidence

The entire `pecos-qec` and `pecos-eeg` suites pass under the strict errors -- no legitimate in-repo circuit trips any of them, so every new rejection is fail-loud gain rather than a behavior tax. Python binding (`InfluenceBuilder.build`) raises `ValueError` with the resolver's message.

Six new tests, each mutation-killed with asserted anchors: the three `AnnotationIngestError` variants, the transactionality guarantee, the sampler's observable drop, and both eeg errors.

## Verification

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target directory (it caught two stray `#[must_use]` attributes and a missing `# Panics` section the suite passed over). `cargo fmt --check` clean. `cargo test` with the `pecos-cli` exclusion set exit 0. Rust doc-test crate exit 0. Full Python lane 4390 passed / 68 skipped exit 0, run with `CARGO_TARGET_DIR` unset. `pre-commit run --all-files` exit 0.

### DIFF
--- a/crates/benchmarks/benches/modules/gpu_influence_sampler.rs
+++ b/crates/benchmarks/benches/modules/gpu_influence_sampler.rs
@@ -107,7 +107,7 @@ fn build_influence_maps(
 ) -> (DagFaultInfluenceMap, GpuInfluenceMapData) {
     let tracked_pauli_qubits: Vec<usize> = (0..num_data).collect();
     let builder = InfluenceBuilder::new(circuit).with_z(&tracked_pauli_qubits);
-    let influence_map = builder.build();
+    let influence_map = builder.build().expect("circuit is replayable");
 
     let (
         num_loc,
--- a/crates/pecos-gpu-sims/examples/cpu_vs_gpu_comparison.rs
+++ b/crates/pecos-gpu-sims/examples/cpu_vs_gpu_comparison.rs
@@ -93,7 +93,7 @@ fn main() {
         // Build influence map
         let tracked_pauli_qubits: Vec<usize> = (0..num_data).collect();
         let builder = InfluenceBuilder::new(&circuit).with_z(&tracked_pauli_qubits);
-        let influence_map = builder.build();
+        let influence_map = builder.build().expect("circuit is replayable");
 
         let num_locations = influence_map.locations.len();
 
@@ -181,7 +181,7 @@ fn main() {
 
         let tracked_pauli_qubits: Vec<usize> = (0..num_data).collect();
         let builder = InfluenceBuilder::new(&circuit).with_z(&tracked_pauli_qubits);
-        let influence_map = builder.build();
+        let influence_map = builder.build().expect("circuit is replayable");
 
         let (
             num_loc,
--- a/crates/pecos-gpu-sims/examples/full_pipeline_example.rs
+++ b/crates/pecos-gpu-sims/examples/full_pipeline_example.rs
@@ -86,7 +86,7 @@ fn main() {
     // Build influence map with a tracked Z Pauli (sensitive to X errors)
     let builder = InfluenceBuilder::new(&circuit).with_z(&[0, 1, 2]);
 
-    let influence_map = builder.build();
+    let influence_map = builder.build().expect("circuit is replayable");
     println!("   Locations: {}", influence_map.locations.len());
     println!("   Detectors: {}", influence_map.detectors.len());
     println!("   Measurements: {}", influence_map.measurements.len());
@@ -161,7 +161,7 @@ fn main() {
     // Build influence map with a tracked X Pauli (sensitive to Z errors on this plaquette)
     let builder = InfluenceBuilder::new(&circuit).with_x(&[0, 1, 2, 3]);
 
-    let influence_map = builder.build();
+    let influence_map = builder.build().expect("circuit is replayable");
     println!("   Locations: {}", influence_map.locations.len());
     println!("   Detectors: {}", influence_map.detectors.len());
 
@@ -224,7 +224,7 @@ fn main() {
     for num_rounds in [1, 2, 4, 8] {
         let circuit = build_repetition_code_circuit(num_rounds);
         let builder = InfluenceBuilder::new(&circuit).with_z(&[0, 1, 2]);
-        let influence_map = builder.build();
+        let influence_map = builder.build().expect("circuit is replayable");
 
         let (
             num_locations,
--- a/crates/pecos-gpu-sims/examples/pipeline_benchmark.rs
+++ b/crates/pecos-gpu-sims/examples/pipeline_benchmark.rs
@@ -158,7 +158,7 @@ fn benchmark_circuit(
     // Build influence map (common to both pipelines)
     let build_start = Instant::now();
     let builder = InfluenceBuilder::new(circuit).with_z(tracked_pauli_qubits);
-    let influence_map = builder.build();
+    let influence_map = builder.build().expect("circuit is replayable");
     let build_time = build_start.elapsed();
 
     let num_locations = influence_map.locations.len();
--- a/crates/pecos-gpu-sims/examples/profile_samplers.rs
+++ b/crates/pecos-gpu-sims/examples/profile_samplers.rs
@@ -514,7 +514,7 @@ fn main() {
         // Build influence map
         let tracked_pauli_qubits: Vec<usize> = (0..num_data).collect();
         let builder = InfluenceBuilder::new(&circuit).with_z(&tracked_pauli_qubits);
-        let influence_map = builder.build();
+        let influence_map = builder.build().expect("circuit is replayable");
         let num_locations = influence_map.locations.len();
 
         // Export for GPU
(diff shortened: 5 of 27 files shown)


ciaranra:
The adversarial round returned REVISE with four findings, all verified against the code by running before acting. Fixes in `b5ba9757c`. My flagged deviation lost, and the round's findings reshaped this PR into the design it should have had from the start.

## Finding 1: `build` infallible was wrong -- reversed, and the hole was deeper

The reviewer's witnesses: an ignored `RZ` manufactured a phantom detector (forward replay skipped it while backward propagation simplified it), and a batched `mz` undercounted detectors. Fixing it exposed worse: the replay also silently skipped **mid-circuit resets**, so a reused qubit was analyzed as still carrying its pre-reset correlations.

`build` now returns `Result<_, InfluenceBuildError>`. The replay applies `PZ`/`QAlloc`, collapses `MeasureLeaked` without a record, refuses batched measurement nodes (representation belongs to the pivot's per-measurement model, per #398 and the design -- refusing now means no wrong answer ever ships, and the model lands once in its designed place), and refuses any gate it cannot represent. Three regression tests pin the witnesses, including a mid-circuit-reset case asserting the post-reset measurement is deterministic on its own.

## Finding 2: the sampler's detector mapping was wrong twice over -- replaced

The first-match-by-qubit/basis auto-detector hop rejected legitimate detectors on qubit-reuse circuits (reviewer witness), and stored auto-detector indices into a field documented -- and consumed by `sample_dual` -- as absolute raw-measurement indices. Both pre-existing; my strictness converted the first from silent to hard failure. The hop is deleted: detector annotations resolve directly to raw measurement indices. A simplification, not a patch.

## Finding 3: the eeg boundary was routable-around -- consolidated

Rather than bolting guards onto each bypass (the duplicated-invariant mistake this project has already paid for), the check moved to where the limitation lives: **`expand_circuit` itself is fallible**, and `ExpandedCircuit::aux_qubit_for_record` is the single record resolver. The builder's own pre-check is deleted as redundant; `dem_simulator`, the heisenberg estimator entrances, and all 13 expansion call sites in the exp bindings route through the same two functions. The exp bindings' silent-drop loops now raise `ValueError` naming the annotation and offset. `run_dem_simulation` returns a real `DemSimulationError` instead of the `.ok()?` chain that erased diagnostics -- closing the note on #395.

Cold clippy then caught five more consumers (`pecos-gpu-sims` examples, a benchmark module) that the CI-shaped test build never compiles. Patched; the exp extension was rebuilt and the Python lane re-run against the new binding rather than carrying forward results from the stale one.

## Finding 4: the transactionality test was vacuous -- removed

Partial state cannot escape a failed ingest because the method consumes `self`; the property is structural and untestable. The test claiming to test it is gone, and the comment now states the guarantee for what it is.

## Verification at `b5ba9757c`

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 (it caught the gpu-sims consumers and an `.err().expect()` pattern across three test sites). `cargo fmt --check` clean. `cargo test` with the `pecos-cli` exclusion set exit 0. Full Python lane 4390 passed / 68 skipped exit 0, both extensions freshly built, `CARGO_TARGET_DIR` unset. `pre-commit run --all-files` exit 0. New guards mutation-killed with asserted anchors: the expansion's measurement check, the record resolver's fallback-to-zero, and the three replay refusals.

ciaranra:
Round 2 returned REVISE with five findings. Four were verified against the code and fixed as stated; one was verified against the code but **falsified by the test suite** — its premise was wrong, and the fix it asked for broke 37 real workflows. Details below. Fixes in `01b289636` and `08e3d4d9e`.

## Finding 1: `MeasureLeaked` corrupted replay record alignment — confirmed, refused

The reviewer's witness reproduced: the replay called `sim.mz()` for a leaked measurement without allocating an external record, and the simulator's internal history counter advanced anyway, desynchronizing every subsequent record index and manufacturing a phantom detector. This is the same defect class as the `symbolic_measurement_history` bug #391 fixed.

The replay now refuses `MeasureLeaked` (`InfluenceBuildError::UnsupportedGate`). Representing leaked measurements correctly belongs to the pivot's per-measurement model (#395, #398); refusing means no wrong answer ships in the meantime. Regression test `a_leaked_measurement_is_refused_not_desynced` pins it, mutation-killed.

## Finding 2: `MeasureFree` normalization "bypasses the entrance" — premise falsified; fixed by lowering, not rejection

The finding's code claim was accurate: the bindings map `MeasureFree` to `GateType::MZ` before `expand_circuit`, so the entrance's `MeasureFree` rejection could never fire on Python-originated circuits. But the recommended fix — reject `MeasureFree` before normalization — was applied and run against the full Python suite: **21 failures + 16 errors**, all real surface-code workflows (`test_analysis_meas_sampling`, `test_meas_sampling_backend`, `test_raw_measurement_result`, docs fault-catalog). `mz_free` is the standard ancilla readout.

The premise was wrong: `MeasureFree` is record-bearing (`consumes_measurement_record` includes it), and the free has no stabilizer effect — so `MeasureFree -> MZ` is a correct lowering, not an evasion. The defect this PR targets is *dropping* measurement records; lowering preserves them exactly. The original rejection (added in round 1 of this PR) was the actual regression.

Fix: `expand_circuit` — the checked entrance — now performs the lowering itself (`GateType::MZ | GateType::MeasureFree` in the walk; only `MeasureLeaked` refused). The binding-level rejections are reverted; both bindings pass `MeasureFree` through faithfully. Tests: `a_measure_free_circuit_expands_with_its_record_intact` (asserts both records survive; mutation-killed by dropping `MeasureFree` from the walk arm) and `a_measure_leaked_circuit_is_refused`.

## Finding 3: Heisenberg helpers silently discarded invalid detector records — confirmed, routed through the resolver

Both estimator entrances kept `if m < len` guards that skipped out-of-range records and analyzed an empty detector as `Ok`. Both now pre-resolve `detector_aux` through `ExpandedCircuit::aux_qubit_for_record` before the 2^n loop and return `EegBuildError` on any unresolvable record.

## Finding 4: stochastic DEM simulation bypassed checked resolution — confirmed, made fallible

`build_tick_circuit` is now fallible; the detector and observable builders route through a shared `resolve_record_offset_ref` helper instead of `filter_map`-dropping invalid offsets, the coherent-synthesis arithmetic is checked, and the `record_idx: usize::MAX` diagnostic is replaced with an error carrying the original signed offset. The reviewer's `rec[-2]` witness now errors instead of returning `Ok`.

## Finding 5: reset witness was count-only — confirmed, strengthened

Both correct and mutated replay produce one detector, so the count assertion was vacuous. The test now asserts the detector contains exactly the post-reset measurement (`measurements.len() == 1`, `measurements[0].qubit == 0`), which kills the no-op-arm mutation.

## Verification at `08e3d4d9e`

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` exit 0 on a fresh target dir. `cargo fmt --check` clean. CI-shaped `cargo test` (pecos-cli exclusion set) exit 0. Both extensions rebuilt via maturin; full Python lane **4390 passed / 68 skipped, exit 0** with `CARGO_TARGET_DIR` unset — the 37 failures introduced by the round's finding-2 fix are gone. `pre-commit run --all-files` exit 0. Mutation checks with asserted anchors: `MeasureLeaked` refusal, `MeasureFree` lowering, reset-witness arm, dual-output index mapping (both "use annotation index" and off-by-one variants).

PR #409, opened 2026-08-02
Pivot annotations to measurement identity (measurement_ids: Vec<MeasId>)

ciaranra:
Step 4 of #387, per `design/measurement-id-annotations.md`: detector and observable annotations now reference measurements by identity instead of by position.

## The change

- `AnnotationKind::{Detector, Observable}` store `measurement_ids: Vec<MeasId>`. The old `measurement_nodes: Vec<usize>` meant record indices in a `TickCircuit` and node ids in a `DagCircuit` -- one type, two meanings, translated (buggily) on every conversion.
- `DagCircuit::detector`/`observable` (and the tick equivalents) take `&[MeasRef]` / `&[TickMeasRef]` and return `Result`. Each reference is validated O(1) against the gate it names -- stale and forged references are rejected at construction (`AnnotationRefError`, `TickAnnotationRefError`). The annotation Pauli is Z on each referenced measurement's own qubit, fixing the batched-gate Z-spray corruption.
- Both circuit conversions copy annotations **verbatim**. The translation maps (`meas_record_to_node`, `dag_node_to_record_indices`, `node_to_meas_idx`) and their `filter_map` silent drops are deleted; `TickMeasRef.record_idx` -- a documented lie for external ids -- is gone.
- Consumers resolve ids through the step-3 helpers (#403): `find_measurement` in the influence builder (one Z term per measurement on its own qubit), `meas_index_of` in the DEM sampler/builder, `aux_qubit_for_id` (via `meas_id_rank`) in eeg. Every unresolvable reference is a typed error naming the id and the annotation.
- Python boundary: plain-int measurement refs are rejected (`TypeError`); `(node, qubit)` / `(tick, gate_idx, qubit)` tuples are resolved against the circuit, never trusted; `annotations()` exposes `measurement_ids`. `DagCircuit::meas_refs(node)` recovers per-measurement refs from a batched `add_gate` node -- a detector on one measurement of a batched gate is now expressible.

## A defect the de-aliased tests exposed

`InfluenceBuilder`-built influence maps carried **two internal orderings**: detectors, propagation data, and the sampler's raw channels in symbolic-replay order, but `measurements`/`meas_ids` re-sorted by id rank -- so `meas_index_of` did not index the raw stream. Invisible while ids were positional (the orders coincide), wrong the moment they were not. `InfluenceBuilder` now emits one order per map (replay order throughout); `DagFaultAnalyzer` maps were already internally consistent. Recorded in the design doc's "As built" section along with two other deviations (O(1) construction validation instead of a per-ref scan; tuple handles instead of opaque Python objects, since the tuples never contain a forgeable id).

## Tests

`crates/pecos-qec/tests/meas_id_pivot_tests.rs`: every tier with non-positional, non-contiguous ids (9, 1, 5) -- influence-relation and per-measurement-influence equivalence between positional and scrambled ids (locations matched by physical key, indices through each map's own ordinal), a sampler dual-output witness with self-verifying non-vacuity asserts, batched-gate single-measurement detector, removed/unknown/record-less reference errors, tick<->dag round-trip of scrambled ids, and a 2^44 sparse id proving no id-sized allocation. Plus eeg byte-identical DEM under scrambled ids, tick/dag forged-ref rejection, and the Python `mz_with_ids` tier. Eight mutations run against the guards (skip validation, restore the id-rank re-sort, resolve by raw id value, spray the Pauli, drop the verbatim copy, skip unresolved ids, index eeg by id value, drop annotations in conversion) -- all killed.

## Verification

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` on a fresh target dir, `cargo fmt --check`, the CI-shaped `cargo test` exclusion set, both extensions rebuilt via maturin, the full Python lane with `CARGO_TARGET_DIR` unset, `pre-commit run --all-files`, and the #391 end-to-end DEM ordering harness (50k shots) at this commit -- all green.

### DIFF
--- a/crates/benchmarks/benches/modules/fault_catalog.rs
+++ b/crates/benchmarks/benches/modules/fault_catalog.rs
@@ -383,13 +383,13 @@ fn build_rotated_z_memory_circuit(distance: usize, rounds: usize) -> Result<Memo
     let final_data_measurements = circuit.tick().mz(&data_qubits);
     let num_measurements = circuit.num_measurements();
 
-    let mut detectors: Vec<Vec<i32>> = Vec::new();
+    let mut detectors: Vec<Vec<usize>> = Vec::new();
 
     for &meas_ref in z_round_measurements[0]
         .iter()
         .take(code.num_z_stabilizers())
     {
-        detectors.push(relative_records(num_measurements, &[meas_ref]));
+        detectors.push(ref_meas_ids(&[meas_ref]));
     }
 
     for round in 1..rounds {
@@ -398,14 +398,14 @@ fn build_rotated_z_memory_circuit(distance: usize, rounds: usize) -> Result<Memo
             .zip(x_round_measurements[round - 1].iter())
             .take(code.num_x_stabilizers())
         {
-            detectors.push(relative_records(num_measurements, &[current, previous]));
+            detectors.push(ref_meas_ids(&[current, previous]));
         }
         for (&current, &previous) in z_round_measurements[round]
             .iter()
             .zip(z_round_measurements[round - 1].iter())
             .take(code.num_z_stabilizers())
         {
-            detectors.push(relative_records(num_measurements, &[current, previous]));
+            detectors.push(ref_meas_ids(&[current, previous]));
         }
     }
 
@@ -418,7 +418,7 @@ fn build_rotated_z_memory_circuit(distance: usize, rounds: usize) -> Result<Memo
                 .into_iter()
                 .map(|q| final_data_measurements[q]),
         );
-        detectors.push(relative_records(num_measurements, &refs));
+        detectors.push(ref_meas_ids(&refs));
     }
 
     let logical_z_refs: Vec<TickMeasRef> = code
@@ -427,14 +427,17 @@ fn build_rotated_z_memory_circuit(distance: usize, rounds: usize) -> Result<Memo
         .iter()
         .map(|&q| final_data_measurements[q])
         .collect();
-    let observables = vec![relative_records(num_measurements, &logical_z_refs)];
+    let observables = vec![ref_meas_ids(&logical_z_refs)];
 
     circuit.set_meta(
         "num_measurements",
         Attribute::String(num_measurements.to_string()),
     );
-    circuit.set_meta("detectors", Attribute::String(records_json(&detectors)));
-    circuit.set_meta("observables", Attribute::String(records_json(&observables)));
+    circuit.set_meta("detectors", Attribute::String(meas_ids_json(&detectors)));
+    circuit.set_meta(
+        "observables",
+        Attribute::String(meas_ids_json(&observables)),
+    );
 
     Ok(MemoryCircuit {
         circuit,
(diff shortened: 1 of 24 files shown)


ciaranra:
Two independent adversarial reviews (blind to each other) both returned REVISE. Every finding was verified against the code before acting -- one review's file citations were entirely hallucinated paths, but its mechanisms checked out against the real files; the other supplied an executed witness. Fixes in `a04b69db7` and the follow-up commit.

## Confirmed and fixed

**Duplicate MeasIds silently mis-bound annotations at two consumers** (both reviews; one demonstrated it by execution: an eeg observable on id 5 silently produced the *other* qubit's error probability when a second measurement also carried id 5). `TickCircuit` legally holds duplicate supplied ids, `try_add_gate` does not advance the mint counter (so supplied/minted collisions are constructible through the public API), and `gate_mut` can duplicate ids on a `DagCircuit`. The influence-builder path was already safe (`find_measurement` returns `Inconsistent`); eeg's expansion (`meas_id_rank` was last-wins) and the sampler's annotation ingestion (`meas_index_of` was first-match) were not. Both now refuse duplicate ids loudly -- `EegBuildError::DuplicateMeasId` and `InvalidMetadata` -- mirroring the guard the DEM JSON path already had. Four regressions pin it, including the supplied/minted collision; all guard mutations killed.

**`measurement_order` + stamped ids mixed two ordinal spaces** (one review; the mechanism verified: `resolve_meas_id_to_tc_index` returns an influence-map ordinal that `build_measurement_mappings` re-fed through `tc_to_influence` as if it were a tick ordinal). The recommended fix -- reject the combination -- was applied and then **falsified by the suite**: 148 failures, because the surface pipeline routinely supplies `measurement_order` on minted-id circuits, where per-qubit chronology holds on both sides and the combination is benign. The fix that survives targets the actual defect: the stamped branch now uses the resolved index directly (the double-composition is gone -- pinned by a regression whose map order disagrees with tick order), and the rejection is narrowed to the genuinely unmappable case, a supplied order with *non-positional* stamped ids. Both directions mutation-tested (always-compose, always-accept, always-reject -- each killed by its own test).

**sim_neo JSON annotation ingestion silently thinned** (both reviews). `create_annotations_from_json` swallowed malformed JSON, `filter_map`-dropped out-of-range records, and skipped weakened annotations. It is now fallible end-to-end: malformed JSON and unresolvable references raise `ValueError` naming the annotation and offset. Making it loud immediately exposed a real casualty of the old behavior: the traced-QIS pipeline emits `meas_ids`-only detector JSON, which the local parser (records-only) could never read -- those circuits had been running with their annotations silently discarded. The parser now accepts both `records` and `meas_ids` entries, resolved against the circuit's refs, so the traced-QIS annotations are ingested for the first time. Two Python regressions added, plus the traced-QIS suite passing with annotations actually applied.

**The record-arithmetic deviation was a latent trap** (both reviews; one asked for it to be recorded, the other for the real migration). The real migration happened: the three files now emit `meas_ids` JSON and read shot bits in id space -- the `record_idx - num_measurements` arithmetic is gone, not renamed. The design's "As built" section records the closure.

**Fixture weakness** (both reviews): `ids[1] = 1` coincided between the positional and scrambled fixtures, so observable retargeting was never exercised. Scrambled sets are now `[9, 4, 7]` -- no id coincides with any position or with the positional twin. Also added: annotation survival across `split_batched_tick_commands`, and the missing-coverage list (duplicate-across-calls, minted/supplied collision, order+ids rejection).

**Doc corrections**: the influence-map ordering docs claimed id-rank universally (false for `InfluenceBuilder` maps since the one-order fix; now documented as one private per-map ordinal, builder-dependent); the "O(1) validation" claim is now stated honestly as linear in the named gate's batch width; the dual-output witness docstring no longer overclaims -- one review correctly showed both its sides resolve through the same `meas_index_of`, so it pins sampler internal consistency while the ordering invariant is killed by the influence-comparison test.

## Rejected, with reasons

**Construction-time global uniqueness validation** (one review's proposed fix for the duplicate-id finding). The other review independently verified that consumers catch `gate_mut` desync via `find_measurement`'s `Unknown/Removed/Inconsistent`, and a global check would reintroduce the O(circuit) scan per reference that the O(1)-validation deviation deliberately avoids. The duplicate-id holes were consumer-boundary defects and are fixed at those boundaries, which is also where the design says reference validation belongs.

## Verification

At the pushed commit: cold `cargo clippy --locked --workspace --all-targets -- -D warnings` on a fresh target dir, `cargo fmt --check`, the CI-shaped `cargo test` exclusion set, both extensions rebuilt, the full Python lane with `CARGO_TARGET_DIR` unset, `pre-commit run --all-files`, and the #391 50k-shot DEM ordering harness -- all green. All four new guards mutation-tested (guard disabled, its regression fails, guard restored); the eight pivot mutations from the previous round remain killed.

Issue #396, opened 2026-07-31
DemBuilder and the fault sampler use different depolarizing-channel conventions


ciaranra:
**Impact: DEM entry probabilities and sampler-side group probabilities are computed under different conventions, so comparing them directly does not match.** Surfaced during end-to-end verification of #391; predates that PR and is unaffected by it.

Two places model the same depolarizing channel differently:

- `DemBuilder` treats an n-way channel as n *independent* channels, each with `q = 1 - (1-p)^(1/n)`. This is documented on `per_channel_probability` in `dem_builder/builder.rs`.
- `build_fault_table` / `RawMeasurementPlan` treat it as *mutually exclusive* 1-of-k alternatives.

## Size of the discrepancy

Individual DEM entry probabilities differ from the corresponding sampler-side group sums at O(p^2/n). Measured at a deliberately large `p2 = 0.12`: 0.03309 versus 0.032 for a per-4-combo group.

The *implied detector marginals* still agree to roughly 2e-6 even at that noise level, because the conventions converge as the higher-order terms cancel across the group. At realistic QEC noise the difference is far below sampling error.

## Why it is worth recording anyway

Nothing is wrong today, and neither side is incorrect on its own terms. But anyone who compares a DEM entry probability against a probability derived from the fault sampler -- validating one against the other, or debugging a mismatch -- will find a discrepancy that looks like a bug and is not. That cost is paid by whoever hits it next, repeatedly, unless it is written down.

A one-line note on each side pointing at the other's convention would be enough. Reconciling the two is not obviously worth doing given the marginals agree.

ciaranra:
Fixed by #420, which removed per_channel_probability and the n-independent-channels convention entirely. Both sides now share the 1-of-k convention: DemBuilder/DemSampler split channels as p1/3 and p2/15 (dem_builder/dem_sampler.rs:2549,2624; dem_builder/mem_builder.rs:217,303; dem_builder/builder.rs:433) and the fault sampler uses conditional probabilities 1/3 and 1/15 (fault_sampler.rs:1091,1133). The issue's concrete discrepancy (DEM entry 0.03309 vs per-4-combo group 0.032 at p2=0.12) can no longer arise -- both conventions now give 0.032.

PR #410, opened 2026-08-03
Pauli propagation collapses instead of resetting at non-destructive measurements

### LINKED ISSUE #394: Pauli propagation clears at non-destructive measurements, dropping errors that should persist
**Impact: a Pauli error is discarded at a non-destructive measurement, so it stops flipping later measurements it should still flip.** Found during review of #391; the behaviour predates that PR and is unchanged by it.

`propagate_tracked_pauli_forward` clears the propagating Pauli after a measurement (`pauli_frame.rs`), and `fault_sampler.rs` does the same. For `MeasureFree` that is right -- the qubit is gone. For a non-destructive `MZ` it is not.

## Why it is wrong

Collapse projects onto an eigenstate; it does not reset to the ideal state. Take an X error before a Z measurement:

- ideal: `|0>` -> measure -> outcome 0, state `|0>`
- faulty: `|0>` -> X -> `|1>` -> measure -> outcome 1, state `|1>`

After the measurement the two states still differ by X, so a second measurement on the same qubit is flipped too. Reproduced directly:

```python
from pecos_rslib import StateVec
s = StateVec(1)
s.run_gate("X", {0})
print(s.run_gate("MZ", {0}))   # {0: 1}
print(s.run_gate("MZ", {0}))   # {0: 1}   <- still flipped
```

Clearing predicts the second measurement is unflipped. This is also how Stim behaves: `M` measures without resetting and the error persists; `MR` resets and clears it.

`MeasureLeaked` is executed as a plain `MZ` with no reset (`quantum.rs:205`), so the same argument applies to it.

## Why it has stayed latent

Surface-code ancillas are re-prepared with `PZ` each round, and `PZ`/`QAlloc` clear legitimately. Data qubits are measured once at the end. So a repeated measurement on one qubit without an intervening reset is rare in the circuits currently exercised, and the two behaviours agree everywhere else.

## The propagation paths do not agree with each other

- `pauli_frame.rs` and `fault_sampler.rs` clear.
- `propagator/pauli.rs` treats `MeasureLeaked` as identity -- no collapse.
- `influence_builder.rs` skips it entirely.
- The physical simulators retain the post-measurement eigenstate, per the reproduction above.

So this is not one wrong site; it is four sites with three different opinions.

## What a fix needs

Clearing should follow whether the gate actually resets or discards the qubit, not whether it is a measurement. That changes fault propagation for any circuit with a repeated measurement, so it wants validating against Stim on circuits that actually contain one -- not just the surface-code cases where the behaviours coincide.

Deliberately not fixed in #391: that PR makes measurement ids mandatory and preserves this behaviour exactly rather than changing it. Changing measurement physics inside it would have made the DEM comparisons there impossible to interpret.

ciaranra:
Fixes #394: Pauli propagation cleared the propagating error at every measurement, predicting that an X before a non-destructive `MZ` stops flipping afterward. Collapse projects; it does not reset -- verified directly on `StateVec` (`X; MZ -> 1; MZ -> 1`): `MZ` measures without preparing, and only the MP* family (or an explicit `PZ`) resets the qubit.

## The rule, in one place

`cross_measurement` in `propagator/pauli.rs` is now the single implementation, generic over both Pauli-propagation representations (`PauliProp`, `BitmaskPauliProp`):

- **`MZ` / `MeasureLeaked`** (non-destructive Z-collapse), forward: the X component survives -- the faulty and ideal runs still differ by X after collapse, so later measurements keep flipping -- and the Z component is absorbed. Backward: the symplectic mirror -- a Z-type observable passes (errors before the measurement flip both it and later measurements), an X-type observable is dropped, because it gains no deterministic dependence on anything before the collapse.
- **`MeasureFree`** (discard): nothing crosses, in either direction.
- `PZ`/`QAlloc` clearing is unchanged -- resets legitimately clear.

All four sites the issue names now agree: `pauli_frame.rs` and `fault_sampler.rs` route their forward walks through the shared rule; the backward walkers (`propagator/dag.rs`, `propagator/tick.rs`) inherit it through `apply_gate`, which previously treated measurements as identity.

## The backward fix kills a phantom the issue did not name

Backward identity-crossing let an X-type observable pass through an earlier collapse, manufacturing influences like "a Z fault before the first measurement flips a later rotated measurement" -- a correlation no simulator reproduces. The new backward drop removes it (pinned by `a_z_before_a_collapse_does_not_haunt_a_later_rotated_measurement`).

The same phantom was propping up two fault-enumeration tests: their tracked `logical_X` operators were placed *after* the transversal Z-readout of every data qubit, where an X-type operator is gauge. They are moved before the readout, where the operator is physically meaningful -- the tracked-Pauli API is positional by design.

## Why this stayed latent

Surface-code circuits re-prepare every ancilla and measure data qubits once, so collapse and reset coincide everywhere the existing suites looked. The one test asserting the old behavior (`test_propagate_x_absorbed_by_mz`) was the defect's own regression test; it is rewritten to the verified physics, with new absorbed-Z and `MeasureFree` discriminators beside it.

## Validation

- Cross-validation against Stim (the reference the issue names, whose `M` measures without resetting) on a circuit that actually repeats a measurement (`TestRepeatedMeasurementCollapse`): mechanism sets and probabilities identical -- the joint `D0 D1` mechanism appears and the old `D0`-only split does not.
- Influence-map discriminators: X-before-repeated-MZ flips both measurements, X-between flips one, Z flips nothing, `MeasureFree` stops propagation where `MZ` does not.
- Unit tests on `cross_measurement` for all four direction/type combinations.
- Four mutations (forward clear-all, backward identity, backward over-clear, `MeasureFree`-as-`MZ`), each killed by a named test.
- Cold clippy pedantic `-D warnings` on a fresh target dir, CI-shaped `cargo test`, full Python lane with both extensions rebuilt, the #391 DEM ordering harness, and pre-commit -- all green at the pushed commit.

### DIFF
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -519,7 +519,9 @@ fn propagate_forward(
             GateType::PZ | GateType::QAlloc if !loc.qubits.is_empty() => {
                 prop.clear_qubit(loc.qubits[0]);
             }
-            // MZ: X component flips the measurement, then qubit state collapses
+            // The X component flips the measurement and then survives the
+            // collapse -- a non-destructive measurement absorbs only the Z
+            // component. Only a discarded qubit (`MeasureFree`) clears fully.
             GateType::MZ | GateType::MeasureFree | GateType::MeasureLeaked
                 if !loc.qubits.is_empty() =>
             {
@@ -529,7 +531,12 @@ fn propagate_forward(
                 {
                     affected.insert(meas_idx);
                 }
-                prop.clear_qubit(q);
+                super::propagator::cross_measurement(
+                    prop,
+                    q,
+                    loc.gate_type,
+                    super::propagator::Direction::Forward,
+                );
             }
             _ => {}
         }
@@ -2327,10 +2334,13 @@ mod tests {
         );
     }
 
+    /// Collapse projects; it does not reset. After the first measurement the
+    /// faulty and ideal states still differ by X, so the second measurement is
+    /// flipped too -- verified directly on `StateVec` (X; MZ -> 1; MZ -> 1)
+    /// and matching Stim's `M`-versus-`MR` distinction. Clearing here was the
+    /// defect: it predicted the second measurement unflipped.
     #[test]
-    fn test_propagate_x_absorbed_by_mz() {
-        // Circuit: MZ(0) MZ(0) — X on q0 should flip first MZ only
-        // (MZ collapses qubit, absorbing the error)
+    fn test_propagate_x_persists_through_nondestructive_mz() {
         let mut tc = TickCircuit::new();
         tc.tick().mz(&[QubitId(0)]);
         tc.tick().mz(&[QubitId(0)]);
@@ -2339,8 +2349,53 @@ mod tests {
         let affected = propagate_single(PauliType::X, 0, 0, &gates, &meas_pos);
         assert_eq!(
             affected,
+            BTreeSet::from([0, 1]),
+            "X before a non-destructive MZ flips it AND every later measurement \
+             on the qubit until a reset or free"
+        );
+    }
+
+    /// The Z component is what collapse absorbs: it commutes with the
+    /// measurement, and both branches land on the same eigenstate up to phase.
+    #[test]
+    fn test_propagate_z_component_absorbed_by_mz() {
+        // Z -> (via H) X flips the first MZ; the surviving component after the
+        // collapse is X only, so an H after the measurement rotates it to Z,
+        // which cannot flip the final MZ. If the Z component survived the
+        // collapse, the recombined Y would flip it.
+        let mut tc = TickCircuit::new();
+        tc.tick().h(&[QubitId(0)]);
+        tc.tick().mz(&[QubitId(0)]);
+        tc.tick().h(&[QubitId(0)]);
+        tc.tick().mz(&[QubitId(0)]);
+
+        let (gates, meas_pos) = flatten_tick_circuit(&tc);
+        let affected = propagate_single(PauliType::Y, 0, 0, &gates, &meas_pos);
+        assert_eq!(
+            affected,
             BTreeSet::from([0]),
-            "X should flip first MZ only, not second"
+            "the Y's X part flips MZ 0 (Y -> H -> Y anticommutes with Z); only \
+             X survives the collapse, and H turns it into Z, invisible to MZ 1"
+        );
+    }
+
+    /// `MeasureFree` discards the qubit, so it genuinely clears: an error
+    /// before it cannot reach anything after.
+    #[test]
+    fn test_propagate_x_cleared_by_measure_free() {
+        let mut tc = TickCircuit::new();
+        tc.tick()
+            .try_add_gate(pecos_core::Gate::mz_free(&[0usize]))
+            .expect("gate is valid");
+        tc.tick().pz(&[QubitId(0)]);
+        tc.tick().mz(&[QubitId(0)]);
+
+        let (gates, meas_pos) = flatten_tick_circuit(&tc);
+        let affected = propagate_single(PauliType::X, 0, 0, &gates, &meas_pos);
+        assert_eq!(
+            affected,
+            BTreeSet::from([0]),
+            "X flips the MeasureFree it precedes and nothing after the discard"
         );
     }
 
--- a/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/fault_sampler.rs
@@ -207,7 +207,9 @@ pub(crate) enum PauliType {
 /// - All single-qubit Cliffords: Clifford conjugation via direct Pauli-basis updates
 /// - All two-qubit Cliffords: Clifford conjugation via direct Pauli-basis updates
 /// - `PZ`, `QAlloc`: absorbs all Pauli components on the reset qubit
-/// - `MZ`: records `X`-component flip, then absorbs all components (state collapse)
+/// - `MZ`, `MeasureLeaked`: records `X`-component flip, then absorbs the Z
+///   component only -- collapse projects, it does not reset, so the X keeps
+///   propagating. `MeasureFree` absorbs everything: the qubit is discarded.
 ///
 /// **No-op** (pass through without noise or transformation):
 /// - `I`, `Idle`, `QFree`, `MeasCrosstalkGlobalPayload`,
--- a/crates/pecos-qec/src/fault_tolerance/influence_builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/influence_builder.rs
@@ -691,33 +691,34 @@ impl<'a> InfluenceBuilder<'a> {
         };
 
         for (det_idx, detector) in detectors.iter().enumerate() {
-            // Build combined Pauli from all measurements in the detector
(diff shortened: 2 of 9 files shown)


ciaranra:
Two independent adversarial reviews (blind arms), both REVISE. Both confirmed the core physics independently -- one from first principles (forward `(x,z)->(x,0)`, backward symplectic dual `(x,z)->(0,z)`, `MeasureFree` zero map, phase irrelevant), one by identifying the rule as the standard Pauli-frame algebra with the measurement gauge fixed to identity, with forward/backward agreement holding by construction. Both adjudicated the silent-drop-versus-error question as built: dropping is correct at the influence layer; loud rejection of non-deterministic detectors belongs at detector validation. The arms were complementary: each found a real defect the other missed.

## Confirmed and fixed

**A detector over repeated same-qubit measurements lost its between-measurement mechanisms** (HIGH; found by one arm, reproduced here by execution: the detector existed, and *no fault location anywhere* influenced it, where the reference DEM has `error(0.125) D0`). `InfluenceBuilder::propagate_detectors` XORed every member's observable at the circuit end before one backward walk, so two same-qubit Z seeds cancelled before propagation started. It now runs one walk per member, seeded at the member's own node; the toggling recorder cancels shared-ancestry influences exactly as the detector's XOR readout does. Pinned by `a_detector_over_repeated_measurements_keeps_between_faults` (between-fault present, prep-side faults cancel); mutation (revert to end-seeding) killed.

**The PR itself recreated the #394 bug class inside `pauli_prop_checker`** (found by the other arm). The checker reads syndromes from the end-state, which was only correct while measurements were identity: after this PR, `MeasureFree` cleared the flip its own measurement recorded, silently zeroing syndromes -- while `propagate_until_tick` kept a private identity dispatch, giving one checker two collapse semantics. Fixed with a checker-local **flip ledger** (`apply_gate_flip_ledger`): measurements keep the X component readable -- including `MeasureFree`, whose flip must survive to the end-read; the next reset clears it exactly when the wire stops carrying it -- and still absorb Z, closing the phantom-rotation hole the identity dispatch had. All four checker walkers now share it. Along the way `propagate_until_tick` turned out to have two more private opinions: a Pauli encoding with **Y and Z swapped** relative to the canonical one (latent wrong syndrome histories for Y/Z faults) and inverted fault-tick skip logic; both unified. Regressions pin the `mz_free` end-read and the ledger crossing; mutations (plain `apply_gate` in the ledger; the swapped encoding) killed.

**Docs corrected**: `apply_gate`'s header no longer claims measurements are identity; the fault-sampler table states the collapse rule; `cross_measurement`'s justification is rewritten in the gauge-fixed frame-algebra terms one review supplied (the earlier independent-run phrasing overclaimed -- correct observation); the "gauge" wording on the repositioned tracked Paulis now says "destroyed as an observable"; `cross_measurement` is `pub(crate)`.

**Tests added per the round**: the SX-rotated repeated measurement (integration coverage for the keep-Z-on-Y branch), and the cross-validation comparison now runs with `p_meas > 0`. That variant surfaced a convention worth knowing: both DEM generation paths model `p_meas` as a physical X before the measurement, so on repeated measurements it joins the propagating joint mechanism rather than flipping one record classically -- the two paths agree mechanism-for-mechanism, and the convention choice is recorded for #396.

## Attempted and reverted, with the reason on record

The same combine-at-end shape exists in the measurement-crosstalk exact replay, and one arm listed it as the same timing error. The per-measurement rewrite that fixes the influence builder **breaks this mode**: a global-crosstalk payload and its victim measurement can live on disconnected wires with no DAG edge ordering them, so topological positions cannot express their circuit-time relationship (verified: the exact-deterministic global test loses its only mechanism). The other arm had rated this site "constrained by the mode's validation; informational". A correct fix needs a tick-aware time model -- out of this PR's scope, reverted, and filed.

## Scoped to #411 (filed, with both arms' evidence)

Crosstalk replay time model (above); `gadget_checker`'s fully independent dispatch (identity preps *and* measurements -- predates this PR, untouched by it); `CliffordGateable::mz` on the frame simulators (identity, no absorb -- a semantics decision, since those types serve both as propagation frames and as flip ledgers); the missing determinism guard on the metadata-detector DEM path (both arms, non-blocking) -- where a user detector on a random measurement yields gauge-dependent rows instead of a loud rejection.

## Verification

At the pushed commit: cold `cargo clippy --locked --workspace --all-targets -- -D warnings` on a fresh target dir, `cargo fmt --check`, the CI-shaped `cargo test` set, both extensions rebuilt, the full Python lane with `CARGO_TARGET_DIR` unset, the cross-validation file (36 tests), the #391 DEM ordering harness, and pre-commit -- all green. Round mutations: end-seeded detector walk, plain-`apply_gate` ledger, swapped `until_tick` encoding -- each killed by its named regression; the four original collapse mutations remain killed.

ciaranra:
Narrow round on the post-review delta (`cd43abd87..b6626879d`), two blind arms. Verdicts split -- one SHIP, one REVISE -- and the split was adjudicated on evidence class: the REVISE arm's finding was read-based, the SHIP arm ran the disputed semantics.

## The split, resolved

The REVISE arm rated the flip ledger a HIGH: one `PauliProp` serves both as measurement record and as physical residual, and physical-state consumers (`check_error_propagation`'s weight gate, `decoder_integration`'s residual checks) would read a discarded wire's retained flip as a physical error. The mechanism is real. The severity is not, on three grounds the other arm established by execution and search: the circuit runner executes `MeasureFree` as a measurement with no free and no reset, so within the checker's own executable model the wire genuinely keeps its flip -- end-read retention *matches* the consumer's semantics; no reachable checker circuit performs free-then-reuse at all; and the retention is byte-identical to the checker's behavior before this PR. The architectural point stands and is recorded in #411 as the long-term shape (a result type carrying a collapse-accurate residual plus per-measurement recorded flips) rather than a blocker.

## Fixed from the round

Both arms converged on one real defect: the ledger's "until a reset clears it" doc was aspirational -- the shared dispatcher keeps preps transparent, so a stale entry survived re-preparation. The ledger now clears at `PZ`/`QAlloc` (matching the fault-sampler walker), a regression pins it, and the guard-disabled mutation fails it.

## Everything else the round confirmed

- The encoding swap in `propagate_until_tick` was a genuine latent bug, not a compensated convention: every producer of `fault.paulis` uses the canonical 1=X, 2=Y, 3=Z, so Y and Z fault syndrome histories were silently exchanged. The new fault-tick rule is correct and no consumer depended on the old pre-crossing read.
- The per-member detector seeding was A/B'd by full influence-map dump against the reviewed base: byte-identical on every ordinary case (different-qubit members, shared ancestry, multi-member), and on repeated-measurement circuits the base had *two* failure modes -- the missing between-measurement mechanisms this round's fix claims, plus phantom attributions of after-measurement faults that the fix also removes.
- Both new checker unit tests fail when run against base semantics -- non-vacuous by execution, not inspection.
- The topological bound on the seeded walks is sound (positions are unique; equality correctly includes the member's own node).
- One noted item for #396's conventions ledger, already recorded: both DEM paths model `p_meas` as a physical X before the measurement, which the new cross-validation test now bakes in as the (internally consistent) convention.

Also filed on #411 from this round: the residual/record separation above, and `propagate_faults`' pre-existing disregard of per-fault placement in multi-fault configurations.

## Verification

At `23448ec5d`: cold clippy `-D warnings`, `cargo fmt --check`, the CI-shaped `cargo test` set, pre-commit -- all green (the delta touches only Rust internals; the full Python lane was green at the reviewed head and runs again in CI). Mutation for the new guard killed; all prior round mutations remain killed.

Issue #412, opened 2026-08-03
First-class measure-and-prepare (MP*) across the gate sets


ciaranra:
**Measure-and-prepare exists only at the simulator trait: `CliffordGateable::mpz` (measure +Z, then prepare |0>) has a default implementation and native state-vec overrides, but no `GateType`, so no circuit can express it and no consumer can reach it.**

Layer inventory today:

- `CliffordGateable::mpz` -- present (default: `mz` + conditional X; overridden in `state_vec_soa`, `state_vec_sparse_aos`). Zero callers outside pecos-simulators.
- `GateType` -- no variant. Circuits express measure-and-reset as the two-gate `mz; pz` idiom.
- Propagation (post #410) -- the idiom composes correctly: collapse at the `MZ` (X survives, Z absorbed), clear at the `PZ`.
- Noise models -- cannot distinguish fused measure-reset noise from independent `mz` + `pz` noise; surface-code circuits pay two noise events where hardware has one operation.
- Translators -- external fused measure-and-prepare ops must be split into two gates on import; export re-fuses nothing.

What first-class MP* would touch, if adopted:

1. `GateType::MPZ` (PECOS naming is the MP* family, measure-then-PREPARE, matching `CliffordGateable::mpz`), `consumes_measurement_record() == true`, `Gate::mpz` constructor, validate rules.
2. Tick/DAG builders (`mpz()` handles returning refs like `mz()`), id minting identical to `mz`.
3. Propagation: one arm in `cross_measurement`/`apply_named_gate` -- forward: record flip then clear all (the measurement's flip is consumed by its record; nothing propagates past the reset); backward: kill, like `PZ`. The checker flip-ledger variant keeps the X readable like its `MeasureFree` handling.
4. Executors/engines: lower to `CliffordGateable::mpz` (finally reachable).
5. Noise models: a distinct measure-prepare noise channel (or explicit reuse of `p_meas`+`p_prep` with documentation).
6. Translators: external fused ops map to `MPZ` one-to-one where a counterpart exists; QASM/PHIR lowering decisions.
7. The rest of the MP* family, decided together or explicitly deferred: `MPX`/`MPY` (measure-and-prepare in other bases), and whether `mz_free` keeps its separate discard identity (it should -- discard is not preparation).

None of this blocks anything current: the `mz; pz` idiom is correct. The value is expressiveness (matching what hardware performs as one operation), noise fidelity, and translator round-trips.

ciaranra:
[...]

Issue #411, opened 2026-08-03
Measurement-crossing holdouts after the collapse-semantics fix


ciaranra:
**Impact: four sites keep their own measurement-crossing or measurement-noise opinions after #410 unified the propagation walkers; each is pre-existing, none is changed by #410, and each can produce wrong or gauge-dependent analyses on the repeated-measurement circuits #410 legalizes.** Split out of PR #410's adversarial review round (two independent arms; findings verified by execution where noted).

## 1. Measurement-crosstalk exact replay assumes a linearization that disconnected wires do not have

`measurement_parity_anticommutes_after_crosstalk_payload` (`dem_builder/builder.rs`) seeds every detector-member observable at the circuit end and walks backward through one arbitrary topological linearization to the payload position. Two problems, one attempted-and-reverted fix:

- Same-qubit measurement pairs cancel their seeds before the walk (the shape #410 fixed in `InfluenceBuilder::propagate_detectors`).
- A per-measurement rewrite seeded at each measurement's own node -- correct for the influence builder -- **breaks this mode**: a global-crosstalk payload and its victim measurement can live on disconnected wires with no DAG edge ordering them, so `topo_position` comparisons are meaningless (verified: the exact-deterministic global test loses its only mechanism). The payload/measurement relationship here is circuit *time*, which the DAG does not represent for disconnected qubits.

A correct fix needs a time model (tick-aware ordering) for the replay, not a topo-order patch. The mode's own validation (`hidden_mz_result_before_crosstalk_payload`) constrains the damage today.

## 2. `gadget_checker` has a fully independent dispatch

`propagate_through_circuit` and the syndrome-history walk treat state prep **and** measurements as identity ("State prep and measurements don't affect Pauli propagation") and count output-error weight on qubits a `MeasureFree` has discarded. It never used the shared `apply_gate`, so #410 did not change its behavior -- it is simply a fourth opinion. Route it through the shared dispatch (or the checker flip-ledger from #410, which is the end-read-compatible variant) and add crossing regressions.

## 3. `CliffordGateable::mz` on the frame simulators neither collapses nor absorbs

[...]

PR #416, opened 2026-08-03
First-class measure-and-prepare: GateType::MPZ core (#412 phase 1)

ciaranra:
Phase 1 of #412: first-class measure-and-prepare in the core gate set. `GateType::MPZ` (measure +Z, then prepare |0>) exists end to end -- gate type, builders, propagation, execution -- so circuits no longer spell the fused hardware operation as two gates, and `CliffordGateable::mpz` is reachable from the circuit layer for the first time.

## What MPZ is, at every layer

- **Gate set**: `GateType::MPZ = 107`, `consumes_measurement_record() == true` (it records like `MZ`), `Gate::mpz`, validate rules, display/diagram entries.
- **Builders**: `DagCircuit::mpz`/`try_mpz` and `TickCircuit` tick handle `mpz()`, returning measurement refs exactly like `mz()` -- ids mint through the shared predicate, so annotations, conversions, and resolution work unchanged.
- **Propagation** (`cross_measurement`): nothing crosses in either direction -- the record flip is taken by the walker before the crossing, then the built-in preparation absorbs the error forward and blocks observables backward. Discriminators pin it: an X before an MPZ flips *its* readout and nothing after the reset; the checker flip-ledger keeps the recorded flip readable (its built-in preparation deliberately does not clear the ledger, mirroring `MeasureFree`).
- **Execution**: all three engine gate loops (state-vector, Clifford-message, general-message) execute `sim.mpz`; ByteMessage gains the `mpz` builder; the symbolic replay, hidden-measurement synthesis, and fault-catalog history lower it as measure-then-prepare; eeg's expansion lowers it to its measurement-plus-projection form. The `pecos-neo` runtime and HUGR/QASM translators refuse it loudly (their vocabularies are later #412 phases).
- **Semantics anchor**: `mpz` and the `mz; pz` idiom build byte-identical eeg DEMs at `p_prep = 0`, and a three-engine execution test pins `X; MPZ; MZ -> [1, 0]`.

## Noise: consistency over fidelity, deliberately

Every noise path models MPZ as the measurement-half fault only (a fault before the readout). The engines runtime, both DEM builders, and eeg therefore agree with each other; the prepare-half fault arrives for all of them at once with #412's dedicated measure-prepare channel, which also needs the after-location the location model does not yet give measurements. Until then, the split spelling's explicit reset draws preparation noise the fused gate does not -- the equivalence test documents this and tests at `p_prep = 0`.

## A simulator bug this surfaced and fixes

Making `mpz` reachable exposed that **`StabVec` had the #394 defect at the simulator level**: its diagonal-frame measurement shortcut flipped the *reported* outcome while collapsing the stored, unflipped eigenstate and discarding the frame -- so `X; MZ; MZ` read `[1, 0]` where the state-vector ground truth reads `[1, 1]`, and the default measure-and-prepare correction composed against the wrong state. Latent for the usual reason: re-prepared ancillas never re-measure. The projection now re-aligns the state with the report (collapse projects; it does not reset), pinned by regressions for both the bare repeated measurement and the `mpz` correction.

## Scope and follow-ups (recorded on #412)

Vocabulary surfaces deliberately out of phase 1: Python bindings (`run_gate` dispatch, tick/dag binding methods), SLR/PHIR/QIS/Guppy/QASM translators, the `pecos-neo` gate set, the dedicated measure-prepare noise channel, and the `MPX`/`MPY` family plus the `MeasureFree -> MZFree` rename candidate. Everything unreachable refuses loudly rather than guessing.

## Verification

Cold `cargo clippy --locked --workspace --all-targets -- -D warnings` on a fresh target dir, `cargo fmt --check`, the CI-shaped `cargo test` set, both extensions rebuilt, the full Python lane with `CARGO_TARGET_DIR` unset, the #391 DEM ordering harness, and pre-commit -- all green at the pushed commit. Mutations killed by named tests: MPZ crossing behaves like `MZ` (keeps X), eeg passes MPZ through unlowered, and all three executor arms run it as a plain `mz` (no reset) -- each caught, plus the StabVec collapse regressions.


ciaranra:
Two independent adversarial reviews (blind arms), both REVISE, with heavily convergent findings -- and one arm's 400-seed differential fuzz *validated the StabVec collapse fix decisively* (0/400 failures on this branch versus 197/400 on dev, including the exact `X; MZ; MZ` witness). Every finding was verified against the code; all fixes are in and mutation-tested.

## Confirmed and fixed

**The checker flip-ledger retained MPZ entries on a live wire** (both arms, HIGH). MPZ hands back a usable |0> wire, so a retained ledger X spread through later gates as a phantom -- and the fused gate disagreed with its own `mz; pz` lowering, whose explicit `PZ` clears. MPZ now clears the ledger like the reset it contains; a lowering-equivalence witness (`fused == split`, plus nothing-spreads-through-the-CX) pins it. Its mid-circuit flips are invisible to end-reads exactly as the lowering's always were; readable mid-circuit flips belong to the residual/record separation already filed on #411.

**The dispatch sweep's blind spot was real** (both arms; the review prompt asked for exactly this hunt). Sites keyed on explicit `MZ`-lists rather than `MeasureFree`-membership were missed and are now fixed, each with the review's named line: `TickFaultAnalyzer::extract_measurements` (silently dropped MPZ records -- run-verified witness, now pinned by a test), the three DEM noise lanes (`dem_sampler`, `mem_builder`, and the sampler's `p_meas` bucket -- MPZ records were noiseless or got 1q-depolarizing rates), `SpacetimeLocation::is_measurement`, the `unitary_rep` diagram family, `GeneralNoiseModel` (tracked MPZ as a 1q-unitary then died on an outcome-count mismatch; now tracked as a measurement with post-engine noise and prepared-qubit bookkeeping), the fourth public `CircuitExecutor`, and `GateType::FromStr` (which also gained the pre-existing missing `MeasureFree`/`MeasureLeaked` arms, closing the Python silent-custom-gate path for the whole family).

**Fixing the sampler lane surfaced one more real defect**: the `InfluenceBuilder` detector walk backward-crossed the seed's *own* measurement node, so an MPZ's reset killed its own seed before the before-location recorded -- its records had no fault mechanisms at all. The walk now treats the seed's own node as the readout it represents (record the before-location, no crossing); behavior-identical for `MZ` (the suite proves it) and correct for MPZ, pinned by the p_meas mechanism test.

**QASM export mangled MPZ** into literal `unknown q[0];` -- it now lowers to its `measure`-then-`reset` spelling (with the classical target when mapped).

**`StabVec::mnz` was broken pre-existing** (one arm read it, the other ran it: the Z-frame compose commutes with a Z readout and can never flip the outcome). Fixed via the trait's reference decomposition; a differential test against `SparseStab` pins `mnz`/`pnz` including the state-after check. The fuzz numbers showing widespread `mnz` disagreement were pre-existing on dev and orthogonal to this PR's eager-X fix, which the same fuzz validated.

**Docs**: the `mpz`/`mz_free` rustdoc splice is repaired; the stale fault-sampler comment updated.

## Deliberately unchanged, with reasons

`AbsorbBasisGates`' `is_z_prep`/`is_z_measure` exclude MPZ -- a conservative miss (skipped optimization, never wrong output); widening them safely needs its own analysis and rides #412. The Python bindings still cannot *construct* MPZ (no `mpz` method) -- that is the planned bindings phase of #412; what this round closed is the silent-custom-gate hole when naming it. The prepare-half noise remains deferred to #412's dedicated channel, unchanged from the PR's documented decision.

## Verification

At the pushed commit: cold `cargo clippy --locked --workspace --all-targets -- -D warnings` on a fresh target dir, `cargo fmt --check`, the CI-shaped `cargo test` set, both extensions rebuilt, the full Python lane with `CARGO_TARGET_DIR` unset, the #391 DEM ordering harness, and pre-commit -- all green. Round mutations, each killed by its named test: tick extraction dropping MPZ, the sampler `p_meas` bucket reverting, the own-node readout rule removed, the ledger retaining MPZ entries, and `mnz` reverting to the Z-frame shortcut -- plus the original PR's mutations, which remain killed.

PR #420, opened 2026-08-03
Align idle-noise vocabulary across DEM, engines and neo, and emit the requested Pauli channel

### LINKED ISSUE #424: Deprecation ramp: align pecos-engines GeneralNoiseModel idle-noise API with the structured vocabulary
## Goal

Align the idle-noise surface of `GeneralNoiseModel` / `GeneralNoiseModelBuilder` in
`crates/pecos-engines` (and its PyO3 fluent builder in
`python/pecos-rslib/src/engine_builders.rs`) with the structured idle-noise vocabulary
(PR #420 layer-1 families; #423 layer-2 channels) — via a staged deprecation ramp,
because this API has external users.

## Ramp

1. **Additive release**: new structured setters (family-style or `with_idle_channels(...)`)
   alongside the existing methods; nothing existing changes behavior.
2. **Warning release**: legacy idle methods emit deprecation warnings. Each warning must
   state the *numerically equivalent* new spelling, not just the new name — see units
   below.
3. **Removal / major release**: legacy methods removed or explicitly re-documented as
   low-level primitives, per what usage telemetry and downstream feedback allow.

## Unit trap (the reason the ramp needs care)

The builder rescales its public inputs before they reach the runtime fields
(`crates/pecos-engines/src/noise/general/builder.rs`, ~line 1006): square-root scaling on
the quadratic rate, a `p_idle_coherent_to_incoherent_factor * 0.5` adjustment on the
stochastic branch, and a cycles-to-radians conversion. The structured vocabulary matches
the *runtime application semantics* (`general.rs`), not these builder input units. A naive
"use `p_idle_sin_squared` instead of `p_idle_quadratic_rate`" migration would silently
change the physics; the conversion formulas must be documented and embedded in the
warnings.

## Alignment notes

- The linear family maps cleanly: engines already uses rate + `SingleQubitWeightedSampler`
  over X/Y/Z/L — the same alphabet the structured models adopted (including first-class
  `L`).
- The `p_idle_coherent: bool` interpretation switch is replaced by law-in-the-name
  families / channel types.
- `p_idle_coherent_to_incoherent_factor` (default 1.5) is an engines-only calibration
  knob; it should get an explicit, documented identity rather than being folded silently
  into any migration.
- Related: #422 (native-surface NoiseModel port), #423 (layer-2 vocabulary + EEG
  alignment). Same standing principle: one idle-noise vocabulary at every level of PECOS,
  capability and units enforced loudly, never silently approximated.

### LINKED ISSUE #428: Reshape the Guppy DEM construction API into the house builder pattern
## Problem

`build_dem_from_guppy(...)` is a verb-named function that takes every option as
a keyword argument and returns a `GuppyDemBuild` noun you then unpack:

```python
dem_build = build_dem_from_guppy(
    rep_code_memory,
    num_qubits=7,
    detectors=detectors,
    observables=observables,
    idle_after_2q_duration=1.0,
    p_idle_linear=0.01,
    p_idle_linear_model={"X": 0.25, "Y": 0.25, "Z": 0.5},
    p_idle_sin_squared=0.03,
    p_idle_sin_squared_model={"Z": 1.0},
    p1=0.002,
    p2=0.02,
    p_meas=0.02,
    p_prep=0.02,
)
dem = dem_build.dem
```

This reads as an inverted builder: the configuration is a flat keyword pile
(now ~30 parameters wide after the structured idle families), the call site
cannot be composed incrementally, and the result object is the thing you would
normally get from `.build()`. It also diverges from the house builder-of-builders
style used by `sim(...)`/`sim_neo(...)`, where configuration is chained and a
terminal call produces the artifact.

## Proposal (sketch, to be settled in review)

```python
dem_build = (
    dem_from_guppy(rep_code_memory)
    .qubits(7)
    .detectors(detectors)
    .observables(observables)
    .idle_after_2q(1.0)
    .noise(gate_noise(p1=0.002, p2=0.02, p_meas=0.02, p_prep=0.02))
    .idle_noise(linear(0.01, {"X": 0.25, "Y": 0.25, "Z": 0.5}), sin_squared(0.03, {"Z": 1.0}))
    .build()
)
dem = dem_build.dem
```

Points to settle:

- Whether the noise configuration groups into its own builder(s) shared with the
  other consumers (ties into #422, #424, #425 and the layer-2 channel vocabulary
  in #423 — a channel list is a natural fit for a chained `.idle_noise(...)`).
- Whether `DetectorErrorModel.from_guppy` (the JSON-metadata entry point) gains
  the same shape or stays the one-call convenience form.
- Migration: both entry points are public and documented; a deprecation ramp is
  needed rather than a rename, as in #424.

## Related

Same consistency theme as #422-#425. Non-blocking for PR #420, which uses the
current shape; the docs there name the result `dem_build` rather than `build` so
the verb/noun collision at least does not appear in the examples.

---

## Unify the two Guppy DEM entry points (added after review of the current surfaces)

There are two entry points that do the same job:

| | `DetectorErrorModel.from_guppy` | `build_dem_from_guppy` |
|---|---|---|
| detectors/observables | JSON strings | typed `Detector`/`Observable` (bare tag strings, `rec[-k]`, `result_ref`) |
| returns | the DEM | `GuppyDemBuild`: `.dem`, `.audit`, `.schema_fingerprint`, `.evaluate_result_columns()` |
| pipeline | trace, normalize, idle passes, attach metadata, Rust builder | identical |

Measured overlap: **38 and 37 parameters, 35 shared**. The only differences are
`detectors_json`/`observables_json`/`num_measurements` versus
`detectors`/`observables`. They produce byte-identical DEMs for the same
specification (verified while writing `docs/workflows/guppy-dem-decoding.md`).

The duplication is not free: every parameter added must be added twice, in two
signatures, two docstrings, and two guard wirings. PR #420 did exactly that four
times (idle passes, the three idle-noise families, the symmetric-default and
`L`-key round, and the `p_idle` removal).

### Direction

One construction pipeline, two convenience shapes at the surface:

- The builder from this issue is the single pipeline and returns the rich
  object (`.dem`, `.audit`, `.evaluate_result_columns()`).
- `DetectorErrorModel.from_guppy` stays as the one-call convenience and keeps
  returning the DEM (changing its return type would be breaking), but delegates
  to the builder and accepts **either** spelling for detectors/observables:
  typed specs or the JSON strings that other tooling emits (e.g.
  `meta_tc.get_meta("detectors")` from the surface builder).
- `build_dem_from_guppy` becomes a deprecated alias for the builder path,
  retired on the ramp described above.

Net effect: parameters live in exactly one place, the typed and JSON spellings
stop being a fork, and the audit trail plus result-column evaluator become
reachable from the documented entry point rather than only from the second one.

### LINKED ISSUE #429: Give the DEM-consuming Python decoders a uniform call and result protocol
## Problem

The Python decoder objects that all consume a DEM expose three different call
and result shapes, so user code cannot iterate over decoders uniformly:

| Decoder | Methods | Per-shot result |
|---|---|---|
| `PyMatchingDecoder` | `decode`, `decode_batch` | `.correction` (per-observable vector) |
| `TesseractDecoder` | `decode`, `decode_batch`, `decode_syndrome` | `.observables_mask` (bitmask) |
| `DemAwareDecoder` | `decode_syndrome` | `.observables_mask` (bitmask) |

Writing an explicit comparison therefore needs a different expression per
decoder:

```python
pymatching_errors += pymatching.decode(syndrome).correction[0] != actual
tesseract_errors += (tesseract.decode_syndrome(syndrome).observables_mask & 1) != actual
bp_osd_errors += (bp_osd.decode_syndrome(syndrome).observables_mask & 1) != actual
```

(from `docs/workflows/guppy-dem-decoding.md`, which shows this honestly rather
than hiding it behind the stringly-typed `SampleBatch.decode_count(text, name)`
shortcut).

## Proposal

Give the DEM-consuming decoders one shared Python protocol: a single per-shot
method name and a single result shape (mask and per-observable vector both
available on the result, rather than one or the other depending on the class),
plus a batch variant with the same contract. Keep the existing methods as
deprecated aliases during a ramp, as in #424.

The Rust side already has this notion: `ObservableDecoder::decode_obs` is the
wide primitive the registry dispatches through
(`python/pecos-rslib/src/fault_tolerance_bindings.rs`,
`create_observable_decoder`). The Python surface should expose the same
uniformity instead of the stringly-typed registry being the only uniform path.

## Related

Consistency series: #422, #423, #424, #425, #428. The stringly-typed
`decode_count(text, "name")` registry is fine as a fast-path convenience, but it
should not be the only way to treat decoders uniformly.

### LINKED ISSUE #430: DemAwareDecoder silently wraps observable bits past 64 (no width guard)
## Bug

`DemAwareDecoder` accepts a DEM with more than 64 observables and then packs
predictions with a `u64` bitmask, so observable indices >= 64 wrap onto low bits.

- `PyDemAwareDecoder::from_dem` builds a `DemCheckMatrix` and never calls
  `ensure_observables_fit_u64` — that guard appears **zero** times in
  `python/pecos-rslib/src/decoder_bindings.rs`.
- At decode time it calls the narrow packer
  `DemCheckMatrix::observables_mask_from_correction`
  (`python/pecos-rslib/src/decoder_bindings.rs:2417-2419`), which does
  `mask |= 1 << i` into a `u64`
  (`crates/pecos-decoder-core/src/dem.rs:629-639`).
- For `i >= 64` that shift panics in debug builds and is masked (`i % 64`) in
  release builds, so observable 64 silently sets the bit for observable 0.

The wide-safe sibling already exists immediately below it:
`observables_obsmask_from_correction` (`crates/pecos-decoder-core/src/dem.rs:646-657`),
which is what the string-registry wrappers use
(`python/pecos-rslib/src/fault_tolerance_bindings.rs:2168`).

Other matching decoders reject wide DEMs loudly instead: Tesseract errors at
construction (`crates/pecos-tesseract/src/decoder.rs:187-198`) and Fusion Blossom
rejects observable index >= 64 at edge insertion
(`crates/pecos-fusion-blossom/src/decoder.rs:689-697`).

## Expected

Either make `DemAwareDecoder` wide (use `observables_obsmask_from_correction`
and widen the result getter), or reject >64-observable DEMs at construction with
the same actionable message the other narrow decoders use. Silent bit-wrapping
is the one outcome that must not remain.

## Severity

`severity:critical` per the repository definition: silently wrong results in
release builds, with no error surfaced. Blast radius is limited to DEMs with
more than 64 observables decoded through `DemAwareDecoder` (the BP+OSD entry
point documented in `docs/workflows/guppy-dem-decoding.md`).

## Provenance

Found by an adversarial design-review panel on
`pecos-docs/design/construction-and-decoding-api-shape.md`, then verified
directly against the code.

### LINKED ISSUE #433: Rename the DEM-construction NoiseModel to say it is a parameter set, not a simulator noise model
## Problem

`pecos.qec.surface.NoiseModel` collides conceptually with the simulator-side noise
*models* while being a different kind of object:

| name | what it is |
|---|---|
| `GeneralNoiseModel` / `DepolarizingNoiseModel` (via `general_noise()`, `depolarizing_noise()`) | executable models that **apply** noise during simulation; fluent Rust-backed builders, 47 and 7 `with_*` methods |
| `pecos.qec.surface.NoiseModel` | a **parameter bag** consumed at DEM construction; a Python dataclass, 30 fields plus derived properties |

The confusion is live rather than theoretical: `docs/workflows/guppy-dem-decoding.md`
now imports `NoiseModel` from `pecos.qec.surface` in a section that also discusses
simulator noise, and since #422 this dataclass is the shared idle-noise vocabulary for
both the Guppy and native-surface DEM routes — so it is no longer a surface-code detail
either.

`NoiseParameters` (or similar) would say what it is.

## Two separable questions

1. **The rename.** ~240 references across ~34 files (source, tests, docs, examples).
   Public, so it needs an alias plus a deprecation ramp rather than a hard rename —
   same discipline as #424.
2. **Whether it should gain `with_*` setters** to mirror `general_noise()`. Recommend
   **no**: it is a dataclass, so `dataclasses.replace(params, p1=...)` already provides
   what `.with_p1(...)` would, and 30 setters would restate the field list a third time
   (fields, docstring, methods) with three places to drift. The conveniences that earn
   their place already exist: `uniform()`, `is_noiseless`, `physical_error_rate`,
   `for_runtime_idle_time_units()`.

Note for whoever picks this up: a "user-facing uses bare setters, low-level uses
`with_*`" rule does **not** hold in this repo. `sim()`'s noise builders are user-facing
and use `with_*`; the Rust-backed `BpOsdBuilder` uses bare. The convention is genuinely
inconsistent, so a naming decision here should be made deliberately rather than by
appeal to an existing rule.

## Why this belongs with #425

#425 already covers aligning the non-idle `GeneralNoiseModel` surfaces with the
structured vocabulary. The DEM-side and engines-side names should be settled **once,
together** — otherwise this gets renamed to `NoiseParameters` now and possibly again
when the unified vocabulary lands. Deliberately kept out of PR #420 for that reason,
and because a 240-site public rename mixed into a large functional PR is hard to review.

Related: #422 (which made this dataclass the shared idle vocabulary), #424, #425.

### LINKED ISSUE #436: pecos-neo importance runner silently drops every injected noise gate except X/Y/Z
The importance-sampling runner discards any noise gate a channel injects that is not a single-qubit Pauli.

`ImportanceRunner::execute_noise_gate` matches only `X`, `Y`, and `Z`, with a bare catch-all that does nothing:

```rust
// exp/pecos-neo/src/sampling/importance_runner.rs:592-608
match gate.gate_type {
    GateType::X => { self.simulator.x(&qubits); }
    GateType::Y => { self.simulator.y(&qubits); }
    GateType::Z => { self.simulator.z(&qubits); }
    _ => {}
}
```

Anything else a noise channel emits -- `RZ` from coherent idle dephasing, `H`, `S`, or any
multi-qubit noise -- is dropped without a warning or error. The simulation runs to completion
and reports results that are quieter than the configured noise model, with nothing indicating
that noise was discarded.

This is unconditional in this runner: it does not depend on how the runner was constructed.

Impact: importance-sampled results are silently wrong for any noise model configured with
mechanisms outside single-qubit Paulis. Because importance sampling is used specifically to
estimate rare logical failures, understated noise is the failure mode that matters most.

Suggested direction: either execute the remaining gate types, or fail loudly when a noise
channel emits a gate this runner cannot represent. Silently continuing is the one option that
should be off the table.

Found while reviewing after-2q idle noise in `exp/pecos-neo`; not addressed there because it
is independent of that change.

### LINKED ISSUE #437: pecos-neo silently discards injected non-Clifford noise gates when no rotation executor is configured
When a noise channel injects a non-Clifford gate and the runner has no rotation executor, the gate is dropped without any diagnostic.

```rust
// exp/pecos-neo/src/runner.rs:2312-2318
// Non-Clifford gates: delegate to rotation executor
other => {
    if let Some(executor) = self.rotation_executor {
        executor(sim, GateId::from(other), gate.angles.as_slice(), qubits);
    }
    // If no rotation executor, silently skip (noise channel injected
    // a gate the simulator cant handle).
}
```

The most direct way to hit this is coherent idle dephasing: `IdleChannel` in coherent mode
emits `RZ(angle)`, so a runner built without rotation support drops every coherent dephasing
event. The configured noise model reports itself as coherent, the run succeeds, and the
results contain no coherent contribution at all.

Impact: results are quieter than the configured noise model, with no signal that anything was
discarded. A user cannot distinguish "coherent dephasing had little effect" from "coherent
dephasing was never applied".

Suggested direction: fail loudly at configuration time when a noise model can emit gates the
selected simulator cannot execute, or at minimum surface a diagnostic on the first discarded
gate. The comment already acknowledges the situation is anomalous; the code should not treat
it as routine.

Related: #436 covers an unconditional version of the same class of problem in the importance
runner.

Found while reviewing after-2q idle noise in `exp/pecos-neo`; not addressed there because it
is independent of that change.

### LINKED ISSUE #438: pecos-neo IdleChannel::from_t1_t2 derives a dimensionally inconsistent quadratic rate and silently clamps short T1/T2
The T1/T2 convenience constructor produces rates that do not match how the rest of `IdleChannel` interprets them.

```rust
// exp/pecos-neo/src/noise/idle.rs:140-151
pub fn from_t1_t2(t1: f64, t2: f64) -> Self {
    // Approximate error rate from T1/T2
    // This is a simplified model
    let linear_rate = 1.0 / t1.max(1.0);
    let quadratic_rate = 1.0 / (t2 * t2).max(1.0);
    ...
}
```

Two separate problems:

1. **Dimensional inconsistency.** `quadratic_rate` is consumed elsewhere as an angle per unit
   time: coherent mode computes `quadratic_rate * duration` as an `RZ` angle, and incoherent
   mode computes `sin(quadratic_rate * factor * duration / 2)^2`. A value of `1/t2^2` has
   units of inverse time squared, so the resulting angle is not a physically meaningful
   function of the requested T2.

2. **Silent clamping.** `.max(1.0)` is applied to the denominators, not the results. For any
   `t1 < 1.0` or `t2 < 1.0` in abstract time units, the derived rate saturates at `1.0` and
   stops responding to the input entirely -- a shorter coherence time produces the same noise
   as `t1 = 1.0`. Nothing reports that the requested value was ignored.

The docstring says "simplified model", which covers approximation but not a rate that does not
track its own input.

Workaround: set `linear_rate` and `quadratic_rate` directly rather than deriving them from
T1/T2.

Suggested direction: derive the quadratic rate with units consistent with its consumers,
and either reject non-positive/very short times explicitly or let the rates scale as given
instead of clamping. Note that `general_builder.rs` also exposes `with_idle_t1_t2`, so any fix
should cover both entry points.

Found while reviewing after-2q idle noise in `exp/pecos-neo`; not addressed there because it
is independent of that change.

ciaranra:
## Summary

Started as idle-gate handling for the Guppy-to-DEM pipeline and grew into an alignment of the
idle-noise vocabulary across `NoiseParameters` (DEM), `general_noise()` (engines) and
`exp/pecos-neo`, plus a correction to what the DEM actually emits.

The three strands:

1. **One vocabulary.** The same rate and the same model dictionary now mean the same thing on
   every surface, with no unit conversion and no mode flags.
2. **The DEM emits the channel it was asked for.** A noise model is a mutually exclusive Pauli
   channel; a DEM is a set of independent mechanisms. Those are two unravelings of the same
   map, and the builder was not converting between them.
3. **One way to ask which observables flipped.** Decoder predictions and sampled ground truth
   now return the same value type, and the superseded spellings are removed rather than kept
   alongside it. Fourteen members answering one question become five. **Breaking.**

The workflow example is the short version — its DEM block and its simulator block are now
character-identical apart from the constructor name:

```python
.with_p_idle_linear(0.01, {"X": 0.25, "Y": 0.25, "Z": 0.5})
.with_p_idle_sin_squared(0.03, {"Z": 1.0})
```

## Behaviour changes reviewers should look at first

These change results or break calls. Nothing here is a silent no-op fix; each one moves numbers
or requires a call-site edit.

| change | impact |
|---|---|
| `general_noise()` is **noiseless by default** | previously applied ten nonzero parameters (`p1`, `p2`, `p_prep`, `p_meas_*`, idle, emission/seepage ratios). `.auto()` restores that preset verbatim. Anything relying on implicit noise now gets silence. |
| **DEM mechanism probabilities changed** | idle *and* gate channels are now converted from the exclusive to the independent unraveling. 1q `6.666667e-04` → `6.671117e-04`; 2q `1.333333e-03` → `1.345946e-03`. The 2q case was ~0.93% of the two-qubit error weight. |
| **neo incoherent quadratic dephasing** | was `sin²(θ)`, now `sin²(θ/2)` — the exact Pauli twirl of its own coherent `RZ(θ)`. Changes results for anyone using incoherent quadratic idle in neo. |
| noise-builder setters renamed | `with_p1_probability` → `with_p1`, `with_meas_probability` → `with_p_meas`, etc. Hard rename, no aliases. |
| `with_p2_idle(probability)` → `with_idle_after_2q(duration)` | different quantity, not just a new name. Old `with_p2_idle(0.01)` with no linear rate now produces nothing; the equivalent is `with_p_idle_linear_rate(0.01).with_idle_after_2q(1.0)`. |
| `with_p_idle_coherent(bool)` → `with_p_idle_quadratic_coherent(bool)` | the flag only ever selected how the *quadratic* rate was read. `with_p_idle_coherent` is now a family setter taking `(rate, model)` over RX/RY/RZ. |
| **importance sampling silently ran unnoised** | `simulation.rs` discarded the configured `NoiseResource` before it reached the runner, so importance sampling through the high-level API applied no noise model at all. Results were near-zero failures with meaningless weights. |
| injected noise gates are never dropped | the importance runner matched only X/Y/Z with a bare `_ => {}` (#436); non-Clifford gates were skipped without a rotation executor (#437). Now: executed where supported, rejected at configuration time with an actionable message, or panic as a broken invariant. |
| **legacy engines idle setters removed** | `with_p_idle_linear_rate`, `with_p_idle_linear_model`, `with_p_idle_quadratic_rate`, `with_p_idle_quadratic_coherent`, `with_p_idle_coherent_to_incoherent_factor`, and both `with_average_p_idle_*` are gone (#424 ramp steps 2-3). The families replace them. **The quadratic conversion is not 1:1**: `with_p_idle_quadratic_rate(r)` becomes `with_p_idle_sin_squared(r * PI, {"Z": 1.0})`, because the legacy rate was cycles-per-time and was scaled by `factor/2 * 2*PI` internally. Copying the number across without the `PI` gives a channel wrong by that factor. |
| **T1/T2 idle derivation changed** | `from_t1_t2` / `with_idle_t1_t2` now use the first-order Pauli twirl with `T2` as total transverse coherence: `rX = rY = 1/(4*T1)`, `rZ = 1/(2*T2) - 1/(4*T1)`, quadratic zero. `T2 > 2*T1` is now rejected rather than silently clamped, and short times are no longer clamped to 1. This moves numbers for every T1/T2 user; a documented example was silently producing maximal noise before (#438). |
| DEM idle config is rate-plus-weight-map | the Rust `NoiseConfig`'s nine per-axis `f64` fields collapse into one `IdleNoiseFamily { rate, weights }` instantiated three times, mirroring `pecos-engines`' sampler shape. An unset map means symmetric X/Y/Z — a property nine flat fields could not have. Emitted DEM text is byte-identical; this is a representation change, not a physics change. |
| **DEM per-axis idle setters removed** | all twelve `with_p_idle_{,x_,y_,z_}{linear,quadratic,quadratic_sine}_rate` setters are gone; the three families are the only public idle API. `NoiseParameters(p_idle_z_linear_rate=...)` now raises `TypeError`. Two of the removed names collided with live `general_noise()` setters meaning different physics. Replacement is `with_p_idle_linear(r, {"Z": 1.0})` etc.; DEM fixtures pin that this reproduces the old output byte-for-byte. |
| neo `with_p_idle_linear(rate)` → `(rate, model)` | `with_p_idle_linear_weights` removed. `exp/` crate, deliberate. |

## Idle-noise vocabulary

Three families on every surface, each named for its own law, each a rate plus a model:

| family | law | model |
|---|---|---|
| `with_p_idle_linear` | `p = rate·m·t` | normalized distribution over X/Y/Z/L |
| `with_p_idle_sin_squared` | `p = sin²(rate·m·t)` | unnormalized multipliers over X/Y/Z/L |
| `with_p_idle_coherent` | `angle = rate·m·t` | unnormalized multipliers over RX/RY/RZ |

Rates are radians per time unit throughout — engines' `2π` and `coherent_to_incoherent_factor/2`
conversions apply only to the legacy `with_p_idle_quadratic_rate` path, which is unchanged.
Mixing a legacy spelling with its family equivalent is a **loud build-time error**; those two
differ by `1.5π` at the old default factor, so silent precedence between them was the bug worth
preventing.

`with_idle_after_2q(duration)` places idle sites on each two-qubit gate operand, matching the
DEM pass, on both engines and neo.

## DEM channel conversion

Independent mechanisms compose — two firing together produce a third error — so writing the
channel's per-Pauli probabilities straight in understates every Pauli. The conversion is exact
and closed-form via the Pauli-channel characters, and runs on **propagated flip-signatures**
rather than Pauli labels, since a DEM mechanism is `(probability, flip-set)` with no Pauli
attached. At that layer undetectable Paulis drop and equal signatures merge by addition, which
collapses most apparent infeasibilities: a 2-qubit gate's 15 Paulis typically become 3
mechanisms.

Exact representability holds iff `pY ≥ pX·pZ` and cyclically. Realistic noise clears this by
orders of magnitude; only an exactly-Y-free channel takes a residual, bounded by `pX·pZ`,
reported as a total-variation distance on `dem.idle_noise_residuals` and warned on at build.

Two pre-existing defects surfaced and are fixed: equal signatures were merged with
`combine_probabilities` (the independent rule) rather than summed, undercounting mutually
exclusive Paulis by `2·p₁·p₂`; and after-2q idle in neo fired only when the gate's Pauli error
also fired.

Prep and measurement are single Bernoulli events and are deliberately **not** converted — a
test pins that they stay exact.

## Controlling the approximation warning

When a channel is not exactly representable, the builder emits a non-negative fit and warns.
The fit reproduces the requested per-Pauli probabilities **exactly** and lets the mismatch fall
on the omitted signature; it is deliberately *not* the total-variation-minimal fit, which would
sit nearer overall only by moving a requested rate (for one worked case, `pX` 0.15 -> 0.1147, a
24% change in the X rate the caller asked for). The reported residual is the true TV distance to
the fit actually emitted -- not a claim of minimality. The trade-off is documented on
`fit_exclusive_signatures` and pinned by a test. `.with_residual_warning_threshold(fraction)` sets the tolerance below which that warning
is not worth raising, as a **fraction of the channel's total error weight** — so one value means
the same thing across a sweep, which an absolute probability cannot. Default `0.0` warns on any
residual; values outside `[0, 1]` are rejected, and a value above `1.0` says explicitly that the
setter is relative, since that is the likely misreading.

The threshold gates the warning only. `dem.idle_noise_residuals` and the audit entry stay fully
populated whatever it is set to, so a quieted approximation remains provable after the fact.
Residuals now also carry `channel_weight` and `relative_magnitude`.

Suppressing warnings wholesale remains `warnings.filterwarnings`' job; this setter encodes a
physics tolerance, not log-noise management.

## One way to ask which observables flipped

**Breaking change.** Eight public members are removed; see the migration table below.

The workflow example used to compare decoder output to ground truth like this:

```python
actual = batch.get_observable_mask(shot) & 1
pymatching_errors += pymatching.decode_syndrome(syndrome).correction[0] != actual
tesseract_errors  += (tesseract.decode_syndrome(syndrome).observables_mask & 1) != actual
bp_osd_errors     += (bp_osd.decode_syndrome(syndrome).observables_mask & 1) != actual
```

Three spellings for one question, because three result types had three shapes. Two defects
followed:

- **`& (1 << i)` cannot fail.** Asking a 2-observable DEM about observable 5 yields `0`, which is
  indistinguishable from "observable 5 did not flip". Off-by-one on an observable index is the
  likeliest mistake once there is more than one observable, and it was silent.
- **"Logical error rate" is ambiguous above one observable.** `predicted != actual`
  (any observable mispredicted) and `predicted[i] != actual[i]` (that observable mispredicted)
  are different numbers. The old loop computed neither in general — it computed observable 0's
  rate under the name "logical error rate", honest only because the example asserts
  `dem.num_observables == 1`.

`ObservableFlips` is one value type returned by both the prediction side and the ground-truth
side, so the loop is now three identical lines and is correct for any number of observables:

```python
actual = batch.get_observable_flips(shot)
pymatching_errors += pymatching.decode_syndrome(syndrome).observable_flips != actual
tesseract_errors  += tesseract.decode_syndrome(syndrome).observable_flips  != actual
bp_osd_errors     += bp_osd.decode_syndrome(syndrome).observable_flips     != actual
```

Its surface is `len`, `[i]`, iteration, `==`, `indices()` and a `mask` escape hatch — the things a
reader already assumes work on a sequence of bools, and nothing more. Asking about one observable
is `predicted[i] != actual[i]`, and it is bounds-checked: `flips[3]` on a 3-observable result
raises `IndexError` naming the index and the count, where the bitmask returned `0`.

Deliberately absent, because each reads clever and then forces the reader to guess what it
returns: a `mismatches()` helper, `^`/`&`/`|` operators, and implicit comparison against `list`
or `int`. Both constructors accept anything integer-like via `__index__` — the accessors this
type replaced returned `list[int]`, and masks routinely arrive as NumPy scalars — but truthiness
is never used, so `from_bits([1, 2, 0])` raises rather than coercing `2`.

### Removed, and what replaces them

Adding the type without removing the alternatives would have left five ways to ask one question,
which is the problem rather than the fix. Every removed member was strictly dominated, so this is
removal rather than redesign:

| removed | replacement |
|---|---|
| `MwpmResult.correction` | `.observable_flips` |
| `MwpmResult.to_list()` | `list(.observable_flips)` |
| `BpResult.to_list()` | `.decoding` |
| `TesseractResult.observable_bits(n)` | `.observable_flips` |
| `TesseractResult.observables_mask` | `.observable_flips.mask` |
| `DemAwareResult.observables_mask` | `.observable_flips.mask` |
| `SampleBatch.get_observable_mask(i)` | `.get_observable_flips(i).mask` |
| `SampleBatch.get_observable_mask_wide(i)` | `.get_observable_flips(i).mask` |

That takes the surface answering this question from fourteen members to five. Each replacement
was measured against the member it replaces **before** the removal, and the captured values are
pinned as fixed expectations in `test_observable_flips.py` rather than recomputed from the new
code. No deprecation shims or forwarding aliases were left behind.

Two element-type notes for anyone migrating. The removed list-returning members yielded `int`
(`[1]`); the replacement yields `bool` (`[True]`). `1 == True`, so comparisons are unaffected,
but sites that serialise or type-check were converted explicitly rather than assumed transparent.
And the narrow/wide split is gone: `get_observable_mask` *raised* above 64 observables while
`ObservableFlips.mask` is arbitrary-precision, so a test that previously asserted the narrow path
raises now simply reads the value.

`SampleBatch` keeps the bulk `observable_flips()` that landed on `dev` in #452 (shots-major
`list[list[bool]]`, real numpy-path rationale). The per-shot accessor added here is
`get_observable_flips(i)`, following the class's existing convention that `get_X(i)` is per shot
and a bare noun is the whole batch.

`BpResult` deliberately gains no such property. Its `decoding` vector is indexed by error
mechanism, not by observable, and five decoders return it with the same `list[int]` shape as
PyMatching's — so `[0]` silently answered a different question. It gets a blunt docstring instead
of a fabricated mapping; per-observable results come from the `from_dem` constructors.

Ground truth is read through the wide path, never the narrow `u64` one, so more than 64
observables are not truncated on either side; a test exercises observable 70 end to end.

## Independent cross review

An adversarial review arm that did not write the code was run over the three areas where a defect
would be silent. It found two real issues, both since fixed and pinned:

**Sampler observable width (severe).** `DemSampler.from_dem_string` hand-rolled a DEM parser that
only read `error(...)` lines, so `logical_observable Lk` and bare `detector Dn` declarations never
reached its counts. The decoder counts declarations; the sampler did not. Because
`ObservableFlips` equality includes length, every shot then compared unequal -- a silent **100%
logical error rate** on the documented sampler-plus-decoder path, with nothing raised. The
duplicate parser is gone; the counts now come from the canonical `parse_dem_metadata`, which also
fixes the identical detector undercount. Two regression tests, mutation-tested.

**"Closest fit" was an overclaim.** Confirmed by brute force: the boundary fit is not
TV-minimal. The math is a defensible choice -- see above -- but the wording was wrong, and is
corrected here and in the code docs.

Reported clean by the same pass: the flip-signature layer (including same-signature collisions
and empty signatures) and the `int` -> `bool` element-type migration.

The review also corrected this PR's own shorthand: the exact representability criterion is
`pI*pY >= pX*pZ`, not `pY >= pX*pZ`. The implementation never used the shortened form.

## Docs

- `docs/workflows/guppy-dem-decoding.md` — five-stage worked example, hand-written repetition
  code, DEM sampling and `sim()` sampling, Tesseract / BP+OSD / PyMatching.
- `docs/user-guide/dem-from-guppy.md` — idle-noise families, stim export, validity domain.
- Design record: `pecos-docs/design/neo-noise-event-emission.md`, including the review panel
  that rejected adding event emission to `NoiseResponse`.

## Issues resolved

Closes #436 — importance runner silently dropped every injected noise gate except X/Y/Z.
Closes #437 — non-Clifford gates silently skipped when no rotation executor is configured.
Closes #433 — `NoiseModel` renamed to `NoiseParameters`, lifted to `pecos`, fluent setters.
Closes #428 — DEM construction reshaped into the house builder pattern.
Closes #430 — `DemAwareDecoder` no longer wraps observable bits past 64.
Closes #438 — the T1/T2 idle derivation is dimensionally consistent and no longer clamps.
Closes #424 — ramp steps 2 and 3; the legacy idle spellings are removed.
Closes #429 — decoder results and sampled truth share one observable-flip type; the redundant
spellings are removed, not deprecated.

## Still open

- #427 — pre-existing on `dev`, not a regression from this PR: the stalling ops live in
  `crates/pecos-hugr`, which this branch does not touch, and `qec-guppy.md` is byte-identical to
  `dev`. Still unisolated; needs its own bisect.
- #458 — two shipped modules (`decode.py`, `reliable_observables.py`) still use NumPy at runtime,
  119 call sites in total. Pre-existing and not introduced here; migrating a decoder's numeric
  core needs its own equivalence testing, so it is filed rather than folded in.
- The Python-side canonical internals still name per-axis rates (419 production references),
  translating to the family form at the Rust boundary. Unreachable from the public API, so this
  is an implementation detail rather than an omission.


docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+```python
+from pecos import depolarizing_noise, selene_engine, sim, stabilizer
+from pecos_rslib.qec import SampleBatch
+
+columns = (
+    sim(rep_code_memory)
+    .classical(selene_engine())
+    .quantum(stabilizer())
+    .qubits(7)
+    .noise(depolarizing_noise().with_uniform_probability(0.01))
```
ciaranra: should use general noise model here

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+given to `sim(...)` — and the default Selene runtime emits no idle gates, so
+idle noise has nothing to attach to on that path. Treat the numbers as coming
+from two different experiments rather than as a like-for-like comparison.
+
+<!--continuation-->
+```python
+from pecos import depolarizing_noise, selene_engine, sim, stabilizer
+from pecos_rslib.qec import SampleBatch
+
+columns = (
```
ciaranra: let's output results first then convert to columns

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+The two paths do not use the same noise: the DEM carries the idle families
+configured in stage 3, while the simulation applies whatever noise model is
+given to `sim(...)` — and the default Selene runtime emits no idle gates, so
+idle noise has nothing to attach to on that path. Treat the numbers as coming
+from two different experiments rather than as a like-for-like comparison.
+
+<!--continuation-->
+```python
+from pecos import depolarizing_noise, selene_engine, sim, stabilizer
+from pecos_rslib.qec import SampleBatch
```
ciaranra: We shouldn't be importing from pecos_rslib directly. SampleBatch might be unclear what it is...

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+returns a per-observable `correction` vector, while Tesseract and BP+OSD return
+an `observables_mask` bitmask.
+
+<!--continuation-->
+```python
+from pecos.decoders import DemAwareDecoder, PyMatchingDecoder, TesseractDecoder
+
+pymatching = PyMatchingDecoder.from_dem(terminal_graphlike_text)
+tesseract = TesseractDecoder.from_dem(source_graphlike_text, preset="fast")
+bp_osd = DemAwareDecoder.from_dem(raw_text, decoder_type="bp_osd")
```
ciaranra: We should have a BpOsdDecoder or a LdpcDecoder...

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+tesseract_errors = 0
+bp_osd_errors = 0
+
+for shot in range(batch.num_shots):
+    syndrome = batch.get_syndrome(shot)
+    actual = batch.get_observable_mask(shot) & 1
+
+    pymatching_errors += pymatching.decode(syndrome).correction[0] != actual
+    tesseract_errors += (tesseract.decode_syndrome(syndrome).observables_mask & 1) != actual
+    bp_osd_errors += (bp_osd.decode_syndrome(syndrome).observables_mask & 1) != actual
```
ciaranra: We should think about having a nice convenience function for each of these decoders that looks nice/same name...

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+    .seed(42)
+    .run(500)
+    .to_shot_map()
+    .to_dict()
+)
+evaluated = dem_build.evaluate_result_columns(columns)
+sim_batch = SampleBatch(
+    [events for events, _ in evaluated],
+    [mask for _, mask in evaluated],
+)
```
ciaranra, line 282: Make a convenience function?

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+
+`to_sampler()` draws detector events and observable flips directly from the
+error model, without simulating the circuit. `get_syndrome()` returns one shot's
+detector bits and `get_observable_mask()` the actual logical flips those shots
+incurred — the ground truth that decoder predictions are scored against.
+
+<!--continuation-->
+```python
+sampler = dem.to_sampler()
+batch = sampler.generate_samples(2000, 1)
```
ciaranra: What is the 1?

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+
+detectors = [
+    Detector("s0_r0"),
+    Detector("s1_r0"),
+    Detector("s0_r0", "s0_r1"),
+    Detector("s1_r0", "s1_r1"),
+    Detector("s0_r1", "m0", "m1"),
+    Detector("s1_r1", "m1", "m2"),
+]
+observables = [Observable("m0")]
```
ciaranra, line 107: These Detectors and Observables should have names to reference?

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+memory errors while making dephasing dominant; its weights are an additive
+probability distribution and must sum to 1. The sine-squared family uses Z only,
+because the sine-law dephasing remnant is Z by nature — its default is symmetric
+across X, Y, and Z, so the single-axis choice is spelled out explicitly. See
+[Idle Noise](../user-guide/dem-from-guppy.md#idle-noise) for the full family and
+model semantics.
+
+`build_dem_from_guppy` returns the DEM together with the audit trail and the
+result-column evaluator used in stage 4b. `DetectorErrorModel.from_guppy` builds
+the same DEM from JSON metadata instead of typed specifications.
```
ciaranra: Why have two things like this?

docs/workflows/guppy-dem-decoding.md
ciaranra: Or... Maybe the name is implied by the indexing... Di... Li... etc...

ciaranra: Should maybe at least put a comment on this... I also wonder if we should just have Detectors(...) and Observables(...) Or DetectorDefs(...) and ObservableDefs(...) that take an array of Detector()/Observable()

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+```python
+from pecos import general_noise, selene_engine, sim, stabilizer
+
+# The same gate noise the DEM was built with, so only the idle treatment differs.
+noise = (
+    general_noise()
+    .with_p1_probability(0.002)
+    .with_p2_probability(0.02)
+    .with_meas_probability(0.02)
+    .with_prep_probability(0.02)
```
ciaranra: We should include idle parameters

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+)
+
+results = sim(rep_code_memory).classical(selene_engine()).quantum(stabilizer()).qubits(7).noise(noise).seed(42).run(500)
```
ciaranra, line 284: We might want to show both how to run guppy... but also maybe get the TickCircuit from the QIS trace and do the same sort of Idle stripping and adding after each two-qubit gate like the DEM generation is doing...

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,293 @@
...
+memory errors while making dephasing dominant; its weights are an additive
+probability distribution and must sum to 1. The sine-squared family uses Z only,
+because the sine-law dephasing remnant is Z by nature — its default is symmetric
+across X, Y, and Z, so the single-axis choice is spelled out explicitly. See
+[Idle Noise](../user-guide/dem-from-guppy.md#idle-noise) for the full family and
+model semantics.
+
+`build_dem_from_guppy` returns the DEM together with the audit trail and the
+result-column evaluator used in stage 4b. `DetectorErrorModel.from_guppy` builds
+the same DEM from JSON metadata instead of typed specifications.
```
ciaranra: Having both build_dem_from_guppy and DetectorErrorModel.from_guppy feels odd... can we unify?

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,322 @@
...
+    general_noise()
+    .with_p1(0.002)
+    .with_p2(0.02)
+    .with_p_meas(0.02)
+    .with_p_prep(0.02)
+    .with_p_idle_linear_rate(0.01)
+    .with_p_idle_linear_model({"X": 0.25, "Y": 0.25, "Z": 0.5})
+    .with_p_idle_coherent(False)
+    .with_p_idle_quadratic_rate(0.03 / math.pi)
+    .with_idle_after_2q(1.0)
```
ciaranra, line 281: do the general_noise().with_idle_after_2q(...) ignore other idles by default like the DEM builder does... should we have and option to not ignore other idles?

docs/workflows/guppy-dem-decoding.md
```diff
@@ -0,0 +1,314 @@
...
+    .with_p1(0.002)
+    .with_p2(0.02)
+    .with_p_meas(0.02)
+    .with_p_prep(0.02)
+    .with_p_idle_linear(0.01, {"X": 0.25, "Y": 0.25, "Z": 0.5})
+    .with_p_idle_sin_squared(0.03, {"Z": 1.0})
+)
+
+dem_build = (
+    DetectorErrorModel.builder()
```
ciaranra, line 161: does the builder have .runtime(...) that accepts Selene-shaped runtime plugins?

crates/benchmarks/benches/modules/noise_models.rs
```diff
@@ -97,8 +97,8 @@ fn bench_depolarizing_noise<M: Measurement>(c: &mut Criterion<M>) {
         // Benchmark mixed gate set (more realistic)
         group.bench_with_input(BenchmarkId::new("mixed", num_gates), &num_gates, |b, &n| {
             let mut noise = DepolarizingNoiseModel::builder()
-                .with_prep_probability(0.001)
-                .with_meas_probability(0.001)
+                .with_p_prep(0.001)
+                .with_p_meas(0.001)
                 .with_single_qubit_probability(0.0005)
                 .with_two_qubit_probability(0.002)
```
ciaranra: this should be using with_p1(...), with_p2(...) etc...

docs/workflows/guppy-dem-decoding.md
ciaranra: Also, are all the parameters of the decoders exposed?

PR #421, opened 2026-08-03
Frontier decoder port (exp/) and decoder-evaluation harness

ciaranra:
## Summary

Adds `pecos-frontier`, a native Rust implementation of the Frontier decoder
(Leverrier & Urbanke, arXiv:2606.20513): pruned ordered dynamic programming that
approximates the logical maximum-likelihood rule by accumulating posterior mass per
logical class (merging degenerate error configurations) rather than searching for a
single most-likely error. This targets qLDPC/DEM decoding problems; the paper reports
state-of-the-art accuracy on the gross code [[144,12,12]] under circuit-level noise
with small retained frontier sizes.

Decision record: the upstream project is a Python package whose engine is a single raw
CPython extension with no standalone C++ target or C API, and its performance-critical
model preparation lives in Python — so this is a native port with the upstream package
used as a verification oracle, not an FFI wrap (unlike Tesseract/PyMatching, which
build standalone C++ from pinned sources).

## What's included (v1, binary model)

- `FrontierDecoder::{from_sparse_dem, from_dem_str}` building columns directly from
  `SparseDem` mechanisms; explicit `FrontierConfig { k, delta, column_order }` with
  fail-fast validation (provisional defaults pending benchmarking). Duplicate indices
  within a mechanism are rejected; `p == 1` mechanisms are supported by exact folding
  into the decode start state.
- Ordered DP core: boundary states keyed by (active residual syndrome, logical label),
  log-sum-exp merging, completed-detector consistency checks, deterministic K/Delta
  pruning (no map-iteration-order dependence; total-order tie-breaks).
- `ObservableDecoder` implementation returning wide `ObsMask` (no 64-observable cap;
  `decode_to_observables` fails loud past 64). `FrontierResult` reports per-label
  retained log masses, total `log_evidence` (upstream semantics), runner-up gap, and
  frontier-size stats.
- No-path decoding failures are errors, never silent fallbacks.
- Lives in `exp/pecos-frontier` (experimental, `publish = false`): the algorithm core is
  oracle-verified, but the crate graduates to `crates/` and the `pecos-decoders`
  meta-crate only after broader validation with real users.

## Verification

- Exact-equivalence property tests: unpruned decoding must reproduce brute-force
  enumeration over all fault subsets (per-label log masses to 1e-9 and argmax) on 33
  seeded random hyperedge DEMs, including reversed column orders.
- Upstream golden-fixture parity: `tests/fixtures/upstream_fixtures.json` is generated
  by the committed script against the upstream Python+C++ package; unpruned per-label
  masses, predicted labels, and log evidence match to 1e-9 across 5 models x 46
  syndromes (including a >64-observable model and a degeneracy case where logical-ML
  disagrees with most-likely-error).
- Guard non-vacuousness was mutation-tested (merge math, close-check, tie-break,
  K truncation each break tests when altered).
- Targeted tests: pruning changes the answer (greedy k=1 wrong, exact right), bitwise
  determinism including ties, fail-loud unexplainable/overpruned syndromes, config and
  DEM validation, trait-object use.

## Accuracy vs upstream (measured)

`score_alpha` suffix-compatibility pruning is now ported (the paper's S = P + alpha*C
scoring; default alpha 0.8 matching upstream), so pruned decoding uses upstream's rule,
not a prefix-mass-only approximation. Golden-fixture parity is enforced for BOTH
unpruned and pruned regimes at 1e-9. A head-to-head on an identical d3 rotated-surface
circuit-noise model (identical deadline-reorder column order, identical 2000 sampled
shots, forward-only, K in {2,4,8,16}) matches upstream's per-shot predictions exactly
in the alpha=0.8 lane at every K; the alpha=0 lane differs on 3-4 shots per 2000 from
equal-score tie ordering only.

Scale parity is also verified on the paper's headline code: the full BB144/gross model
(936 detectors x 8784 mechanisms x 12 observables, deadline order, K=16) decodes
upstream-sampled corpora with EXACT per-shot agreement — 500/500 at p=0.001 and 300/300
at p=0.003, the latter including all 12 failing and 9 no-path shots, i.e. the near-tie
regime where numeric drift would surface. A model containing a probability-1 mechanism
(where upstream silently falls back to its multi-choice engine while this port folds the
forced mechanism into the start state) agrees to zero deviation in labels and per-label
masses. Runtime at BB144 scale is ~3.1x upstream (vs 1.3-1.5x at d3) — upstream's
incrementally-maintained suffix-score tables win as the boundary grows; that is the
documented deferred performance item. Runtime is ~1.3-1.5x upstream's C++ per shot. The
`bridge_ab` example is the committed cross-implementation harness.

## Review

Two independent adversarial Codex arms (diff review + fresh-session semantics attack
against the paper and upstream implementation) ran before marking ready; confirmed
findings (wide-label tie-break word order, duplicate-index semantics, p == 1 support,
pruning-rule disclosure, mass-naming precision) are fixed in the final commits.

## Ordering, committee, and Python API (added after the initial review rounds)

- `deadline_column_order` / `backward_deadline_column_order`: native generation of
  upstream's ordering heuristics, permutation-exact against upstream on golden fixtures.
- `FrontierCommittee`: forward/backward two-leg decoding with upstream's selection rule,
  parity-tested (status, label, direction, evidence) including fixtures where the
  backward leg wins.
- Python bindings in `pecos_rslib_exp`: `FrontierDecoder` / `FrontierCommitteeDecoder`
  with the `pecos_rslib.decoders` API shape (`from_dem` + kwargs, sparse `decode`,
  dense `decode_syndrome`, `decode_batch`), `column_order` presets matching upstream
  vocabulary, and arbitrary-precision `observables_mask` (works past 64 observables).
  Verified by maturin-built extension + pytest (6 tests), reproduced independently.

## Out of scope (follow-ups tracked in the design note)

- The multi-choice (grouped hyperedge) model, the integer max-log ("frontierLite")
  metric mode, and performance work (incremental suffix-score deltas); tracked in the
  design note with reasons.

## Decoder-evaluation work merged in (formerly PR #432)

At the maintainer's request this PR now also carries the decoder-evaluation branch that
grew out of the port. Full history and per-round verification notes are on #432 (closed,
superseded by this PR); summary:

- **Paired DUT/reference comparison**: `SampleBatch.compare_decoders` returns the 3x3
  joint outcome counts (correct / mismatch / decode error per decoder) with Jeffreys
  intervals on the headline proportions — answering "would a stronger decoder rescue
  these same shots?" instead of a bare LER.
- **Shot-corpus save/load**: single self-describing file with whole-file SHA-256
  authentication (a one-field header edit is rejected, not silently absorbed), bounded
  dimensions, embedded exact DEM with loaded-corpus DEM binding, resolved-seed
  provenance, and canonical padding. Adversarially reviewed; the header-tamper defect
  found by that review was reproduced by experiment and then fixed and pinned by
  permanent tests.
- **Two pre-existing `dev` defect fixes** (requested on-PR instead of as issues
  #456/#457): 32-bit dimension-count overflow in all three DEM parsers, and decoder
  errors being folded into logical-error counts (with the parallel path able to score an
  erroring decoder as *correct*); all scoring now fails loud through one unit-tested
  helper, with healthy-decoder counts pinned bit-identical.

Design record for the evaluation work, including two adversarial design-review rounds
and every reversed scope cut: `pecos-docs` `design/decoder-failure-diagnosis.md`.


ciaranra:
Superseded by the restructured stack, now fully merged into dev: #473 (decoder-defaults fixes), #496 (decoder-evaluation harness with authenticated shot corpora), #475 (the trellis engine and BpTrellisDecoder), and #504 (the Frontier parity port as an identity layer over the engine). Everything this PR carried was rebuilt against current dev and landed through those four; the extensive review history here remains the design record for the port and harness work. Closing without merge.

PR #459, opened 2026-08-09
BpTrellisDecoder: BP-guided trellis decoder foundations (telemetry, parity harness, merging, BP scoring)

ciaranra:
## Summary

Foundation layer for the experimental sibling-decoder program (design:
`pecos-docs` `design/coset-decoder-sketch.md`, v2 after a three-arm blind review panel
plus a verification re-review). Stacked on #421 (`frontier-decoder`); rebase onto dev
after that merges. Everything here is additive to `exp/pecos-frontier` and
parity-pinned.

### M0a — telemetry, status envelope, verification instruments

- `FrontierResult` gains `transitions` (candidate branch evaluations), `dropped_states`,
  `dropped_log_mass` (documented as pruning accounting, not a lost-posterior bound), and
  `status: Exact | Pruned{k_capped, delta_pruned}` (flags, no precedence; `NoPath`
  remains an `Err`; no speculative budget machinery). `build_seconds` on decoders.
- **Bitwise snapshot harness**: 128 scenarios (all fixture decodes incl. committee and
  12 NoPath outcomes) captured from pre-change HEAD as `to_bits()` patterns; the test
  asserts bit equality, and regeneration requires an explicit env-var opt-in. This is
  the executable form of the parity contract for all future sibling work.
- First production `debug_assert`s in the crate: column-order permutation and mask
  invariants (close-masks partition touched detectors; final active mask empty), named
  in one place. The final-mask invariant immediately caught forced-only detector bits
  lingering in active masks; they are now excluded at the owning layer (provably always
  zero in every state; the pre-change snapshot confirms numerical inertness).
- `tests/MUTANTS.md`: enumerated manual-mutation checklist — exact quoted edits, named
  killer tests, and documented EQUIVALENT rows with reasons.

### M0b — opt-in indistinguishable-mechanism merging

`FrontierConfig.merge_indistinguishable` (default **false**): mechanisms with identical
detector and observable sets fold into their first occurrence with XOR-combined
probability `p1(1-p2)+p2(1-p1)` — exactly mass-preserving for independent binary
mechanisms (same idea as Tesseract's preprocessing). Default-off because the merge is
mathematically exact but floating-point-path different, and the parity contract is now
bitwise; the snapshot passes untouched. Exactness is tested against the brute-force
enumerator operating on the ORIGINAL unmerged mechanisms (the enumerator knows nothing
about merging, so a wrong formula cannot hide), plus ordering-preservation,
forced-vs-probabilistic non-merging, and committee propagation tests.

### Verification

41 Rust tests + 9 Python tests (extension rebuilt via maturin, run independently);
clippy pedantic `-D warnings` clean on both crates; repo-wide pre-commit clean; all
seven new mutants from MUTANTS.md executed manually (three M0a telemetry mutants, three
M0b merge mutants, one legacy spot-check) — every one killed by its named tests.

### Program context

Two mechanisms from the design were falsified by pre-registered measurements before any
implementation (recorded in the design docs): syndrome-adaptive ordering (boundary-width
inflation on 800/800 BB144 shots at every radius) and imported-DEM automorphism legs
(WL refinement proves the circuit DEM's automorphism group trivial). Next on this
branch: the M3 score-only BP-informed pruning prototype bundled atomically with
promoting `mini_bp` out of `pecos-uf-decoder`.

### M3 bundle — native BP promotion + score-only BP-informed pruning (third commit)

**`pecos_decoder_core::bp`** (new public module, promoted from `pecos-uf-decoder`'s
private `mini_bp`): `BpGraph::{from_dcm, from_sparse_dem}`, min-sum BP returning
per-mechanism posterior LLRs into caller-owned storage via a reusable `BpScratch`,
beliefs only — never corrections. Hardening over the private helper: exact syndrome
dimension validation (the old code silently zero-filled short input), and full internal
scratch reset (the old code relied on the caller clearing a message buffer). Numerics
pinned bit-for-bit across the move by a 6-scenario pre-refactor posterior fixture
covering flooding, serial scheduling, and the EWA path. BP+UF now consumes the public
surface; the private copy is deleted.

**Deliberate breaking change** (pre-1.0 crate): `BpWeightProvider::compute_weights` now
returns `Result` and the trait gains `check_count()`, because the caller audit found the
belief-matching zero-defect fast path could bypass BP validation entirely; dimensions
are now checked before that shortcut. Single in-tree implementor updated in this diff.

**`bp_score_iterations`** on `FrontierConfig` (default 0 = off, bitwise snapshot
untouched): when enabled, each decode runs N min-sum iterations on the shot's RESIDUAL
syndrome (observed XOR forced — the forced-folding discipline), converts posteriors to
clamped probabilities, and rebuilds ONLY the suffix-compatibility tables from the
resulting moments. Masks, ordering, and all branch masses stay DEM-true: BP decides
what pruning keeps, never what mass anything carries (the statistically valid form —
prior-reweighting double-counts syndrome evidence and is explicitly excluded). The BP
graph is built from exactly the post-order, post-filter, post-merge column sequence the
DP consumes, and construction skips BP entirely for unpruned configs. `bp_seconds`
telemetry included so experiments charge BP inside matched work.

Verified: 14 Rust suites across four crates + 10 Python tests re-run independently;
three new MUTANTS.md mutants executed and killed (DEM-priors-instead-of-BP, clamp
removal, fast-path bypass — the last needing one corrected mutation attempt, the
faulty first try documented rather than counted).

Next on this branch: the matched-work experiment on the BB144 corpora deciding whether
BP-informed pruning earns its place.

### BpTrellisDecoder — the named decoder (fourth commit)

The matched-work experiment validated BP-guided pruning decisively on the BB144
p=0.003 corpus (results recorded in the design doc): BP at K=4 beats the static score
at K=16 on both accuracy and time (8 vs 12 failures, 43.7 vs 61.3 ms/shot), and BP at
K=8 reaches within noise of the near-exact floor (2 failures vs 1 at static K=256) at
15x less work. On that basis the mechanism graduated to a named public type:

**`BpTrellisDecoder`** — PECOS's own decoder of the trellis class: an approximate
logical maximum-likelihood (degeneracy-aware, coset-mass) decoder, exact in the
unpruned limit, with BP-guided retention. Named for the algorithm class (trellis) and
the winning mechanism (BP), per the compositional house pattern (`BpUfDecoder`);
project names like "Frontier" stay on ports of those projects. Docs carry the three
honesty qualifiers: optimality is relative to the supplied DEM; the approximation has
no certified discarded-mass bound; BP guides only what pruning keeps, never mass
arithmetic.

It is a thin facade over the shared engine — every configuration is pinned
bit-identical to a hand-mapped `FrontierDecoder` by equivalence tests — with defaults
that ARE its identity and are mutation-pinned: `bp_score_iterations=5`,
`merge_indistinguishable=true`, and the deadline order computed by default (the parity
port deliberately preserves input order; PECOS's own decoder defaults to the good
order). `k=8` default documented as provisional pending broader validation. Python
bindings included; `bridge_ab` gained a BP-iterations argument (used for the
experiment).

Verified: 51 Rust tests + 12 Python tests re-run independently; snapshot untouched;
three default-pinning mutants executed and killed (one first attempt failed to compile
and was redone from MUTANTS.md's exact edit — a non-compiling mutant proves nothing).

### Escalation ladder (fifth commit) — and the program's closing scoreboard

`BpTrellisConfig::escalation_ks` (default empty): on a no-path failure, retry the shot
at each successive K using ladder decoders pre-built at construction. The mechanism was
validated by pre-registered experiment before implementation: on the BB144 p=0.003
corpus, K=2 base + no-path escalation to K=16 achieves the same 2 failures/300 as
single-K=16 at roughly half the average work (48.0 vs 84.9 ms/shot). The experiment
also produced a finding the docs now state plainly: the terminal posterior gap — the
literature's proposed escalation trigger — is vacuous at small K (0/300 shots retain a
runner-up label; pruning deletes every competitor), while no-path is a free trigger
with perfect precision that caught 18 of 19 failures. Only no-path escalates; a
returned-but-wrong prediction never does (no oracle in production) — the honest limit,
documented. Whole-shot telemetry (`transitions`, `bp_seconds`,
`escalation_rungs_used`) accumulates across attempted rungs.

Verified: 57 Rust + 13 Python tests re-run independently; snapshot untouched; all
three documented escalation mutants executed from MUTANTS.md's exact edits and killed
(3, 4, and 4 test failures).

**Mechanism scoreboard for the whole program** (full records in `pecos-docs`
`design/coset-decoder-sketch.md`): syndrome-adaptive ordering — falsified
(boundary-width inflation, 800/800 shots); k-leg ordering committee — falsified at the
oracle bound (a perfect selector loses to single-leg K-escalation at every
matched-work point); automorphism legs — parked (WL refinement proves the circuit
DEM's automorphism group trivial; needs a PECOS producer); streaming — deferred
(premature-posterior counterexample; named prerequisites); grouped columns — parked
(producer-gated); BP-guided pruning — validated and shipped as the decoder's defining
mechanism; no-path escalation — validated and shipped. Remaining before any external
claim: the evaluation campaign (Relay-BP and beam-search baselines, p-sweeps, second
code, Pareto curves).

### Crate split — `exp/pecos-bp-trellis` (sixth commit group)

Owner decision: PECOS's own decoder does not live inside the parity-port crate.
`BpTrellisDecoder`, `BpTrellisConfig`, `TrellisOrdering`, the escalation ladder, their
test suite, and their mutant rows moved to a new `exp/pecos-bp-trellis` crate that
depends on `pecos-frontier` for the engine. `exp/pecos-frontier` is again purely the
Frontier parity port. Exactly two engine items became public to support the facade
(`FrontierDecodeAttempt` and `FrontierDecoder::decode_attempt`, both documented); no
engine fields were exposed and no logic moved or duplicated. The Python API is
unchanged. Bitwise snapshot and upstream fixtures byte-untouched. Verified: 47+11 Rust
tests, 13 Python tests, cold clippy pedantic on all three crates, `cargo doc` link
check, repo-wide pre-commit.

Two supporting fixes in the same push: `BpOsdBuilder`/`BpLsdBuilder` Python bindings
now accept a per-column error channel (the Rust core always supported it; a scalar-only
binding handicaps any circuit-DEM baseline — found by the evaluation campaign; three
new tests prove priors steer the decode and mismatched lengths fail loudly), and the
release-profile unused-parameter warnings in `debug_assert_model_invariants` are fixed
with an explicit cfg pair.

### Evaluation campaign (phase 1, complete — record in pecos-docs `design/bp-trellis-evaluation-campaign.md`)

Six corpora (BB144 p in {0.001, 0.002, 0.003} both memory bases; rotated_surface_d5 at
p in {0.003, 0.005}), every decoder consuming identical DEMs with per-mechanism priors,
Jeffreys 95% intervals. Honest headline: BP-guided pruning and no-path escalation both
revalidated at scale, but on BB144 circuit noise BP+OSD-CS-7 leads the Pareto front
(BpTrellis needs K=64 and ~8x the time to reach CI-overlapping accuracy), while
BpTrellis dominates every BP-family baseline on the surface-d5 corpora at comparable
cost. Also surfaced and FIXED on owner instruction: PECOS's Relay-BP shipped default left
gamma0 disabled, costing ~100x in LER on BB144 (99/1000 -> 1/1000). The fix adds a
shared `pecos_relay_bp::DEFAULT_GAMMA0 = 0.65` used by both the Rust builder and the
Python binding (drift-proof), with a defaults regression test; post-fix shipped
defaults verified 0/400 on the exposing corpus. Phase 2 ran the external beam-search
oracle (arXiv:2512.07057; built from source, CC BY-NC-SA so comparison-only forever):
it confirms the BB144 picture (W=64: 1/1000 at 30 ms; 0/1000 on memory_X) while
BpTrellis keeps the accuracy lead on surface-d5. The three "saturated" BB144 trellis
failures were diagnosed as early-pruning artifacts, not ML losses: all three decode
correctly at K=256 with the true label winning by ~42 nats, and pure-mass retention
(alpha=0) no-paths them even at K=512 — the suffix-compatibility score is
load-bearing. Full record: pecos-docs `design/bp-trellis-evaluation-campaign.md`.

### Trellis engine layering (seventh commit group)

Owner decision: PECOS's decoder must not depend on the parity-port crate. The trellis
DP engine now lives in a neutral `exp/pecos-trellis` crate (`TrellisDecoder`,
`TrellisConfig`, `TrellisResult`, ...), moved verbatim (identifier renames plus
neutralized validation strings only — verified by reverse-rename diff, where the only
addition is the re-expressed BP-state-independence test). `exp/pecos-frontier` is now
pure port identity: type aliases onto the engine types, `FrontierCommittee`, the
bridge_ab harness, and the byte-frozen parity fixtures/snapshot — all untouched and
green. `exp/pecos-bp-trellis` depends only on the engine. The design had an
independent adversarial review before implementation (record + companion in
pecos-docs `design/trellis-engine-crate-layering.md`); its seven must-fixes (alias
completeness, orphan-rule impl relocation, unit-test ownership split, committee
error-helper duplication, identifier-accurate MUTANTS anchors incl. one pre-existing
stale anchor, dependency plumbing, textual-drift policy) are all incorporated.
Verified independently: cold clippy pedantic on four crates, 43+4+11 Rust tests,
13/13 Python tests unchanged, `cargo check --workspace`, doc link check, repo lint.

### Independent adversarial review + fixes (eighth commit group)

A blind Claude review arm attacked the five commits from the crate split through the
layering. Its confirmed findings, all fixed here: `gamma0` is the global memory-BP
gate for the entire relay ensemble (per-leg disordered strengths are only consulted
when it is set) — the docs previously told a wrong "pre-relay leg" story and the
generic `CheckMatrixDecoder` relay path still built the inert configuration; the
relay RNG is construction-seeded and never reseeded per decode, so reused-decoder
outcomes are decode-order dependent (now documented on seed/decoder docs; per-decode
reseeding left as an explicit open decision); the shared `DEFAULT_GAMMA0` const had
captured `RelayBpBuilder`'s rustdoc block (builder was undocumented — doctest listing
proved it); the layering's BP-state-independence test was unfalsifiable and is
replaced by a real pin (committee legs' BP graph/scratch addresses pairwise distinct
via a `#[doc(hidden)]` accessor) plus an error-text equality test across the
committee and direct paths; a deterministic behavioral pin now locks the default
(memory on: 8/8 syndromes converge and seeds are live; memory off: 5/8 and seeds
inert); Python `.pyi` stubs updated for the `error_rate` unions; unused `rayon`
dependency removed; provenance docs across the three trellis crates reconciled
(engine is PECOS-native, parity pinning is pecos-frontier's property). Verified
independently: cold clippy pedantic on six crates, 53+3+45+11 Rust tests, 41 Python
binding tests, workspace check, repo lint.


ciaranra:
Superseded by #475 (trellis engine + BpTrellisDecoder, including the telemetry, snapshot-parity, merging, BP promotion, and escalation-ladder work this PR carried) and #504 (the Frontier port layer), both now merged into dev. The mechanism-selection research record and per-commit review history here remain the reference for how BpTrellisDecoder's defaults were chosen. Closing without merge.

Issue #458, opened 2026-08-09
Remove NumPy from all non-test code; keep the native layer NumPy-compatible


ciaranra:
## Summary

NumPy is used in non-test code across the repository. The standing rule is that PECOS owns its
own numeric primitives; NumPy belongs in the test suite as an oracle, not anywhere else.

The native layer should stay **NumPy-compatible where that makes sense** -- same names, same
argument shapes, same mental model -- so migration is mechanical and nobody has to relearn an
API. Compatible in shape, not dependent in fact.

## Extent

13 non-test files, 224 call sites.

| area | file | sites |
|---|---|---:|
| shipped library | `python/quantum-pecos/src/pecos/qec/surface/decode.py` | 114 |
| shipped library | `python/quantum-pecos/src/pecos/qec/reliable_observables.py` | 6 |
| examples | `examples/surface/dem_decomposition_diagnostics.py` | 28 |
| examples | `examples/surface/native_dem_threshold_sweep.py` | 25 |
| examples | `examples/surface/graphlike_dem_projection_benchmark.py` | 5 |
| examples | `examples/surface_code_threshold.ipynb` | 16 |
| examples | `examples/surface_code_thresholds.ipynb` | 15 |
| examples | `examples/surface_code_experiments.ipynb` | 4 |
| examples | `examples/surface_code_noisy_decoding.ipynb` | 1 |
| examples | `examples/measurement_throughput_benchmark.ipynb` | 1 |
| scripts | `scripts/compare_meas_sampling_pipeline.py` | 2 |
| docs | `docs/user-guide/parallel-execution.md` | 6 |
| docs | `docs/development/parallel-blocks-and-optimization.md` | 1 |

Docs matter as much as code here: the user guide teaches the pattern, so leaving NumPy in a
snippet propagates it into user code.

## Two independent defects

**1. The architecture rule.** Shipped library code, examples, scripts, and docs should all use
the native layer.

**2. NumPy is an undeclared dependency.** `quantum-pecos` declares no NumPy requirement in
`dependencies` or in any optional-dependency group, yet imports it unguarded at module scope
(`decode.py:54`, `reliable_observables.py:42`, plus a function-local import at `decode.py:3473`).
It resolves today only because NumPy arrives transitively via `pytket` and `selene-sim`. Nothing
is broken right now, so the impact is latent: if either drops NumPy, both modules fail at
**import** time with `ModuleNotFoundError`. Completing (1) makes this moot.

## What the native layer already provides

[...]

PR #462, opened 2026-08-09
Native numeric layer: unsigned dtype constructors and NumPy-parity helpers (#458 step 1)

ciaranra:
## Summary

Step 1 of #458: close the native numeric layer's gaps against what the 224 NumPy call sites in
non-test code actually use, so the migration itself can be mechanical. No call site is migrated
here.

The single blocker was an inconsistency, not a missing capability: `Array` fully supports
unsigned integers (`DType::U8`, `ArrayData::U8`, and `astype(dtypes.uint8)` all worked), but the
constructors' dtype allowlist predated that support and rejected them. `np.uint8` alone is 143 of
the 224 sites.

## What changed

- **Constructors accept unsigned dtypes.** `array`, `asarray`, `zeros`, `ones`, `arange`, and
  `Array(...)` take `uint8/16/32/64` under both `uintN` and `uN` spellings. `full`/`empty` do not
  exist in this layer, so there was nothing to widen.
- **Out-of-range values raise instead of wrapping.** The pre-existing `x as u8` conversion
  silently wrapped 256 to 0; conversion is now checked end to end (`OverflowError` naming value
  and dtype).
- **Unsigned reductions were silently wrong and are now right.** `sum`/`max`/`min` routed dtype
  kind `"i" | "u"` through an i64 extraction, so unsigned buffers were reinterpreted; `uint64`
  values above `i64::MAX` now reduce exactly. `any`/`all` rejected unsigned arrays outright
  (found by probing the newly reachable range after the constructor widening) and now accept
  them, oracle-checked against NumPy.
- **New NumPy-parity helpers**: `bool_`, `nonzero`, `zeros_like`, `issubdtype`/`integer`, and
  `Array.tolist()`; `asarray` and all of the above re-exported from `pecos`.
- **Error messages tell the truth**: each constructor's `Supported:` list matches what it
  accepts, asserted in tests so it cannot drift.
- `__array_interface__` typestring for `uint8` canonicalized to `|u1`; objects advertising the
  array interface no longer fall through to sequence parsing on error, so precise dtype errors
  surface instead of being masked.

## Verification

- clippy `-D warnings` clean (cold on changed code), `cargo fmt --check` clean.
- `pecos-rslib` tests: 1123 passed (93 of them new, in `test_native_numeric_gaps.py`).
- `pecos` suite: 2130 passed; `qec` suite: 1147 passed.
- Two-pass `pre-commit` clean, re-run after merging `dev` so the #460 NumPy ban applies: the new
  test file's NumPy usage goes through the test exemption as designed.
- Every guard mutation-tested: constructor dispatch, range checks, cast/indexing/arithmetic/
  comparison over the unsigned range, buffer export, checked `uint64` sum overflow, helper
  semantics against the NumPy oracle, and the `pecos` re-exports. 38 observed failing cases
  across mutation batches, plus 6 for the `any`/`all` fix.

NumPy appears in tests only, as the oracle for exactly these comparisons.

## Notes for review

This is numerics; an independent adversarial review arm is queued before merge. The
`any`/`all` gap is the instructive one: it was invisible to the implementation's own tests and
surfaced only by sweeping every `num` function over the newly-constructible unsigned range --
widening a constructor makes downstream helpers reachable by values nobody re-checked.


ciaranra:
Hold for the fix round: the queued adversarial review returned eleven findings, and all eleven
reproduced by execution. Highest severity: big-endian unsigned buffers read without byte-swap,
native-Array and NumPy-scalar inputs bypassing the checked conversion (silent wrap), a Rust panic
crossing the pyo3 boundary on negative-stride views, and a signed-path regression (`dtype="i8"`
now means int8 where dev followed NumPy's 8-byte convention). One finding (0-d axis handling) is
pre-existing dev behavior on the signed path and is deliberately kept consistent rather than
changed. Fixes with per-finding regression tests and mutation checks are in progress; the PR
should not merge until they land and the review arm re-verifies.

ciaranra:
Fix rounds complete and pushed (`cf81e7ea2`). All eleven review findings addressed:

- Findings 1-9 and 11 fixed with per-finding regression tests and mutation checks. The
  orchestrator re-ran every one of the review's reproductions against the rebuilt extension
  before commit.
- Finding 10 (0-d axis) went further than the review asked: 0-d reductions now match NumPy for
  both signed and unsigned (dev previously raised for both; that divergence is gone).
- The endianness fix was widened to ALL dtypes at the shared ingest layer, closing the
  pre-existing dev corruption tracked in #463 (big-endian ints, floats -- previously read as
  byte-swapped garbage / denormals).
- The dtype spelling table resolving the `"i8"` collision: byte-count typestrings keep NumPy's
  meaning (`"i8"` = 8-byte int, as on dev); long names (`"uint8"`) and `dtypes.*` objects carry
  PECOS bit-width meaning.

Cold verification on the final tree: clippy 0 (`-D warnings`), fmt clean, rslib 1439 passed,
pecos 2130, qec 1147, pre-commit two passes exit 0 (17 hooks).

Remaining before merge: the adversarial arm re-verifies the fixed tree (queued behind its
current task). The hold stands until that lands clean.

ciaranra:
Re-verification complete; hold lifted from this side.

The re-review confirmed all eleven original findings dead (re-ran each reproduction) and swept
the fix-created surface clean across 2,600+ executed cases -- route-equivalence at boundary
values across all four ingest paths, endianness including complex round-trips with NaN/inf/signed
zero, negative and zero strides, the spelling table against dev, Pauli-cast arms, and byte-hash
identity with dev for native-endian inputs.

It found one new High: 0-d arrays with the DEFAULT axis (`axis=None`) still took the legacy
dispatch, where a 0-d buffer extracts as empty -- `num.sum(0-d)` silently returned 0 (pre-existing
on dev, not fix-introduced; the fix rounds had repaired only the explicit-axis forms). Fixed at
the shared gate (`8b3b6ae74`): 0-d now routes to the scalar path for every axis form. Oracle
tests across int8/int64/float32/complex128/uint8/bool added; mutation of the gate kills exactly
the five new cases. rslib suite: 1445 passed; clippy/fmt/pre-commit clean.

From this side the PR is ready for maintainer review: three review rounds total, all findings
executed rather than assumed, and every behavioral delta vs dev enumerated in the PR discussion.

PR #472, opened 2026-08-10
Guard archive entry paths and migrate off the abandoned sevenz-rust

ciaranra:
## Why

`cargo deny` began failing on every branch after RUSTSEC-2026-0245 was published: a path-traversal in `sevenz-rust`, which the workspace depends on at `sevenz-rust = "0.6"`. The advisory has **no fixed version** — the project is abandoned and its repository deleted.

Investigating the advisory surfaced a more direct problem than the dependency itself. The advisory faults `sevenz-rust`'s convenience extractors for joining archive entry names onto the destination without validation. PECOS does not call those; it drives `SevenZReader` directly — and **reproduced the same flaw in its own callback**:

```rust
let file_path = dest.join(entry_name);   // entry_name is untrusted archive data
```

An archive entry named `../../../../home/user/.bashrc` or `/etc/cron.d/pwn` would therefore be written outside the extraction directory.

The other two extraction sites in the crate were already correct, which is what makes this one an outlier worth fixing rather than a policy question: the zip path uses `zip`'s `enclosed_name()`, and the tar path relies on `unpack`, which validates internally.

## What changed

- **`pecos_build::extract::contained_entry_path(dest, entry_name)`** — resolves an entry name inside a destination, rejecting absolute paths, `..` components, and empty names. It normalizes `\` to `/` first, so an archive authored on Windows cannot smuggle `..\..` past a Unix-only component walk. Placed in `extract.rs` beside the other archive utilities, with a doc note on why only readers lacking their own guard need it.
- **The CUDA Windows-installer extraction** now resolves every entry through that helper before creating anything on disk.
- **A documented `cargo deny` ignore for RUSTSEC-2026-0245**, stating precisely why PECOS is not exposed: the affected helpers are unused, and the one call site validates entry names itself. The comment points at the call site so the exception can be re-audited rather than trusted.

## Verification

- Six unit tests on the helper covering legitimate nested names, Windows separators, `..` escapes (four spellings including backslash forms), absolute paths, and empty names.
- **Mutation-tested**: removing the `..` rejection makes `parent_components_are_rejected` fail; restored and confirmed byte-identical.
- **Behavioural probe** against realistic attack names — `../../../../home/victim/.bashrc`, `/etc/cron.d/pwn`, `nvcc\..\..\..\evil.dll`, `cudart/../../escape.so` all blocked, `nvcc/bin/nvcc` still allowed.
- `cargo deny --locked --all-features check advisories bans sources` (the exact CI invocation): `advisories ok, bans ok, sources ok`.
- `cargo test -p pecos-build`, `cargo clippy --locked -p pecos-build --all-targets -- -D warnings`, `cargo fmt --all --check`, and `just lint` all clean.

Note: `cargo deny check licenses` fails, both here and on unmodified `dev` — there is no `[licenses]` section, so cargo-deny's default applies. CI does not run that check, so it is untouched and out of scope here.

## Follow-up left open

The dependency is still abandoned. Replacing it (a maintained fork, or dropping 7z support and asking Windows users to install CUDA themselves) is a dependency-policy decision worth making deliberately rather than inside a security fix; the ignore comment flags it. Separately, none of the three CUDA downloads carry a `sha256`, so archive integrity currently rests entirely on TLS — worth its own look.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -484,7 +484,7 @@ dependencies = [
  "bitflags 2.13.0",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -502,15 +502,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
 dependencies = [
  "bit-vec 0.6.3",
-]
-
-[[package]]
-name = "bit-set"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f0481a0e032742109b1133a095184ee93d88f3dc9e0d28a5d033dc77a073f44f"
-dependencies = [
- "bit-vec 0.7.0",
 ]
 
 [[package]]
@@ -536,12 +527,6 @@ name = "bit-vec"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
-
-[[package]]
-name = "bit-vec"
-version = "0.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2c54ff287cfc0a34f38a6b832ea1bd8e448a330b3e40a50859e6488bee07f22"
 
 [[package]]
 name = "bit-vec"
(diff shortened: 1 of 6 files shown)


ciaranra:
Follow-up commit takes the real fix instead of a suppression: `pecos-build` now depends on `sevenz-rust2` 0.21, the maintained successor named by RUSTSEC-2026-0245, so the advisory leaves the tree entirely and the `deny.toml` ignore added earlier in this PR is removed. No `osv-scanner.toml` entry is needed either.

`ArchiveReader::open` replaces `SevenZReader::new`, which drops the explicit file-length argument; `Password::empty()` and `for_each_entries` are unchanged. The containment check stays as defence in depth. The dependency is taken with `default-features = false`: the defaults add an encoder plus aes256/bzip2/ppmd and the `util` convenience wrappers (the module the advisory is about), none of which the installer path uses, while LZMA/LZMA2 and the BCJ filters are unconditional. Net effect on the lock is six fewer crates.

Verified: `cargo deny --locked --all-features check advisories` reports `advisories ok` (it failed on `dev` before this), `cargo clippy -p pecos-build --all-targets` is clean on a cold rebuild of the changed file, `./scripts/dependency-integrity-check.sh` passes, and `just lint check` passes.

PR #476, opened 2026-08-10
Add fill/flatten/ravel/reshape to Array (#470)

### LINKED ISSUE #470: Array lacks fill/flatten/reshape/ravel, needed for the remaining NumPy migration
## Summary

Migrating the first files off runtime NumPy (#458 step 2) surfaced shape-manipulation methods
that `Array` does not have. These are not blockers for the files migrated so far, but the
remaining ~200 sites in `decode.py`, the surface examples, and the notebooks will need them.

| missing | NumPy equivalent | encountered where |
|---|---|---|
| `Array.fill(value)` | `ndarray.fill` | `scripts/compare_meas_sampling_pipeline.py` clears a reused syndrome buffer between shots |
| `Array.flatten()` / `ravel()` | `ndarray.flatten` | flattening a 2-D detection-event array before a batch decode |
| `Array.reshape(shape)` | `ndarray.reshape` | expected wherever detection events are viewed shot-major vs flat |

Verified absent by execution against `dev` (post-#462): `hasattr(Array, name)` is `False` for
`fill`, `flatten`, `reshape`, and `ravel`.

## Workarounds used, and why they are not the answer

The 2-D flatten became a row-major comprehension:

```python
flat = [value for row in asarray(values, dtype=dtypes.uint8) for value in row]
```

That is fine for one site but is O(n) Python for something the layer should do natively, and it
loses the `Array` type on the way out. The `fill(0)` case was left as a reported gap rather than
rewritten as slice assignment, so the requirement is recorded rather than papered over.

## Suggested scope

Add `fill`, `flatten`/`ravel`, and `reshape` to `Array` with NumPy-compatible signatures and
semantics (including whether the result is a view or a copy -- state it explicitly in the
docstrings, since NumPy's `flatten` copies and `ravel` may not). NumPy stays the test oracle.

Blocks the later, larger steps of #458; not blocking step 2.

ciaranra:
## Summary

Closes #470. Adds `fill`, `flatten`, `ravel`, and `reshape` to `Array`, the shape-manipulation
methods the remaining #458 migration needs: `decode.py` (113 sites) and the surface examples
(53) work with 2-D detection-event arrays, and without these every site needs a hand-rolled
comprehension -- O(n) Python that also drops the `Array` type. One such workaround was already
documented in the migration pattern; it is now replaced with the real call.

All 15 `ArrayData` variants are covered, including `Pauli` and `PauliString`.

## Semantics, and two honest divergences from NumPy

| method | behavior |
|---|---|
| `fill(value)` | in-place, returns `None`, scalar only; routed through the constructors' existing checked conversion so range/type errors match #462's behavior exactly |
| `flatten()` | independent 1-D C-order copy |
| `ravel()` | independent 1-D C-order copy -- **NumPy may return a view here; PECOS always copies** |
| `reshape(shape)` | tuple or varargs, one `-1` inferred; **also returns an owned copy** |

`Array` owns its buffer and has no view/lifetime representation, so both divergences are
structural. Both docstrings say "copy" rather than claiming view semantics -- an honest copy
beats a silently-copying "view". Note for later work: `reshape` copying means it is not free in
a hot path; #458 step 4 should measure rather than assume.

`fill` deliberately reuses the existing checked-conversion path rather than adding range checking
of its own -- #462's review found four ingest routes disagreeing at the boundaries, and this does
not become a fifth.

## Verification

Independently swept against NumPy as the oracle: 5 dtypes x 3 dimensionalities x 4 methods,
**zero mismatches**; copies confirmed independent (mutating a `flatten`/`ravel` result leaves the
original unchanged); `fill(256)` and `fill(-1)` on uint8 both raise `OverflowError`; reshape
rejects count mismatch, two `-1`s, and uneven inference, each naming the real numbers.

clippy `-D warnings` clean; fmt clean; pecos-rslib 1520 passed; qec 1488 passed; pre-commit two
passes exit 0. Mutation table in the implementation record covers the element-count check, `-1`
inference, the fill range guard, and copy semantics.

Built on `ndarray`'s own `fill`, `as_standard_layout().into_owned()`, and
`into_shape_with_order` -- no hand-rolled row-major index arithmetic, which is where off-by-one
and non-contiguous bugs live.


ciaranra:
Fix round complete (`86196ab92`). All five findings from the independent pass are addressed, each
verified by execution before commit:

| finding | now |
|---|---|
| `int8.fill(3.0)` leaked `AttributeError` from `__index__` | `TypeError: value 3.0 is a float; an int8 array requires an integer -- cast explicitly` |
| **the constructors leaked the same error** | `array([1.5], dtype=uint8)` now raises the identical typed error -- fixed at the shared conversion layer, so every ingest route agrees |
| `reshape([2, 2])` rejected | list, tuple, and varargs forms all accepted |
| `uint64.fill(2**64)` message omitted the value | `OverflowError: value 18446744073709551616 is out of range for uint64` |
| `-1`-only inference; `fill` rejecting NumPy's truthiness/string coercion | documented as deliberate divergences |

NumPy's behavior was **not** adopted for the float case: it accepts `fill(3.75)` on an integer
array and silently truncates to `3`. Silent data loss is what this layer exists to prevent, so
PECOS raises and asks for an explicit cast. Same reasoning for `bool.fill('x')`, which NumPy
coerces to `True`.

The constructor leak is the notable find: fixing `fill` alone would have left two ingest routes
disagreeing at the same boundary -- the exact defect class #462's review turned up four instances
of. It is now one path.

Verification: clippy `-D warnings` clean; fmt clean; pecos-rslib **1539** passed; qec **1489**;
docs **271** passed, 0 failed; pre-commit two passes exit 0. 0 behind dev.

PR #479, opened 2026-08-11
Verify cuQuantum archive checksums and stop skipping verification silently

ciaranra:
Follow-up to the archive path-traversal fix (#472), which noted that none of the CUDA-family downloads carried a checksum.

## What was actually wrong

Three distinct problems, only the first of which was the one originally suspected:

1. **cuQuantum downloads were unverified — needlessly.** The code carried `sha256: None` with the comment *"NVIDIA does not publish SHA256 checksums for cuQuantum downloads."* That is incorrect: NVIDIA publishes a release manifest at `.../compute/cuquantum/redist/redistrib_<release>.json` with a SHA256 for every platform and CUDA variant.
2. **A cached archive was never verified at all.** Verification sat inside the `else` branch of `if archive_path.exists()`, so a checksum was only ever checked on the download that created the file. A cached archive — which persists in `~/.pecos` between runs, and which a later run happily reuses — was trusted purely because some earlier run had left it there.
3. **Absent checksums failed open, silently.** cuQuantum and CUDA used `if let Some(hash)`, and cmake printed `Skipped (checksum not available)` and returned `Ok(())`. A missing digest was indistinguishable from a passing one in every case.

## What changed

**cuQuantum — now genuinely verified.** `sha256` is no longer optional; the four published digests (linux-x86_64 and linux-sbsa, CUDA 12 and 13) are recorded with the manifest URL and refresh procedure in a doc comment. Any platform/CUDA combination NVIDIA does not publish is rejected by `checksum_for` with the supported list in the message.

**Verification now runs on every install**, cached or freshly downloaded, in both the cuQuantum and CUDA installers.

**cmake fails instead of skipping.** Every archive it can select has an entry, so a miss means the table and the platform matching have drifted — a bookkeeping mistake that must not silently disable integrity checking.

**Windows cuQuantum now fails with the reason.** The entire `redist/cuquantum/windows-x86_64/` path returns 404 upstream (verified for both CUDA 12 and 13, and for the directory itself), so the URL this code constructed could never have succeeded. It now reports that NVIDIA publishes no Windows archive for this release instead of surfacing an opaque download error.

**CUDA is documented as unverified, loudly.** NVIDIA publishes only MD5 for the CUDA *local installers* — there is no `sha256sum.txt` beside the `md5sum.txt`. Rather than invent a digest, the installer prints an unmissable warning naming the exposure, and the field documents both proper fixes (below).

## Verification

- The recorded `linux-x86_64` CUDA 12 digest was checked against the real artifact: downloaded all 192,439,484 bytes and computed `77138174ece7a9a9108e2d44b8ea1e1f2e29874544ea940e13c0d607ad15a358` — matching both the manifest and the constant, and the manifest's recorded size. The digests are transcribed from the manifest, not self-computed, so a substituted archive cannot certify itself.
- Ten unit tests, including a well-formedness guard (64-char lowercase hex) that mechanically catches transcription damage on future refreshes, and a duplicate-entry guard.
- **Mutation-tested**: corrupting a digest fails the well-formedness test; making the lookup ignore the CUDA major fails the unpublished-combination test. Restored byte-identical after each.
- `cargo test -p pecos-build`, `cargo clippy --locked -p pecos-build --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo deny --locked --all-features check advisories bans sources`, and `just lint` all clean.

## Follow-ups, each needing a call that is not mine

- **CUDA integrity.** Either verify NVIDIA's published MD5 (sound against substitution in this threat model — it needs a second-preimage, not a collision — but requires an MD5 implementation, i.e. a new dependency), or install from the CUDA **redistributable** components, which publish SHA256 in `redistrib_<version>.json`. The second is the better design and is nearly free functionally: this installer already downloads a 4.1 GB local installer and discards everything except nvcc, cudart, and cublas, which redist ships separately. It would also remove the only use of the abandoned `sevenz-rust` dependency and its RUSTSEC-2026-0245 exception, since the Windows `.exe` extraction path would no longer be needed.
- **Windows cuQuantum support** is currently dead upstream; whether to keep the branch, document the manual route, or drop the platform is a support decision.


ciaranra:
Digest verification is now complete: **all four** recorded checksums have been checked against the real artifacts, not just transcribed from the manifest.

The PR description understated this — it reported only the `linux-x86_64` / CUDA 12 digest as verified. Each of the remaining three was downloaded in full and hashed:

| Archive | Result |
|---|---|
| `cuquantum-linux-x86_64-25.11.1.11_cuda12-archive.tar.xz` | matches (192,439,484 bytes) |
| `cuquantum-linux-x86_64-25.11.1.11_cuda13-archive.tar.xz` | matches |
| `cuquantum-linux-sbsa-25.11.1.11_cuda12-archive.tar.xz` | matches |
| `cuquantum-linux-sbsa-25.11.1.11_cuda13-archive.tar.xz` | matches |

Each digest still originates from NVIDIA's published manifest rather than from the local download, so a substituted archive cannot certify itself; the download only confirms that the constant in the source matches the artifact NVIDIA actually serves. A transcription error in any of the three unverified entries would have broken cuQuantum installation on that platform, so this was worth closing before review rather than after.

Separately confirmed while reviewing: every platform branch in cmake's `archive_name` match has a corresponding `CMAKE_CHECKSUMS` entry (linux x86_64 and aarch64, macOS universal, Windows x86_64 and arm64), so the new hard error in place of the silent skip is unreachable for supported platforms and cannot turn a working install into a failure.

ciaranra:
Pushed fixes for the two regressions this PR itself introduced, both found by independent adversarial review and both agreed on by two review arms.

**A failed verification was a permanent dead end.** Making verification mandatory on cached archives meant a partial download — network drop, Ctrl-C, killed process — left a file that every later run found, reported as a mismatch, and refused, with no path printed and no `--force` recovery (`--force` clears the install directory, not the sibling cache). The previous behaviour silently used the bad archive; this PR turned that into a wedge. Now:

- Downloads land on a `<filename>.part` sibling and are renamed into the cache **only after** the checksum matches, so nothing is ever published at the cache path unverified. That closes the partial-file, empty-file, and concurrent-writer cases together rather than one at a time.
- A cached archive that fails verification is discarded and re-fetched once before erroring — the same recovery `download.rs` and the cmake installer already perform. Not a new pattern, just the one that was already the house convention.

**cmake's new hard error sat inside the retry loop.** A table/platform drift is decided entirely by constants, so no retry could change it — yet it would have cost five downloads and roughly 150 seconds of backoff before telling anyone. The digest lookup now happens before the first download, in `expected_checksum`, and `verify_checksum` takes the resolved `&str`. Two tests pin it: every archive `get_download_url` can select has a recorded checksum, and an unrecorded name fails immediately.

**Verification also no longer reads the whole archive into memory.** Since this PR made verification run on every install rather than once per download, the previous `fs::read` allocated ~200 MB every time, worst on the memory-constrained SBSA target. It now hashes in 64 KiB chunks. The review suggested reusing an existing streaming helper in the LLVM installer, but on inspection that one also uses `fs::read`, so there was nothing to reuse — a local incremental loop was the smaller change. Proved equivalent rather than assumed: streaming and whole-file digests are identical for 0, 1, 65535, 65536, 65537, and 200000-byte inputs, covering the buffer boundary in both directions.

Gates: `cargo test -p pecos-build` (55 tests), clippy `-D warnings`, fmt, and `just lint` all clean.

Still open from the same review, deliberately not in this PR:
- **cuTensor is downloaded and extracted with no verification at all**, on the same `build.rs` line as cuQuantum, even though NVIDIA publishes its SHA256 (confirmed: `redistrib_2.4.1.json`). Its own PR — separate dependency, separate table.
- **`build.rs` discards install failures** via `if let Ok(...)`, and installer output goes to cargo's captured stdout, so a mismatch on the auto-install path is invisible without `-vv`. That makes this PR's mandatory check advisory exactly where it matters most, and wants the `cargo:warning=` treatment the cuTensor code already uses.

ciaranra:
Follow-on fix from the cross review of the sibling cuTensor PR (#495), which found the same defect this PR introduced.

The `.part` download name here was deterministic (`archive.part`), shared across every process using the cache — and the comment explicitly claimed "two installs racing on the same cache cannot leave a file that later runs mistake for a verified archive," which was false. Two concurrent installs sharing `~/.pecos` (one per worktree is a normal workflow) could truncate each other's partial between one install's verify and its rename, so the bytes verified would not be the bytes renamed into place.

The partial name now carries `std::process::id()`, and the comment states the actual guarantee. The cached-fail `remove_file` also now tolerates `NotFound` so a concurrent discard doesn't abort this install's refetch.

Gates clean: `cargo test -p pecos-build`, clippy `-D warnings`, fmt, `just lint`.

simbot912:
Great security work here! 🔒 Unverified downloads in build scripts are a serious supply chain risk, especially for scientific software that gets installed in research environments.

A few thoughts on the cuQuantum checksum verification:

1. **NVIDIA's release manifest** — Good catch that they do publish SHA256 checksums. The manifest URL pattern you found (`.../compute/cuquantum/redist/`) should be documented in the README for users who want to manually verify.

2. **Automated verification workflow suggestion**: Consider adding a CI job that periodically fetches NVIDIA's manifest and compares against the hardcoded checksums, alerting on mismatches. This catches both NVIDIA-side changes and local corruption.

3. **For cuTensor (#495)** — Since NVIDIA doesn't publish checksums for cuTensor, you could implement a weaker but still useful integrity check: verify the archive extracts to the expected directory structure and expected binary sizes, or use a hash of the previous successful build as a baseline.

This is the kind of rigor that makes QEC software trustworthy for actual quantum hardware testing. 👏

PR #480, opened 2026-08-11
feat: support Selene 0.3, Guppy v1, and Python 3.12+

qartik:
## Summary

- Upgrade `selene-sim` and all Selene plugin runtime dependencies to `~=0.3.0`.
- Upgrade Guppy to `~=1.0.1`, with compatible HUGR and tket constraints, and retain Guppy v1 semantics throughout generated programs: `output()`, explicit `Measurement.read()`, and `collect_measurements()` for measurement arrays.
- Support Python 3.12+ by dropping Python 3.10 and 3.11. The wheel ABI, cibuildwheel targets, release smoke test, package metadata, classifiers, and consistency guard now agree on 3.12.
- Adapt PECOS Selene plugins, the shared adapter, and runtime loading to the Selene Core 0.3 batched-operation and descriptor ABI.
- Extend HUGR-to-AST compatibility for Guppy v1 function/CFG layouts while explicitly rejecting sequential or nested conditionals until recursive lowering is implemented.
- Migrate experimental Zlup/Zluppy packages to Python 3.12+ and preserve Zlup HUGR `Measure` results as direct booleans.

## Why

Guppy 0.21 constrained `selene-sim<0.3`. Guppy 1.0.1 is the compatible stable release and requires Python 3.12 or newer.

## Known limitations

- The deterministic Z-basis gate-local canonical-frame test is a strict expected failure. Raw gate-local records still demask correctly through `PauliFrameLookup`; only canonical runtime emission diverges. The X-basis regression remains required, and an unexpected pass of the Z case fails CI.
- PECOS currently rejects `output(tag, array(m0, m1, m2))` when its complete runtime result-ID list is not in certified source order. This preserves correctness rather than silently misbinding detector and observable provenance; the root-cause investigation is tracked in #489.
- SZZ traced-QIS DEM construction now fails loudly when final-readout provenance is reversed, rather than producing a potentially misbound DEM. The affected tests are strict expected failures pending the root-cause fix in #498.

## Validation

- Focused signed-zero NumPy-parity regression (xdist): `8 passed`.
- `just python-ci-core debug` ran with xdist after the current `dev` merge. Its two `pecos-rslib` lanes passed: `2,112 passed` and `106 passed`.
- The `quantum-pecos` lane reported `4,989 passed, 71 skipped, 23 xfailed`, plus three inherited platform-sensitive capture failures: the degenerate `d3-x-bp_osd` correction tie, a one-ULP DEM diagnostic aggregate, and bootstrap CI values differing at approximately `4e-16`.
- `RUSTUP_TOOLCHAIN=stable just python-ci-lint`, `cargo fmt --check`, and `uv run --frozen pre-commit run --all-files` passed before the current merge; `cargo fmt --check` also passed after it.
- `uv lock --check` and `uv run --frozen python scripts/check_python_workspace.py` passed before the current merge.


python/quantum-pecos/tests/qec/surface/test_pauli_mask_harvest.py
```diff
@@ -642,7 +642,23 @@ def _assert_canonical_frame_output_matches_lookup(
     assert not observables.any()
 
 
-@pytest.mark.parametrize("basis", ["Z", "X"])
+@pytest.mark.parametrize(
+    "basis",
+    [
+        pytest.param(
+            "Z",
+            marks=pytest.mark.xfail(
```
qartik, line 652: here's the expected failure

ciaranra:
## Review: multi-reviewer panel — verdict **REVISE**

Reviewed at `c1a7bab`, merge base `04d5959b7` (branch is 3 commits behind `dev`).

Three correctness regressions were found that produce **silently wrong results**, all three reproduced by execution rather than inferred from reading. In each case, the check that would have caught the regression was removed or rewritten in this same PR. CI is fully green (84 pass, 7 skipped) — see "Why CI is green" below.

### Methodology

Five independent reviewers, each given only the diff slice and an adversarial brief, all blind to each other and to the orchestrator's own reading:

| Arm | Model | Scope |
| --- | --- | --- |
| Fresh-context ×4 | Claude Opus 5 (Claude Code subagents) | (1) Selene 0.3 FFI/ABI, (2) HUGR engine + fixtures, (3) Python/Guppy v1 migration, (4) deps/CI/packaging |
| Cross-model ×1 | OpenAI Codex CLI 0.145.0, read-only sandbox | Selene FFI + Guppy measurement semantics |
| Fusion + verification | Claude Opus 5 | Cross-checked every headline finding against the code or by running it |

Two notes on weighting, for transparency. First, the branch name indicates Codex assisted in authoring this change, so the Codex arm was treated as supplementary corroboration rather than as the independent cross-model check — models tend to self-endorse their own output. Second, every finding below was re-verified by the fusion step against the actual code or by execution; findings that did not survive that check were dropped, and one arm's severity rating was lowered (noted inline).

---

## Blocking

### 1. Array-valued result tags bind measurement IDs in reverse order

The `tagged_readout` program that this PR rewrote in `docs/user-guide/inferred-guppy-dem.md` was, at the merge base, a CI-tested snippet asserting:

```python
assert inferred.detector_supports   == ((0, 1),)
assert inferred.observable_supports == ((0,), (1,), (2,))
```

Running the same program on this branch (only `.read()` added):

```
detector_supports   = ((2, 1),)
observable_supports = ((2,), (1,), (0,))
raw_binding         = runtime_result_ids
```

`logical_z = m0` now binds to physical measurement 2. Every detector and observable built from an `array(...)`-valued tag attaches to the wrong physical measurement.

Two details make this worse than a plain ordering bug:

- PECOS reports `raw_binding == "runtime_result_ids"` — its *highest*-confidence provenance mode, meaning "every raw output retained its QIS measurement ID directly" — while the IDs are reversed.
- The sibling `reordered_readout` example in the same document still passes, because it resolves through the `probe_correlated_result_ids` path instead. So the failure is specific to the direct-binding path, and the surviving doc test does not cover it.

`collect_measurements(measure_array(...))` is **not** affected (`result_ids = [0,1,2,3]`, verified), so the surface generators are safe. The corruption is specific to `array(...)`-constructed tags, which remain a documented public form.

The root cause was not localized to a line; it is most likely in the changed array-output handling in `crates/pecos-hugr`, not in `result_tags.rs`.

### 2. The guard that would have caught #1 was deleted (`qec/dem.py`)

`python/quantum-pecos/src/pecos/qec/dem.py:867-914` at the merge base cross-checked the generator-certified positional layout against the runtime trace's own scalar result IDs, raising:

> generator layout binds {tag} occurrence {n} to source measurement {x}, but the runtime trace reports result id {y}; abstract and source measurement order have diverged

That is precisely the failure mode in #1. The 47 lines were removed; what remains (`dem.py:844-866`) validates per-tag cardinality and slot uniqueness only, which no permutation can trip.

In the same PR, `docs/user-guide/inferred-guppy-dem.md:251-262` was rewritten from `result("physical", array(m0, m1, m2))` to seven repeated scalar `result()` calls. The array form still compiles fine under guppylang 1.0.1, so this was not a language-forced edit. After this PR, no test and no doc example exercises array-valued raw or observable tags.

Suggested resolution: restore the guard, narrowed to the slots that still expose scalar provenance, rather than removing it.

Also note the surrounding prose at `inferred-guppy-dem.md:41` and `:240` still states that an array-valued observable "expands into one DEM observable per element" and "contributes its elements in array order", which is currently false.

### 3. `hugr_to_ast.py` drops trailing measurements and reorders gates across control flow

Both behaviours reproduced on this branch:

```
source: q=qubit(); h(q); r=measure(q).read(); q2=qubit(); if r: x(q2); return measure(q2).read()
result: [Prepare, H, Measure, Prepare, IfStmt]        <- final Measure absent; register m1 never declared

source: q=qubit(); h(q); while count<3: x(q) ...; z(q); return measure(q).read()
result: [Prepare, H, Z, WhileStmt]                    <- Z hoisted BEFORE the loop; Measure absent
```

Both originate in lines added by this PR:

- `python/quantum-pecos/src/pecos/circuit_converters/hugr_to_ast.py:366-370` — the `MEASURE_OPERATIONS` exclusion, justified by the comment "Measurements directly below the function are its return value and are not SLR statements." A measurement is a circuit operation regardless of whether its value is returned; excluding it drops the readout and its classical register from the converted circuit.
- `hugr_to_ast.py:355-359` — `operation_parent_idx = min(cfg.blocks)`. The comment justifies this for *setup* operations preceding the CFG, but it is applied to every `FuncDefn`-parented operation, including those following the CFG. For non-commuting gates this changes the circuit's unitary.

Downstream impact: `guppy_to_ast(...)` → `generate(ast, "qasm"|"stim")` emits a circuit missing its readout, and any AST-derived DEM, Stim export, or resource count is wrong.

`python/quantum-pecos/tests/selene/test_hugr_to_ast.py` passes 17/17 because `test_conditional_circuit` and `test_simple_loop` assert qubit capacity and that exactly one `IfStmt`/`WhileStmt` exists — never statement order, register declarations, or the trailing measurement. The PR's diff to that file is 20 mechanical `.read()` renames.

---

## Why CI is green

All 84 checks pass. That is consistent with the above rather than evidence against it:

- #1 is uncovered because the doc example exercising it was rewritten to scalars and the `dem.py` guard was deleted, both in this PR.
- #3 is uncovered because the converter's tests assert structure counts, not order or readout.

The generated doc tests, the `qec` suites, and `test_hugr_to_ast.py` all pass on the branch. The findings above come from executing the affected programs and comparing against the merge base's own previously-passing assertions.

---

## Should fix before merge

### Selene 0.3 FFI

Corroborated independently by the Codex arm and the fresh-context FFI arm.

- **No `api_version` or `struct_size` validation** (`crates/pecos-qis/src/selene_runtime.rs:321-341`). Upstream's own loader (`selene-core` `runtime/plugin.rs:211-226`) reads `struct_size` first, rejects an undersized descriptor before touching any other field, then validates `api_version`. PECOS null-checks and immediately materializes `&*descriptor`. A short or partially zero-initialized descriptor is undefined behaviour rather than a clean version error, and ~22 non-`Option` function-pointer fields are unchecked. (`exit_fn` is correctly treated as `Option`.)
- **Only one of the two descriptor entry points is supported.** Upstream accepts both the data symbol `selene_runtime_plugin_descriptor_v1` (tried first) and the accessor `selene_runtime_get_plugin_descriptor_v1`. PECOS accepts only the accessor, so a conforming C runtime plugin exporting the data symbol is rejected.
- **The hand-mirrored `#[repr(C)]` struct and `std::mem::transmute` should be removed** (`selene_runtime.rs:71-89`, `:1096-1110`). Both arms confirmed the layout is *correct today* — verified byte-for-byte (interface 72/8, handle 80/8, `interface` at offset 8) and slot-by-slot by invoking all nine callbacks through a real upstream handle. The problem is forward robustness: upstream marks `RuntimeGetOperationInterface` `#[non_exhaustive]`, so an appended field silently produces a short by-value aggregate with no compile error. `selene-core` is now a direct dependency and ships `BatchBuilder` for exactly this purpose (`operation/plugin.rs:22`), which is what upstream's own `RuntimePluginInterface::get_next_operations` uses. Adopting it deletes the mirror, the transmute, and the nine `runtime_batch_*` callbacks. One behavioural nuance to preserve: PECOS terminates on `!batch.invoked`, upstream on `!batch.is_empty()`; these differ only for a batch that calls `set_batch_time` alone.
- **`RPPGate` is silently discarded** (`selene_runtime.rs:189-199` pushes `RuntimeScheduledOp::Custom`; `:1554` drops `Custom` with an empty arm). A defined Pauli-product rotation vanishes, `mark_gate_end` never fires for those qubits, and no idle is inserted — the simulator proceeds on a wrong state with no diagnostic. The inconsistency is self-evidencing: the simulator side rejects RPP loudly in `pecos-selene-core/src/lib.rs:167-169` and in all five plugins. Not reachable with the bundled runtimes today, since PECOS never calls `rpp_gate_fn`; a third-party runtime synthesizing RPP makes it live. (The `Custom` drop itself is pre-existing and defensible — custom-op semantics are opaque by definition. Only the RPP mapping is new.)
- **`SeleneRuntime::new()` and `selene_runtime_plugin(library_file)` now fail against the bundled runtimes.** Both default to empty init args (`selene_runtime.rs:344`, `python/pecos-rslib/src/engine_builders.rs:167-185`), and Selene 0.3 makes the seven duration flags required with no defaults. Confirmed by loading the built runtimes directly: init with no args returns errno -1, with the clap diagnostic going to the plugin's stderr. `selene_runtime_auto` falls to `_ => &[]` for any third-party runtime. Fail-loud is right; the regression and the opaque error message are not.

### Tests that cannot fail

- `crates/pecos-qis/src/selene_runtimes.rs:418-432` — asserts the constant's own shape back at itself (length 7, each element starts with `--duration-ns-` and ends `=0`). `--duration-ns-rxyz=0` satisfies every assertion and then fails at init. The five soft-RZ duration flags are never checked, only `.last()`. Nothing invokes the plugin.
- `crates/pecos-hugr/src/result_tags.rs:216-229` — `indirect_call_is_nontrivial_control_flow` no longer loads `indirect.hugr`; it constructs an `OpType::CallIndirect` and asserts the match arm recognizes it, which cannot fail while the arm exists. The doc comment above it still claims it proves the arm is load-bearing "for this fixture", and the `INDIRECT` `include_bytes!` was removed. Cause: guppylang 1.0.1 inlines the helper and DCEs the indirect call, so the fixture no longer contains `core.call_indirect` and the real assertion would now fail.
- `crates/pecos-hugr/src/engine.rs` — four call tests relaxed in the same commit: `call_targets.len() >= 2` → `quantum_ops.len() >= 6`; `func_defns.len() >= 2` → `quantum_ops.len() >= 3`; a `num_inputs >= 2` check → `gate_type == CX`; and `test_ry_angle_tuple` lost both runtime-chain assertions that existed specifically to distinguish the runtime path from static extraction. Separately, `test_ch_gate_full_execution_completes_cleanly` retains its comment about the Call-argument-ordering stall but now guards nothing, since the regenerated `ch_gate.hugr` contains zero `core.call`.
- `python/quantum-pecos/tests/guppy/test_rotation_extension.py:41-43` — `rz_calls >= 2` → `>= 1` with the comment "Guppy 1's optimizer *may* fuse consecutive Z rotations"; the actual count was not determined. The body still emits two distinct rotations. Line 51 also replaced the deliberate `rz(q, pi * 2.0)` full-rotation edge case with `rz(q, pi / 2)`.

Related: the new plain-dataflow call-frame path (`engine.rs:1196-1260`, `control_flow/call.rs:43-104`, ~90 lines replacing a previously loud error) has no Rust-level fixture coverage — `core.call` count is 0 across all 23 regenerated `.hugr` fixtures.

### Other correctness

- **`tket.qsystem.sol` is never dispatched** (`crates/pecos-hugr/src/engine/handlers.rs:144`). This PR added `"tket.qsystem.helios"` and omitted Sol, whose extension id is `"tket.qsystem.sol"` (tket-qsystem 0.27.2, `sol.rs:34`). PECOS accepts `QSystemPlatform::Sol` as first-class (`pecos-hugr-qis/src/compiler.rs:436`). A Sol-lowered HUGR defers every op and ends as a starved-deferred-nodes stall naming no platform. One-word fix.
- **PHIR parser selects the first `FuncDefn`, not the entrypoint** (`crates/pecos-phir/src/hugr_parser.rs:304-309`). The new unconditional `return Some(node)` means the scan can never reach a later function, so a module with more than one `define-func` emits a helper's body as the whole program. `hugr.entrypoint()` is what the engine's own new code uses.
- **Order-port off-by-one at `engine.rs:2090-2093`.** The `_ =>` arm uses `hugr.num_inputs(output)` raw, which counts the `StateOrder` port. This is the same defect the PR correctly fixes, with explanatory comments, at `cfg.rs:1080` and `conditional.rs:376`. A non-`FuncDefn` entrypoint returning one value gets a spurious trailing `None`, so results are keyed `return_0` instead of `return`.
- **Unbounded recursion replaces the stall diagnostic** (`engine.rs:1495-1503`): `if !self.work_queue.is_empty() { return self.process_hugr_impl(); }` with no depth bound and no guaranteed TCO in debug builds. A `loop` with a progress flag is the same logic without the stack growth.
- **The known-limitation configuration is unguarded.** The PR body's claim is accurate and was verified: the xfail is `strict=True`, `[X]` passes, and both raw gate-local demask tests pass, so "raw gate-local records still demask correctly" is backed by live assertions. Under `--runxfail` the Z failure is a spurious detection event at detector index 2 in 3 of 4 shots — systematic, not flaky. However `TwirlConfig(site_schedule="before_two_qubit_gate", frame_output="canonical")` is still accepted by `_twirl_config.py:91-120` and still generates a program, so a user selecting it silently gets corrupted canonical syndrome records. `validate_runtime_supported`'s docstring states its purpose is to make unsupported configurations fail loudly at this boundary. Either fix the frame propagation or reject the combination.
- **`examples/simple_run_guppy_api.py:220`** was not migrated (`return measure(qubit())` as `bool`) and raises `GuppyTypeError` under 1.0.1. It is masked by `except Exception` at `:226-227`, so the demo prints a success line while testing nothing. No CI job executes `examples/**`. Five notebooks remain on pre-1.0 Guppy, including `examples/surface_code_selene_demo.ipynb`, the notebook twin of a `.py` example this PR did migrate.
- **New skip aborts a sweep** (`python/quantum-pecos/tests/guppy/test_hugr_compiler_parity.py:225-227`). `pytest.skip()` raises, so the first fixture matching `"Conflicting signature"` skips every remaining fixture in the loop, not just the offending one. It is also keyed on failure text rather than a property of the fixture. An allow-list plus `continue` would be closer to the intent.

<details>
<summary><b>Lower severity (click to expand)</b></summary>

- **Two of the seven new workspace clippy allows are unknown lint names.** Reproduced on stable 1.97.1: `clippy::assert_is_empty` is unknown on both stable and nightly (clippy suggests `const_is_empty`), and `clippy::chunks_exact_to_as_chunks` exists only on nightly 1.99. Both emit `unknown lint` warnings. This does not fail CI (`unknown_lints` ignores `-D warnings`), but it indicates the allow-list was validated against nightly rather than the stable toolchain CI runs.
- **The same commit both fixes and suppresses the same lints.** `975b701d1` adds allows for `chunks_exact_to_as_chunks`, `manual_slice_fill` and `drain_collect` while simultaneously rewriting those exact call sites in `pecos-random` (9 sites), `pecos-core/src/bitset.rs:207`, and `pecos-quantum/src/tick_circuit.rs:2553`. The allows are what persist, and ~89 remaining `chunks_exact` sites will never be flagged again.
- **The blanket `#[allow(clippy::unnecessary_wraps, clippy::unused_self)]` on all five plugins is self-inflicted.** The 0.3 trait drops only `rxy`/`rz`/`rzz`/`measure`/`reset`; `shot_start`, `shot_end`, `postselect`, `get_metric`, `dump_state` and `exit` all still exist on `SimulatorInterface` and did not need to move to inherent impls. Clippy does not fire these lints inside trait impls. A `--force-warn` run confirms every suppressed warning comes from exactly the six methods that did not need moving; zero are needed for the five that did. Leaving them in the trait impl removes the need for both allows.
- **MSRV.** `as_chunks_mut` was stabilized in Rust 1.88; no `rust-version` is declared in any `Cargo.toml`. A downstream consumer on older stable gets a raw `E0658` from inside a dependency rather than a resolver-level error.
- **`selene-core` is a non-optional dependency of `pecos-qis`** (`crates/pecos-qis/Cargo.toml:70-73`) although its only use is behind `#[cfg(feature = "selene")]`. `cargo build -p pecos-qis --no-default-features` now requires network access to the Quantinuum git remote.
- **Descriptor re-resolved on every FFI operation** (~15 call sites). Each gate, measurement and allocation performs a fresh `dlsym` plus accessor call. Caching it in `load_plugin` alongside `instance` would be straightforward. Separately, `reset_plugin_instance` (`:519`) now propagates `?` before clearing `self.instance`/`self.library`, inverting the previous cleanup invariant.
- **Comments that now contradict the code**: `selene_runtime.rs:2115-2117` describes `shot_end(instance, shot_id, seed)` (upstream is `shot_end_fn(handle)`); `:303-306` says the `(shot_id, seed)` pair must be retained until shot end, but it is now destructured to `(_shot_id, _seed)`; `:2007-2009` says "an ABSENT symbol stays legal" for the result setters, which are now called unconditionally; `:955-957` and `:1823-1833` describe a barrier fallback that is now unreachable, leaving `call_runtime_global_barrier` unable to return anything but `Ok(true)` or `Err`. Also `executor.rs:1968-1973` justifies not linking the helios archive by a symbol collision that no longer exists in 0.3.
- **Five near-identical `handle_operations` copies** while `pecos-selene-core::SeleneAdapter` implements the same loop and is referenced by nothing but its own test. The bodies were diffed mechanically and are byte-identical apart from error strings, so there is no copy-paste bug — but the PR updates the adapter and adds five more copies of its logic in one changeset.
- **Python floor stragglers.** `python/quantum-pecos/.readthedocs.yaml:11` still builds on 3.11 and installs a package now requiring `>=3.12`, so the first docs build after merge fails. `docs/user-guide/cuda-setup.md:331` still says `conda create -n pecos-cuda python=3.11`, in the same document whose line 68 this PR updated to "3.12, 3.13, or 3.14". The `python-version-consistency` job cannot catch either — `scripts/check_python_workspace.py` validates package versions and uv workspace metadata, never `requires-python` or classifiers.
- **`ruff.toml` remains at `target-version = "py310"`** while black moved to `py312` and `requires-python` to `>=3.12` — three baselines in one PR. This currently hides four `UP036` outdated-version blocks (the 3.10 `Self` shims in the cuda simulator modules and `qgate_base.py`), and the cross-reference comment in `pyproject.toml` that used to keep black and ruff pinned together was deleted.
- **`osv-scanner.toml:134`** suppresses RUSTSEC-2026-0197 with the reason "no swap_columns call exists in PECOS or in the pinned hugr-core 0.27.1/tket 0.19.0 sources… **Re-check on hugr/tket bumps**." This PR bumps those to 0.29.3 / 0.21.2 and does not touch the file. `cgmath 0.18.0` is still in `Cargo.lock`, so the suppression remains load-bearing but its recorded evidence is stale.
- **Plugin `pyproject.toml` still declare `selene-core>=0.2`** in all five plugins while the code targets the 0.3 interface and the test extra pins `selene-sim~=0.3.0`. In practice pip resolves to the newest, so this needs a co-installed cap to actually break — noted as looseness rather than a live failure. (One arm rated this high; lowered here after checking resolution behaviour.)
- **Orphaned artifacts**: `crates/pecos-hugr-qis/tests/fixtures/szz_barrier_{collision,wrong_signature,single_wrong_abi}.hugr` now have zero references after their three tests were deleted and only partially replaced; `crates/pecos-hugr/tests/fixtures/indirect.hugr` is still generated but consumed by nothing; `docs/assets/test-data/repetition_code.hugr` and `crates/pecos/tests/test_data/hugr/bell_state.ll` were not regenerated despite their generator and the LLVM lowering changing.
- **Deprecation debt**: guppylang 1.0.1 defines `result` as a deprecated alias for `output`. PECOS codegen now emits `output(...)`, but every migrated test, doc snippet and example still imports and calls `result`. A second migration is already owed.
- **`decode.py:1025-1027`** widens the values/result_ids mismatch suppression to `init_synx`, `init_synz` and any `final:meas:*` tag. The widening is factually necessary (scalar output tags now return empty `result_ids`), but it converts a fail-loud provenance check into a silent skip for a growing tag set, and the justifying comment is inaccurate for `final:meas:*`, which is bound via the `final` array rather than by scalar tags.

</details>

---

## Verified correct

Stated for balance — these were specifically attacked and held up:

- **The version cohort is genuinely compatible.** `Cargo.lock` resolves hugr/hugr-core/hugr-model/hugr-llvm 0.29.3, tket 0.21.2, tket-qsystem 0.27.2; `uv.lock` resolves hugr-py 0.18.3, guppylang 1.0.1, tket-py 0.15.6. hugr-py 0.18.3 is built from the same monorepo tag as hugr-core 0.29.3. Confirmed against upstream changelogs — this is one coordinated release train, and the Rust side can read what the Python side emits.
- **`.read()` placement is correct at every migrated call site**, including inside loops and conditionals in `guppy_gen/{surface,color,transversal}.py` and the `surface_code_d{3,5,7}.py` examples. No read was moved earlier or later than the original measurement point.
- **`collect_measurements()` is index-preserving** (`array(m.read() for m in measurements)`), verified empirically, so the surface generators' `measure_array` rewrites are semantics-preserving and downstream detector/observable indexing is not permuted.
- **Per-shot reset is complete.** Every field of `HugrEngine`, `WireState`, `MeasurementState` and `ExtensionState` was enumerated against the reset paths; the new `outcomes` map is cleared, and `test_measurement_state_reset_clears_outcomes` passes.
- **The lazy-measurement indexing is sound** — `measurement_index = mappings.len()` and `global_idx = processed_count + local_idx` index the same vector; read-late, read-twice and never-read all behave correctly, and `FutureToMeasurement` forwards the index rather than the qubit across Call/CFG/TailLoop/Conditional boundaries.
- **All 23 `.hugr` fixtures regenerate byte-for-byte** from the committed scripts under guppylang 1.0.1 / hugr-py 0.18.4. The 6–10× size growth is encoding (zstd off plus embedded extension-registry JSON), not semantics: `looped.hugr` still has a real back-edge, the conditionals still have two successors joining at an exit, `forloop_h_test` still has exactly one `H`.
- **The Selene ABI mirror is correct as of this rev** — proven by layout comparison and by invoking all nine callback slots through a genuine upstream handle, plus an end-to-end dlopen of both bundled runtimes (init → shot_start → qalloc → gates → measure → barrier → set_bool_result → shot_end → exit, all errno 0).
- **`MeasureLeaked` uses the correct encoding** (0/1 ordinary, 2 leaked); emitting 0/1 from non-leakage-aware simulators matches upstream's own reference simulators.
- **The forced `--duration-ns-*=0` and `--max-batch-size=1` args reproduce 0.2 timing exactly** and match the plugins' clap field sets — behaviour-preserving, though the rationale is not recorded anywhere in the code.
- **`shot_end` losing `shot_id`/`seed` is correct** per upstream 0.3; only the leftover state and comments are stale.
- **The kebab→snake clippy key rename is behaviour-neutral and correct** — the old hyphenated keys did work, and the rename silences a Cargo deprecation warning about hyphenated lint names.
- **No Python test assertion was deleted or gutted** — assert/`pytest.raises` counts are unchanged across every touched Python test file, and no `try/except`, `getattr` default, `strict=False` or `approx` was introduced. (The relaxations listed above are in Rust tests and in the two Python cases named explicitly.)
- **Doc snippets remain executable and covered**; the generated doc tests pass with no new skip markers.

---

## Suggested order of work

1. Fix #1 (array result-ID reversal) and restore the `dem.py` guard in narrowed form; re-add an array-valued doc example or test so the path is covered again.
2. Fix #3 (measurement drop and gate hoist in `hugr_to_ast.py`), and extend `test_hugr_to_ast.py` to assert statement order, register declarations, and trailing readout.
3. Selene FFI: add descriptor validation, accept both entry-point forms, and replace the mirror/transmute with `BatchBuilder`; make RPP an explicit error rather than a silent drop; restore working defaults for `SeleneRuntime::new()`.
4. Restore the tests listed under "Tests that cannot fail", or replace them with assertions that can.
5. The remaining items are independently landable.

Happy to supply the reproduction scripts for #1 and #3 if useful.

qartik:
@ciaranra anymore feedback?

ciaranra:
## Re-review at `6eebab3` — verdict **REVISE**

Re-reviewed the three fix commits (`c1a7babd..6eebab3`) with a second panel scoped to that range only. Much of the previous round is genuinely fixed. However, the `hugr_to_ast.py` fix introduced **two new critical defects**, and the restored `dem.py` guard does not cover the case it was written for. Net: the blocker count went up, not down.

CI is green (86 pass, 7 skipped), and the full `tests/qec` suite passes (1491 passed, exit 0). As last round, that is consistent with the findings rather than evidence against them.

### Methodology

Four independent reviewers scoped to the fix commits, blind to each other and to my own reading: three fresh-context Claude Opus 5 subagents (one per area: the `hugr_to_ast` rewrite, the Selene descriptor hardening, the `dem.py` guard plus CI/packaging) and one OpenAI Codex CLI 0.145.0 read-only arm across the whole range. Fusion and verification by Claude Opus 5. Every finding below was independently re-verified against the code or by execution before inclusion; findings that did not survive were dropped, and one arm's severity was lowered (noted inline).

---

## Fixed — verified

- **The two original `hugr_to_ast` symptoms** are fixed for the simple cases. Trailing measurements are retained and post-CFG operations stay after the CFG. The new assertions in `test_hugr_to_ast.py` are real (`ast.body[-1]` is a `MeasureOp`, `ast.body[-2].gate == GateKind.Z`, two `RegisterDecl`) and would catch a straight revert.
- **RPP is no longer silently dropped** — now a real `RuntimeScheduledOp::Rpp` carrying all four parameters, erroring at the conversion site, with `test_runtime_batch_rpp_fails_loudly` covering it.
- **Descriptor loading is properly hardened.** `struct_size` is read from the header and checked *strictly before* any full-struct read (verified byte-for-byte against upstream's `copy_plugin_descriptor`), `api_version` is validated with semantics identical to upstream's, both entry-point symbols are supported, and `&*descriptor` is replaced by a by-value copy. Accepting `struct_size > expected` is correct for an append-only ABI. The reviewer that attacked this diff hardest rated it a net improvement with no new defects, and credited an uncredited one: the new `api_version` gate materially reduces the `#[non_exhaustive]` mirror-struct risk raised last round, because upstream requires a minor bump for an ABI-breaking change.
- **`tket.qsystem.sol` dispatched; PHIR parser uses `hugr.entrypoint()`; the unbounded recursion is now a `loop`; the order-port off-by-one uses `dataflow_signature().input_count()`.**
- **The dropped CI jobs were genuinely obsolete, and removing them was mandatory rather than tidy.** The 8 deleted `test_abi3_wheels` entries were 3.10/3.11 crossed with 4 platforms; those wheels now declare `Requires-Python: >=3.12`, so the jobs would have hard-failed after the floor bump. No platform, wheel-install smoke test, or LLVM build lost coverage; all four platforms keep 3.12/3.13/3.14. In `python-test.yml` no job was deleted at all — three `python-version` values moved 3.10 → 3.12. This closes the concern I raised when that commit landed.
- **Packaging is self-consistent**: all 10 workspace members at `>= 3.12` with matching classifiers; `uv lock --check`, `scripts/check_python_workspace.py` and `ruff check` all exit 0. `ruff.toml` at `py312` surfaced no new unaddressed diagnostics.
- The in-loop `pytest.skip` in the parity sweep, the broken `simple_run_guppy_api.py` example, `.readthedocs.yaml`, `cuda-setup.md`, `osv-scanner.toml`, and the plugin `selene-core` pins are all addressed.

---

## New — introduced by the fix commits

### N1. Postlude operations get a guessed qubit target (critical)

`hugr_to_ast.py:1102-1103` (gates) and `:1115-1116` (measurements) added a fallback: when a postlude operation's qubit operand fails to resolve, it takes `max(self.qubit_allocations.values())` — the highest allocated index. Instrumentation shows the fallback fires in **every** postlude case; it is correct only by coincidence.

```python
q0 = qubit(); q1 = qubit()
h(q0)
r = measure(q1).read()
if r:
    x(q0)
return measure(q0).read()
```

```
actual: [Prepare, Prepare, MeasureOp[1], H[0], IfStmt, MeasureOp[1]]
                                                        ^ measure(q0) emitted as qubit 1
```

Qubit 0 is never measured; qubit 1 is measured twice. A trailing `z(q0)` likewise emits `Z [1]`.

This is a regression *from* the previous state and a worse class of bug: before, these operations were visibly missing, which a reviewer can see. Now they are present with wrong operands, which nobody sees.

### N2. The same fallback emits structurally invalid two-qubit gates (critical)

The fallback returns a **one-element** list regardless of gate arity.

```python
cx(q1, q2)   # after a conditional
```

```
actual: CX[2]        expected: CX[1, 2]
```

Nothing validates arity downstream, and both `generate(ast, "stim")` and `generate(ast, "qasm")` then **silently omit the CX** — the entangling gate disappears from the emitted circuit with no error at any layer.

### N3. Two trailing measurements can be bound to each other's qubits (critical)

`_capture_cfg_output_qubits` (`hugr_to_ast.py:1014-1024`) binds CFG output ports by node without distinguishing source port, so both ports can bind to the same qubit. The `if len(qubit_indices) == 1` guard at `:1022-1023` detects *disagreement* between branches, not *wrongness* — unanimous-but-wrong passes, and genuine disagreement produces no binding and hands the operation to the N1 guess. For a function returning `(measure(q1), measure(q2))`, the emitted measurements are `M[2]` then `M[1]`.

**Why the new tests miss all three:** `test_conditional_circuit` asserts the trailing measurement exists and its position, but never its target index — and its program measures the *highest* qubit, so the guess is right by luck. `test_simple_loop` has exactly one qubit, so the guess cannot be wrong. Minimum additions: assert `targets[0].index` on every postlude operation, add a two-qubit variant whose trailing operation targets the *lower* index, and add a trailing two-qubit gate asserting `len(targets) == 2`.

---

## The restored `dem.py` guard does not cover the array case

### N4. The guard is blind to exactly the case its comment claims to catch (high)

The comment at `dem.py:874-876` states the check "catches a reversed array provenance record". It cannot, for a structural reason.

The generator records array-valued measurements as synthetic per-element tags — `layout.append((f"{tag}:meas:{element}", 0))` at `guppy_gen/surface.py:2886` — with slot index `0`. At runtime, a real `make_surface_code(3, 2, "Z")` build produces:

```
final          n_values=9  ids=[20,21,...,28]   <- the array, ids PRESENT
final:meas:0   n_values=1  ids=[]               <- the slots the layout references, ids ABSENT
...
runtime tags NOT referenced by layout: ['final', 'init_synx', 'synx', 'synz']
```

The guard correlates strictly by tag name. The nine `final:meas:N` lookups are skipped by the length-mismatch check at `dem.py:887`, and the honest evidence — the `final` array's nine ids — is never consulted, because no layout entry carries the tag `final`. Re-invoking the unmodified function with the real circuit, the real traces, and a reversed `final:meas:*` layout returns **accepted**, binding every data-qubit measurement to the inverted id.

Net effect: the guard covers the 20 scalar slots that were never the concern, and gives zero coverage to the 9 array-lowered slots that were.

`qec/surface/decode.py:994` already performs the needed mapping (`array_trace_ids[name][0][element]`), so the correspondence exists elsewhere in the codebase and can be reused.

### N5. The original reversal is still live on the inference path (blocking, unchanged)

Reproduced unchanged on `6eebab3` — byte-identical to the previous round:

```
detector_supports   = ((2, 1),)      expected ((0, 1),)
observable_supports = ((2,), (1,), (0,))   expected ((0,), (1,), (2,))
raw_binding         = runtime_result_ids
```

The reason the fix does not reach it: `_generator_certified_result_traces` is called only from `_compiler_certified_result_traces` (`dem.py:734`), gated on `generator_layout is not None`, and reached only from `GuppyDemBuilder.build`. `infer_guppy_dem_annotations` has its own provenance block in `guppy_output_dem.py:302-348`, which the fix commits do not touch. There, runtime ids *are* the binding (`raw_ids.extend(result_ids)` at `:315`, `raw_binding = "runtime_result_ids"` at `:348`) and the only check is **set** equality at `:319` — never order. A reversed-but-complete list is accepted at face value; correlation runs only when ids are *missing* (`:324`). The same pattern is unguarded at `decode.py:994`, where the dense-range check at `:1000` passes any permutation.

So the reversal is unguarded in three sites; the fix hardened one of them, and per N4 hardened it where the array case cannot reach.

The new test `test_direct_array_outputs_preserve_runtime_result_id_order` uses `collect_measurements(measure_array(...))`, which is index-preserving and was never affected. It pins a path that already worked, not the broken shape (`array(m0, m1, m2)` over separately measured qubits).

---

## Pre-existing, but newly disguised

Two reviewers found these independently; I reproduced both. They predate this PR, but the rewrite now appends a plausible trailing measurement, so the broken output *looks* complete:

- **A second sequential conditional is silently dropped.** For `if r0: X; Z; if r1: Y; H`, the AST contains the first `If`, `Z` and `H`, but no `Y` and no second `IfStmt`. `:776-778` emits the continuation block's operations flat and never recurses, so if the continuation is itself a conditional header, everything past it vanishes.
- **Loop-plus-conditional shapes lose nearly everything.** A conditional inside a loop yields `[WhileStmt, PrepareOp, GateOp/H, MeasureOp]` — empty loop body, allocation emitted *after* the loop. Conditional detection is skipped whenever `cfg.loops` is non-empty (`:401-402`), and `_build_loop_statements` visits only entry/header/body/exit. Also: loop bodies drop their gates as soon as more than one qubit is present, which is why the passing `test_simple_loop` (one qubit) does not see it.
- The module's own docstring example at `hugr_to_ast.py:36-45` (Bell state) emits both `measure(q0)` and `measure(q1)` as `M [1]`.

Given the breadth here, the honest options are to make CFG traversal port-aware and recursive, or to **fail loudly** on the shapes that are not supported. Emitting a confident, wrong circuit is the worst of the three.

---

## Carried over, unchanged

- The hand-written `#[repr(C)]` mirror and `std::mem::transmute` remain (`selene_runtime.rs:78`, `:1131`). Layout re-verified exact at this rev, and the new `api_version` check reduces the risk — but `BatchBuilder` was confirmed present and sufficient (`runtime_get_operation()` hands back the real handle, so `#[non_exhaustive]` is not an obstacle), and would delete both. Note one migration detail: PECOS terminates its drain loop on `batch.invoked`, which `set_batch_time` also sets, so a plugin emitting only timing spins forever; `BatchBuilder`'s `is_empty()` fixes that latent hang.
- The descriptor is still re-resolved on **every** gate, measurement and allocation — 15 call sites, each a `dlsym` plus (now) a 224-byte by-value copy. Upstream resolves once at load. `load_plugin` already has the descriptor in hand.
- `api_version` is validated *after* the full-struct copy; it sits at a fixed header offset and could be read before it, like `struct_size`.
- `Operation::Custom` still discards tag/data/len (`:180`) and is dropped at `:1598` — the same shape as the RPP bug, one match arm below it. I rate this lower than the reviewer who raised it: it is pre-existing, and unlike RPP a custom operation has no defined semantics for PECOS to get wrong. Worth an explicit error or a comment explaining the asymmetry.
- Two unknown clippy lint names still allowed (`assert_is_empty` does not exist on any toolchain; `chunks_exact_to_as_chunks` is nightly-only), still no `rust-version` despite the 1.88 floor from `as_chunks_mut`.
- Weakened tests not restored: the `result_tags.rs` tautology, the vacuous `selene_runtimes` argument test, the two `engine.rs` call assertions, and `rz_calls >= 1`.
- Minor stragglers: `exp/zlup/pyproject.toml:5` still `>=3.10`; the `tomli` fallbacks in `scripts/check_python_workspace.py:21-25` and `test_selene_plugin_workspace.py:10` are unreachable under a 3.12 floor; all five `Cargo.toml` still use `pyo3/abi3-py310` while advertising `Requires-Python: >=3.12`; and `python-version-consistency` still checks package versions rather than `requires-python`, so the alignment this PR did by hand has no automated drift guard.

---

## Suggested order

1. Remove the `max(...)` qubit guess and make the carried-wire mappings port-aware (N1-N3). If an operand cannot be resolved, raise — do not guess. Add the target-index assertions described above.
2. Decide the CFG scope: either recursive traversal for sequential/nested conditionals and loop-plus-conditional, or an explicit hard error for unsupported shapes.
3. Extend the `dem.py` guard to resolve `f"{base}:meas:{element}"` against the `base` array trace (N4), reusing the mapping at `decode.py:994`.
4. Apply the same provenance check to `guppy_output_dem.py` and `decode.py`, so a complete-but-permuted id list is cross-checked rather than trusted (N5), and add a test using `array(m0, m1, m2)` specifically.
5. The carried-over items are independently landable.

Reproduction scripts for N1, N2 and N5 available on request.

ciaranra:
## Re-review at `310c2e0` — one silent-wrong case left, plus four durability notes

Verified by re-running the earlier reproductions against `310c2e0` (fresh scripts, not reused ones). Most of the previous round is genuinely resolved, and two of the fixes are the right *shape* of fix rather than symptom patches — credit for that below. CI is green (86 pass, 7 skipped); the affected suites give 334 passed, 1 xfailed.

### Fixed — confirmed by execution

- **Guessed qubit targets are gone.** `measure(q0)` after a conditional now emits `MeasureOp[0]` (was `MeasureOp[1]`), and a trailing `cx(q1, q2)` emits `CX[1, 2]` with correct arity (was `CX[2]`, which stim/qasm silently dropped). The `max(qubit_allocations.values())` fallback is removed.
- **The array result-ID reversal no longer mis-binds.** It now raises: `result tag 'physical' must expose physical measurements in certified source order; tag ids=[2, 1, 0], source ids=[0, 1, 2]`.
- **Loop + conditional now fails loudly** with `UnsupportedHugrStructureError: HUGR CFG combines a loop with a conditional; recursive lowering is not supported`, instead of emitting an empty `WhileStmt` with the body discarded.
- **The guard's array blind spot is addressed** — `f"{base}:meas:{element}"` slots now resolve against the base array trace.

---

## Remaining: two sequential conditionals silently drop the second

```python
q = qubit()
r0 = measure(qubit()).read()
if r0:
    x(q)
z(q)
r1 = measure(qubit()).read()
if r1:
    y(q)
h(q)
measure(q).read()
```

```
actual: [Prepare, Prepare, MeasureOp[1], IfStmt, Z[0], Prepare, MeasureOp[2], H[0], MeasureOp[0]]
```

One `IfStmt` where the source has two, and `y(q)` is absent entirely. No error, no rejection. Its siblings were each either fixed or made to fail loudly; this one still emits a confident, incomplete circuit.

### The underlying issue: the converter now has three contracts

`hugr_to_ast` currently responds to CFG shapes in three different ways — lowers correctly (simple conditionals), raises `UnsupportedHugrStructureError` (loop + conditional), or silently emits a partial circuit (sequential conditionals). The third is not a missing fix so much as an inconsistent contract.

Suggested principle: enumerate the supported shapes and **raise on everything else**. Sequential conditionals then get either real recursive traversal or the explicit error — but not a wrong answer presented as a right one.

The structural change that would end this class: a completeness check asserting that every quantum operation in the HUGR appears exactly once in the emitted AST, raising otherwise. That single assertion would have caught the guessed targets, the invalid two-qubit gate, the swapped trailing measurements, sequential conditionals, and loop-plus-conditional — in one pass, without each needing to be reported. Across three rounds the pattern has been that the reported case gets fixed and its siblings survive; a completeness invariant breaks that cycle rather than extending it.

---

## Four durability notes

These are all cases where the current code works but encodes the fix in a form that will not age well. Raising them now while the code is warm.

**1. The array reversal deserves a root cause, not a refusal.** The error message prints `tag ids=[2, 1, 0], source ids=[0, 1, 2]` — PECOS holds both orderings at the point it gives up, so it knows the permutation exactly. The open question is *which side reverses*: whether guppylang 1.0 lowers `array(m0, m1, m2)` in reverse, or whether PECOS reads the collected order backwards. Neither the fix nor the tests determine this. Worth settling, because the answer picks the durable fix: if it is upstream, report it and bind by certified source order meanwhile; if it is PECOS, correct the extraction. As it stands, a documented public form that produced correct DEM bindings at the merge base now raises — that is a capability regression held in place by a guard. Fail-loud is the right interim state; a tracking issue would keep it from becoming the permanent answer.

**2. The `final` carve-out encodes a real fact in the most fragile available form.** The comment is correct that `final`'s QIS IDs are physical lowering order while its named slots are data-qubit semantic order, so a direct comparison would false-positive. But `base_tag != "final"` excludes a tag by literal string: any future generator with the same property needs another string added, and the root cause — two distinct orderings sharing one representation — stays in place. Representing physical-lowering order and semantic slot order as separately named things would let the guard compare like with like, remove the need for any exemption, and give the surface `final` slots real coverage rather than a documented exception.

**3. `BatchBuilder` over the hand-written mirror.** The descriptor hardening was a genuine improvement, and the new `api_version` check narrows the risk window. But `selene_runtime.rs` still carries a `#[repr(C)]` copy of an upstream `#[non_exhaustive]` struct plus a `std::mem::transmute`, so an upstream field addition still compiles cleanly and breaks at runtime. Upstream's `BatchBuilder::runtime_get_operation()` returns the real handle and removes both. It would also fix a latent hang: the drain loop terminates on `batch.invoked`, which `set_batch_time` sets, so a plugin emitting only timing spins forever — `BatchBuilder`'s `is_empty()` does not have that failure mode.

**4. Restore the tests, and drop the two lint names that do not exist.** The `result_tags.rs` `indirect_call_is_nontrivial_control_flow` tautology and the `selene_runtimes.rs` argument test that asserts a constant's shape back at itself are permanent blind spots rather than deferred work; likewise the relaxed `engine.rs` call assertions and `rz_calls >= 1`. Separately, `clippy::assert_is_empty` does not exist on any toolchain and `clippy::chunks_exact_to_as_chunks` is nightly-only — they suppress nothing and mask that the allow-list was validated against a different toolchain than CI runs. Minor and adjacent: `pyo3/abi3-py310` across the five crates now contradicts `Requires-Python: >=3.12`.

---

Overall this round moved in the right direction, and two of the fixes — the explicit `UnsupportedHugrStructureError` and the base/element mapping — are the durable kind rather than symptom patches. The remaining silent case and note 1 are the two worth resolving before merge; notes 2 through 4 are independently landable.

ciaranra:
## Release-lane break from the Python floor bump, and the durable fix for it

Separate from the correctness findings — this one is in the packaging chain, it does not show up in this PR's green CI, and it lands at release time.

### The break

`.github/workflows/python-release.yml:793-798` installs the built wheels with a 3.10 interpreter:

```
/opt/python/cp310-cp310/bin/python -m pip install --no-index \
  ./pecos-rslib-wheel/*.whl \
  ./pecos-rslib-llvm-wheel/*.whl
```

After this PR those wheels carry `Requires-Python: >=3.12`, which pip enforces on install — including a direct wheel path with `--no-index`. So `test_manylinux_2_28_wheels` should fail.

It is invisible here because that job is release-lane and path-filtered: it reported `skipping` on the first check run for this PR and is absent from the current one. Nothing in the PR lane can catch it.

(Reported from pip's documented `Requires-Python` behaviour and the wheel metadata this PR now produces — I did not execute it, having no 3.10 interpreter to hand. Worth a confirming run before relying on my read.)

### Why the one-line fix is not the right one

Changing `cp310-cp310` to `cp312-cp312` makes the job pass. It also leaves the packaging chain internally contradictory, because that interpreter is only one of four places the Python floor is encoded, and this PR has moved some of them but not others:

| Where the floor is stated | Current value |
| --- | --- |
| `requires-python` (all 10 workspace packages) | **3.12** — moved by this PR |
| Limited-API floor: `pyo3/abi3-py310` in 5 crates | 3.10 |
| Build interpreter: `CIBW_BUILD: "cp310-*"` (`python-release.yml:418`, `:567`) | 3.10 |
| Release smoke interpreter (`:793`) | 3.10 |

The wheels advertise `Requires-Python: >=3.12` while being built as `cp310-abi3` by a 3.10 interpreter. Patching only the fourth row keeps three rows disagreeing with the declared floor and leaves the next person to discover the mismatch the same way.

### The proper fix: align the whole chain on 3.12

The decision was effectively already taken in this PR when `pecos-rslib`, `-llvm`, `-cuda` and `-exp` moved to `requires-python >= 3.12`. Raising the limited-API floor to match is the change that makes the rest coherent, and it is safe: a higher `abi3-py*` target only widens the available limited API, so nothing can fail to compile. pyo3 0.29 ships the `abi3-py312` feature.

1. `pyo3/abi3-py310` → `pyo3/abi3-py312` in `python/pecos-rslib/Cargo.toml:40`, `python/pecos-rslib-llvm/Cargo.toml:23`, `python/pecos-rslib-cuda/Cargo.toml:24`, `python/pecos-rslib-exp/Cargo.toml:23`, `exp/zluppy/Cargo.toml:27`.
2. `CIBW_BUILD: "cp310-*"` → `"cp312-*"` at `python-release.yml:418` and `:567`. This one is required by the change rather than optional: leaving cibuildwheel on a 3.10 interpreter while requesting the 3.12 limited API is at best incoherent, and I have not verified whether pyo3 rejects that configuration outright or merely warns.
3. `/opt/python/cp310-cp310` → `/opt/python/cp312-cp312` at `python-release.yml:793-798`.
4. Update the explanatory comments that name the old floor: `Cargo.toml:55` and `python/pecos-rslib/Cargo.toml:29-31`.

User impact: none. The wheel tag becomes `cp312-abi3`, but 3.10 and 3.11 users are already excluded by `Requires-Python: >=3.12` as of this PR, so no install that works today stops working. The existing `abi3audit --strict` steps continue to validate limited-API conformance and need no change.

### The part that makes it durable

The four rows above drifted apart because nothing checks them against each other. `python-version-consistency` sounds like it would, but `scripts/check_python_workspace.py` validates package **versions**, uv workspace membership, internal `==` pins and the cuda extra pairing — it never reads `requires-python`, classifiers, the abi3 features, or the CI interpreters. It exits 0 on this PR today, with the mismatch in place.

The lasting fix is to extend that checker to assert the invariant directly: every package's `requires-python` floor, every crate's `abi3-py*` feature, `CIBW_BUILD`, and the release smoke interpreter all name the same minimum version, and the classifiers enumerate exactly the supported set. That converts a hand-alignment anyone can silently undo into a CI failure, and it is the same class of guard as the completeness check suggested for the HUGR converter: assert the invariant once rather than re-discovering each violation as a separate bug report.

Without it, this PR aligns the rows by hand and the next floor bump gets to find the stragglers again — which is how the 3.10 entries this PR is still cleaning up came to exist.

### Related stragglers in the same family

- `exp/zlup/pyproject.toml:5` is still `requires-python = ">=3.10"`. Not a uv workspace member, so the checker never sees it — an example of the gap above.
- `scripts/check_python_workspace.py:21-25` and `python/quantum-pecos/tests/pecos/test_selene_plugin_workspace.py:10` still carry `tomli` fallbacks for "Python 3.10". `tomllib` has been stdlib since 3.11, so both branches are unreachable under a `>=3.12` floor. The checker's own error text still says "Python 3.11+ or the 'tomli' package is required".

ciaranra:
## Re-review at `1cdc072` — blocking findings resolved

All items I raised as blocking are closed. Verified by re-running the earlier reproductions against this head rather than by reading the diff.

| Finding | Status |
| --- | --- |
| Array result-ID reversal | Fails loudly; new test pins the exact shape `array(m0, m1, m2)` and asserts the raise |
| Guessed qubit targets / CX arity | Fixed; re-checked for regression — `MeasureOp[0]`, `CX[1, 2]` |
| Two sequential conditionals | Now raises `UnsupportedHugrStructureError` — this was the last silent-wrong path |
| Loop + conditional | Raises |
| abi3 chain | Five crates on `abi3-py312`, both `CIBW_BUILD` on `cp312-*`, manylinux smoke on `cp312-cp312`, comments updated |
| Two non-existent clippy lints | Removed |
| `exp/zlup` floor, `tomli` fallbacks | Bumped / removed |

`tests/selene` + `tests/qec`: 1514 passed, 1 skipped, 2 xfailed.

### On the new consistency guard

I mutation-tested `scripts/check_python_workspace.py` rather than trusting it. Each of the four rows fails independently, with a message naming the file and the expectation:

```
[abi3]  ABI3 feature must be exactly 'abi3-py312', found ['abi3-py310']
[cibw]  expected two 'CIBW_BUILD: "cp312-*"' release-wheel targets
[floor] [project].requires-python must be >=3.12
[smoke] expected four '/opt/python/cp312-cp312/bin/python' smoke commands
```

That is the durable half of the packaging fix — the alignment can no longer silently rot, which is what let the four rows drift apart in the first place.

The converter's contract is also coherent now: it lowers the shapes it supports and raises on the rest. A completeness invariant would still be stronger than shape-by-shape rejection, since it would cover shapes nobody has thought to enumerate — but the safety property is achieved either way.

### Not blocking, still open

- The hand-written `#[repr(C)]` mirror and `std::mem::transmute` in `selene_runtime.rs`, and the descriptor re-resolved at all 15 per-operation call sites where upstream resolves once at load.
- `Operation::Custom` still discards its tag/data/len and is dropped silently — the same shape as the RPP bug that was fixed, one match arm below it.
- No `rust-version` declared, despite `as_chunks_mut` putting the effective floor at Rust 1.88.
- Four tests remain weakened: the `result_tags.rs` `indirect_call_is_nontrivial_control_flow` tautology, the `selene_runtimes.rs` argument test that asserts a constant's shape back at itself, two relaxed `engine.rs` call assertions, and `rz_calls >= 1`.

Each is independently landable and none gates this PR.

### Scope of this review

#489 correctly parks the array-output capability with the right question attached, and I have added some evidence there that should narrow it.

To be clear about what this review does and does not certify: it means the findings I raised are closed, not that the PR is defect-free. The panel never ran the full test suite, never exercised a real selene-sim host end-to-end, and reviewed the regenerated `.hugr` fixtures structurally rather than semantically. Merge readiness and CI are yours to call.

ciaranra:
## Adversarial cross-review of `6eebab3..4fb593a`

The commits after `6eebab3` — including a +320/−171 rewrite of `hugr_to_ast.py` and a new +90-line consistency guard — had not had independent review; my earlier rounds verified them only by re-running the previous findings' reproductions. This round covers that window with a fresh panel: three Claude Opus 5 subagents (converter behaviour, consistency guard, and a boundary-probing arm) plus an OpenAI Codex CLI read-only arm, all blind to each other, with fusion and verification by Claude Opus 5. Every claim below was re-checked against the code or by execution.

### First, a concern of mine that turned out to be unfounded

I went in expecting the new `UnsupportedHugrStructureError` paths to be over-broad — rejecting programs that previously converted correctly. **They are not.** One arm ran 36 Guppy programs against each historical version of the converter inside a single guppylang 1.0.1 environment, which isolates the converter change from the Guppy version bump. Every shape now rejected was already producing wrong output beforehand:

| program | at `6eebab3` | at `4fb593a` |
| --- | --- | --- |
| `if a and b: x(q2)` | `WHILE m1: <whole program>`, X dropped | rejected |
| `cx(q1, q3)` after an `if`, 3 live qubits | `CX q[3]` — arity 1 | **correct** |
| `cz(q1, q0)` | `CZ q[0,0]` | **correct** |
| conditional with both arms returning | wrong qubit in the else arm | **correct** |

Confirmed correct at this head across straight-line 1–3 qubit programs, `if` with and without `else`, `if` plus a straight-line tail, both-arms-return, two-qubit gates inside a branch, asymmetric branches with a join, a qubit allocated inside a branch, empty-then, reversed operand order, and inlined helper calls. The rewrite is a substantial net correctness win, and the rejections are the right stopgap versus the silent qubit corruption they replaced.

Also retired: the shipped `guppy_gen` generators were already unconvertible at `6eebab3` (via `TailLoop`), so this range did not newly break them.

### Items this PR owes

**1. The consistency guard fires on legitimate maintenance edits.** Both reproduced:

```
[add a third, CORRECT `CIBW_BUILD: "cp312-*"`]   error: expected two 'CIBW_BUILD: "cp312-*"' release-wheel targets
[normalize quoting to 'cp312-*']                 error: expected two 'CIBW_BUILD: "cp312-*"' release-wheel targets
```

The first is the problem: the check counts occurrences rather than validating each one, so *more* compliant targets is an error. `pecos-rslib-cuda` and `-exp` already carry abi3 features but have no release job — adding one is the obvious next step and hits this immediately. The four-occurrence smoke-command check has the same shape: hoisting the repeated interpreter path into a shell variable, or adding a fifth assertion, both fail. A guard that blocks unrelated work gets deleted rather than fixed.

Root cause is self-inflicted: `python-version-consistency.yml` runs bare `python` with no dependency install, so the script is stdlib-only and cannot parse YAML. Adding PyYAML (or using `uv run`, as the Justfile already does) allows reading `jobs.*.steps[*].env.CIBW_BUILD` and asserting each value equals `cp312-*` with no count constraint. The abi3 check has the same fix available for free — `tomllib` is already imported, so parsing `[features]` instead of regexing the file text would stop comments from tripping it.

Related false negatives worth closing while you are in there: `ABI3_MANIFESTS` is a hardcoded five-entry tuple, so a new crate added with `abi3-py310` passes green — and because pyo3 selects the *lowest* enabled `abi3-py3X` across unified features, that one crate would silently lower the whole workspace floor. The `exp/` pyproject paths are hardcoded too, while the workflow trigger is already the broader `exp/**/pyproject.toml`.

**2. `exp/zlup/uv.lock` is stale.** `exp/zlup/pyproject.toml:5` moved to `>=3.12` but `exp/zlup/uv.lock:3` still records `>=3.10`. Verified: `cd exp/zlup && uv lock --check` reports *"The lockfile at `uv.lock` needs to be updated"*. No workflow references `exp/zlup`, so CI does not catch it. One `uv lock` in that directory.

**3. Validation ordering gives the wrong diagnosis for every TailLoop program.** `_analyze_cfg` (which now calls `_validate_control_flow_shape`) runs before `_check_for_loops` at `hugr_to_ast.py:256-259`, so the accurate message is shadowed:

```
make_surface_code(distance=3, num_rounds=1, basis="Z")
  at 6eebab3:  "HUGR contains TailLoop structure (while/for loop)."   <- accurate
  at 4fb593a:  "HUGR CFG has sequential or nested conditionals"       <- wrong reason
```

Diagnostics only — these never converted — but it will mislead whoever picks up the converter work.

**4. `for _ in range(n)` is now rejected** (`hugr_to_ast.py:409-431`), where it previously produced a `WHILE` with a fabricated condition and a lost trip count. Strictly an improvement in honesty, but `for` loops are ordinary QEC vocabulary and there is now no path for them at all. Worth an explicit note in whatever documents the supported envelope.

### Pre-existing, and explicitly not blocking this PR

Two arms independently found, and I reproduced, defects that predate #480 — I confirmed against the merge base that `params` was never populated and `_get_condition_variable` was already used the same way. These should not gate a dependency upgrade, but they are more severe than anything above and deserve their own issue:

- **Branch conditions are fabricated, not read from the predicate wire** (`hugr_to_ast.py:988-997`). `_get_condition_variable()` returns the last measurement seen. Reproduced: with `a` measured before `b`, `if a: x(q)` emits `IfStmt cond=VarExpr(name='m1')` — the wrong measurement. Reported alongside: `s = not r; if s:` loses the negation, and a purely classical `if n > 2:` invents `m0`, producing QASM that reads a creg *before* the measurement that fills it, while `validate()` returns `valid=True`.
- **While conditions and trip counts are lost the same way** (`:964-966`); `test_simple_loop` asserts the body and qubit index but never the condition.
- **Rotation angles are dropped** (`:1117-1123`). Reproduced: `rz(q, pi/4)` emits `RZ params=()`.
- **CFG output ports are renumbered with `enumerate`** (`:1029`), so any classical value carried across a conditional shifts every qubit binding. The correct approach already exists in the same file at `_capture_cfg_qubit_ports:811-851`, which filters by typed port.
- **No notion of function boundaries** (`:312`) — all `FuncDefn`s' blocks merge into one dict and `Call` is never handled. The shipped d=3 surface HUGR has 4 `FuncDefn`, 5 `CFG`, 20 `DataflowBlock`, 3 `TailLoop`, 3 `Call`.
- **A silent flat fallback** at `:734-742` concatenates blocks in dict order when no pattern is recognised, discarding control flow — contrary to the fail-loud stance the rest of this work adopts.

The context that frames all of these: **every Guppy program PECOS itself ships is currently unconvertible.** All six `guppy_gen` factories raise, before and after this PR. The converter has only ever handled hand-written toy circuits, and the subset that does convert is the more dangerous half, because it yields plausible ASTs with wrong conditions that validation accepts. That is an argument for marking this module experimental until the predicate wire is actually read — not an argument against this PR.

### Verdict

The four items above are small and belong here; items 1 and 2 are the ones I would not merge without, since a guard that blocks legitimate edits will not survive contact and a stale lockfile is one command. The converter work in this range is a genuine improvement and I would ship it as-is.

Scope of this review, as before: it certifies that the findings raised were examined, not that the PR is defect-free. No panel ran the full suite or exercised a real selene-sim host end-to-end.

ciaranra:
## Panel on `4fb593a..3a6ec79`, plus the physics check nobody had run

Two things to report: the toolchain upgrade produces **correct physics** — which closes the caveat I have attached to every previous comment — and four small items in this range that I would fix here. Pre-existing defects found along the way are filed separately (#497, #498, #499) rather than raised as blockers.

Panel: three fresh Claude Opus 5 subagents (physics equivalence, provenance end-state, workspace guard) plus an OpenAI Codex CLI read-only arm on the three newest commits, all blind to each other; fusion and verification by Claude Opus 5. Every item below was re-checked against the code or by execution.

### The consistency guard rewrite is correct in both directions

I re-ran the mutation matrix from my last comment. The false positives are gone and nothing was loosened:

| edit | before | now |
| --- | --- | --- |
| add a third, correct `CIBW_BUILD` | error | **passes** |
| normalize quote style | error | **passes** |
| downgrade abi3 / CIBW / floor / smoke interpreter | caught | caught, with value-based messages |
| **new crate carrying `abi3-py310`** | **passed green** | **caught** |

Parsing YAML under `uv run --frozen` instead of counting strings was the right call, and globbing manifests closed the false negative as well. `exp/zlup/uv.lock` now passes `uv lock --check`.

### Physics verification

Seven simulator backends, 59 analytic programs, seeded and reproducible. Every deterministic prediction held with **zero violations** over tens of thousands of shots; every probabilistic one landed within ~2σ. The hazards this PR specifically created were all exercised directly and are clean:

- Conditionals fire in both polarities — `if b` and `if not b` each correlate 20000/20000.
- Teleportation of |1⟩ recovers 1 in 20000/20000 with both Bell bits 50/50; |+⟩ read in X basis gives 0 in 20000/20000.
- Loop measurements are independent and correctly ordered — 8 fresh coins per shot, adjacent-bit agreement 0.5, deterministic pattern `[1,0,0,1,0,0,1,0]` reproduced 4000/4000.
- Resets exact (post-reset 0 in 20000/20000); no cross-shot contamination (lag-1 and Wald–Wolfowitz |z| ≤ 1.05 on all seven backends).
- All four Bell states pinned in two bases; GHZ3/GHZ4 with zero forbidden strings.
- Surface-code LER decreases with distance below threshold, increases above it, ≈50% at p=0.3.

Clifford-only backends refuse non-Clifford programs loudly rather than answering wrongly. So the batched-operation ABI, the measurement-future rewrite, and the per-shot state clearing are all behaving.

### Four items I would fix in this PR

**1. `tket.global_phase` and `tket.modifier` are missing from the extension registry this PR curated.** `crates/pecos-hugr/src/loader.rs` was edited here to drop `bool` and add `guppy` and `measurement`, but `global_phase` was not added. Consequence: `rz(q, angle(2.0))` fails to load on both PECOS paths (`Failed to load HUGR`) while direct Selene runs it — at exactly ±2 half-turns guppylang drops the `Rz` and emits `tket.global_phase.global_phase`. ±1, ±1.5, 2.5, 3 and 4 half-turns all work, so this is a narrow but real hole. The same omission is in `crates/pecos-hugr-qis/src/utils.rs`, and `crates/pecos-phir/src/hugr_parser.rs` is missing four.

tket 0.21 exposes `tket::extension::tket_extensions() -> [Arc<Extension>; 8]` containing exactly the right set, including `global_phase` and `modifier`. Using it at all three sites removes the need to maintain the list by hand.

**2. `TailLoop` bypasses the new PHIR rejection.** `hugr_parser.rs:264` rejects `Conditional | Call | CallIndirect | LoadFunction`, but `TailLoop` appears nowhere in the file, and `process_children` only handles extension ops — so a `TailLoop` container is skipped and its nested quantum operations are silently omitted. That is the exact failure mode the commit was written to close, left open for one container type.

**3. The same rejection is scoped too broadly.** `reject_control_flow` iterates `hugr.nodes()` — every node in the module — before `find_operations_container` selects the entrypoint. An unused helper containing a conditional therefore rejects an otherwise straight-line entrypoint that previously converted. Restricting it to the entrypoint's reachable hierarchy would fix that without weakening item 2.

**4. The workflow path triggers do not cover `crates/**/Cargo.toml`.** The script now globs every manifest, but `python-version-consistency.yml` triggers only on `pyproject.toml`, `python/**`, `exp/**` and the root `Cargo.toml`. Since the workspace includes `crates/pecos*` by glob, a new crate there with a stale abi3 feature still escapes CI. Script fixed, trigger not.

### On the two provenance guards

#498 documents a live mis-binding in the SZZ surface path: `abstract == traced_qis` holds for the default path in both bases and fails for SZZ in both, and correcting the reversed `final` IDs restores parity. I reproduced this independently.

The underlying reversal may well predate this PR — I could not establish that either way. What is in scope here is that the two guards added during this review are shaped so the case passes: `decode.py:1049` accepts a contiguous run in either direction, and `dem.py:897` excludes the `final` tag by name, which is the only array-valued output the surface path has. Renaming the tag makes the guard fire with the correct diagnosis. Whatever is decided about the root cause, those two lines are worth revisiting here, since they were written in response to my earlier comments.

### One pattern, four instances

Each of these came from a hand-maintained enumeration that missed a member: the extension registry (missed `global_phase`), the PHIR rejection denylist (missed `TailLoop`), `ABI3_MANIFESTS` (missed unlisted crates), the CIBW occurrence counts (missed a third job). Three have already been converted to derived or globbed forms in this PR, and they are the ones that now behave. `tket_extensions()` is the same conversion available for the fourth; an allowlist walk — raise on any node that is neither a handled op nor an explicitly-known-safe container — is the equivalent for the PHIR rejection.

### Filed separately

- #497 — `mast` silently returns 0 for every non-Clifford rotation. Isolated away from Selene to `exp/pecos-stab-tn`; untouched by this PR. Shipped plugin, silently wrong.
- #498 — the SZZ final-array reversal above.
- #499 — internal HUGR engine: repeated `result()` tags truncated to the last occurrence (the only other silent data loss found), plus `project_z` stalls and no qubit recycling, both fail-loud.

As always, this certifies that what was examined held up, not that the PR is defect-free.

ciaranra:
## Verification at `2643800` — all four items from my last comment are fixed

Confirmed by execution, not by reading the diff:

| item | evidence |
| --- | --- |
| Extension registries | Hand-maintained lists replaced by `tket::extension::tket_extensions()` at all three sites. The previously-failing `rz(q, angle(2.0))` (exact ±2 half-turns → `tket.global_phase`) now loads and runs end-to-end on both the internal HUGR engine and the Selene path, with the expected ~50/50 outcomes after an H |
| `TailLoop` bypass | In the rejection match (`hugr_parser.rs:269`); 147 `pecos-phir` tests pass including the new coverage |
| Rejection scope | Now `once(entrypoint).chain(hugr.descendants(entrypoint))` rather than every module node, so unreachable helpers no longer poison a straight-line entrypoint |
| CI trigger | `crates/**/Cargo.toml` added for both `push` and `pull_request` |

Rust suites fresh-compiled: 272 tests green. CI: 87 pass, 7 skipped.

**A trap for anyone verifying locally:** `test_global_phase_lowering_loads` fails against a venv whose `pecos_rslib_llvm` wheel predates this commit — the registry fix lives in Rust, so Python cannot see it until the wheel is rebuilt (`just build-debug`). I hit exactly this false failure before rebuilding; after the rebuild it passes 6/6. CI builds wheels fresh, so it is not affected.

### The one open thread

The only unresolved item from my reviews is the pair of provenance guard lines tied to #498 (the SZZ reversed `final` binding): `decode.py:1049` accepts descending contiguity, and `dem.py:897` excludes the `final` tag from the array cross-check. Since the reversal's root cause (Guppy/Selene lowering vs. PECOS trace capture) is undetermined and tracked in #498, resolving it there rather than in this PR is a reasonable call — I flag it only so the merge decision is made with it in view.

Everything else raised across these review rounds is closed, converted into tests, or parked with a tracking issue (#489, #493, #494, #497, #498, #499).

ciaranra:
## One last request before merge: make SZZ fail loud instead of silently wrong

From the review side this is the final item. Everything else across the rounds is closed and verified; this PR is otherwise merge-ready.

As #498 documents, the SZZ surface path currently ships a DEM bound to reversed `final` measurement IDs — silently. The root cause is undetermined and rightly tracked there, not here. But this PR's own arc has been converting silent-wrong into fail-loud (the converter rejections, the PHIR rejections, the provenance guards), and the SZZ path is the one remaining case where a known-wrong result is still produced without complaint. The same conversion is available here in two lines plus one test annotation:

1. **`decode.py:1049`** — accept ascending contiguity only:
   ```python
   step = 1 if final_ids[1] > final_ids[0] else -1
   ```
   becomes ascending-only (drop the `-1` branch). A descending `final` then raises the existing "contiguous in one source-order direction" error instead of binding reversed.

2. **`dem.py:897`** — remove the `base_tag != "final"` exclusion from the array cross-check. With it gone, the guard fires on the SZZ case with the correct diagnosis (verified during review by renaming the tag: *"has measurement id 20, but the generator-certified layout requires 12"*).

**Expected fallout, stated up front:** `test_from_guppy_szz_surface_code_is_byte_identical_to_reference` will fail — it currently passes *with* the reversed binding, because it feeds `ref.get_meta("detectors")` from the traced path into `from_guppy`, so both sides share whatever binding decode.py produced. It should become `xfail(strict=True, reason="#498: SZZ final provenance reversed")`, the same pattern this PR already uses for the canonical gate-local Z case. Any other SZZ-path test that starts raising is telling the truth and should be xfailed against #498 likewise.

The honest consequence is that SZZ becomes unusable-but-loud until #498 is resolved, instead of usable-but-wrong. Every non-SZZ configuration is unaffected — during review, `abstract == traced_qis` DEM equality was verified to hold for the default path across d ∈ {3,5}, rounds ∈ {0,1,2}, both bases.

If the judgment instead is that an operational-but-wrong SZZ is preferable in the interim (e.g. because the d=3 mirror may be benign for symmetric patches — unproven either way), that is a defensible owner's call; in that case merge as-is with #498 open, and this comment stands only as the record that the choice was made deliberately.

qartik:
@ciaranra can this be approved and merged now?

ciaranra:
Yeah, I was trying to get a bunch of other stuff merged in and considering holding back on this one and creating a pre-release right before this since it will disrupt a couple of people... but it is probably fine... they can adapt or use the older version of PECOS.

PR #486, opened 2026-08-12
Support indexed assignment and boolean-mask indexing on Array (#482)

### LINKED ISSUE #482: Array.__setitem__ does not support indexed (fancy) assignment
## Summary

`Array.__setitem__` accepts only integers and slices in an index tuple
(`python/pecos-rslib/src/pecos_array.rs:1445`), so NumPy's indexed ("fancy") assignment is
unavailable:

```python
a = zeros((2, 4), dtype=dtypes.uint8)
a[0, [1, 3]] = 1
# TypeError: indices must be integers or slices
# NumPy: [[0, 1, 0, 1], [0, 0, 0, 0]]
```

Found while migrating `examples/surface/dem_decomposition_diagnostics.py` for #458 step 3a,
where `detection_events[index, detectors] = 1` sets a whole row's detector positions at once --
a natural way to build dense detector rows from a sparse key list.

The equivalent per-element loop works and produces identical results, so this is not a blocker,
but it is O(n) Python for something the layer should do in one call, and `decode.py` (#458 step
4, 113 sites) is very likely to hit the same shape when building detection-event arrays.

## Scope

Support sequence-of-index assignment in `__setitem__` for at least the 1-D and
`(int, sequence)` forms, matching NumPy's semantics: out-of-range indices raise `IndexError`,
negative indices count from the end, and a scalar right-hand side broadcasts to every selected
position. Reading (`a[0, [1, 3]]`) should be considered at the same time if the same machinery
covers it -- say so either way.

NumPy is the test oracle. Note that PECOS deliberately rejects some NumPy coercions (see the
"divergences from NumPy" list in `docs/development/DEVELOPMENT.md`); assignment value checking
should follow the existing checked-conversion path rather than adding a new one, so all ingest
routes keep agreeing.

Blocks nothing outright; makes #458 steps 3a and 4 cleaner.

ciaranra:
## Summary

Closes #482. `Array.__setitem__` accepted only integers and slices, so NumPy's indexed
assignment was unavailable:

```python
a = zeros((2, 4), dtype=dtypes.uint8)
a[0, [1, 3]] = 1     # was: TypeError: indices must be integers or slices
```

That is how dense detector rows get built from sparse keys, and #458 step 3a had to fall back to
a per-element loop. Boolean-mask reads (`a[a != b]`) failed for the same reason. Both now work.

## What changed

Index normalization was **extracted into one resolver shared by reads and writes** rather than
duplicated -- `__getitem__` already did fancy indexing, so writes reuse it. Assignment values go
through the constructor/`fill` checked-conversion path, so out-of-range raises rather than wraps
and every ingest route stays in agreement (the defect class #462's review found four instances
of). Boolean-mask **assignment** fell out naturally and is supported.

## Honest limits, raised rather than half-implemented

- **1-D masks only.** A 1-D mask on a multi-dimensional array selects along the indexed axis,
  matching NumPy's row selection. A multi-dimensional mask raises `NotImplementedError` -- NumPy's
  flattened elementwise semantics need a different execution model, and a partial version
  pretending to be complete is worse than a clear limit.
- **One fancy-indexed axis per tuple.** More raises explicitly, rather than partially
  implementing NumPy's advanced-index broadcasting.

**No operators were added.** `Array ^ Array` was explicitly excluded: `__and__` on `Array` is the
Kronecker product, so a bitwise `__xor__` would put `&` and `^` in different algebras on one
type. That is an API-taste decision for the maintainer, not part of a gap fix.

## Verified against NumPy

Every form checked by execution, matching NumPy exactly: the motivating
`detection_events[i, dets] = 1`, 1-D scalar RHS, negative indices, **duplicate indices
(last-write-wins)**, mask reads, mask assignment. Guards each raise the right type naming real
values: `IndexError: index 5 is out of bounds for axis 0 of length 2`, RHS shape mismatch,
`OverflowError: value 256 is out of range for uint8`, mask-length mismatch, and the 2-D mask
limit.

## Verification

clippy `-D warnings` clean; fmt clean; pecos-rslib **1572** passed; qec **1489**; pre-commit two
passes exit 0; 0 behind dev. Mutation table in the implementation record covers the bounds check,
length-match check, value range check, and mask-length check.

### DIFF
--- a/python/pecos-rslib/src/pecos_array.rs
+++ b/python/pecos-rslib/src/pecos_array.rs
@@ -67,11 +67,18 @@ pub enum ArrayData {
     PauliString(ArrayD<PauliString>),
 }
 
-/// Represents an indexing operation: either an integer index or a slice
-#[derive(Debug, Clone, Copy)]
+/// Represents an indexing operation.
+#[derive(Debug, Clone)]
 enum IndexOp {
     Integer(isize),
     Slice(isize, isize, isize),
+    Fancy(Vec<usize>),
+}
+
+struct AssignmentSelection {
+    ranges: Vec<Vec<usize>>,
+    integer_axes: Vec<usize>,
+    shape: Vec<usize>,
 }
 
 impl ArrayData {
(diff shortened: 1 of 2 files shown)


ciaranra:
Independent pass complete, four findings, all fixed and verified by execution (`ec6fb59c3`).

**The critical one was a regression this PR would have introduced.** Mixed tuples where an
integer/fancy index is *separated by a slice* silently wrote to the wrong elements:

```
a[1, 0:2, [0, 2]] = [[100, 101], [102, 103]]
  PECOS  [[100, 10, 101], [102, 13, 103], ...]
  NumPy  [[100, 10, 102], [101, 13, 103], ...]
```

No exception, no warning. Provenance checked against `origin/dev`: that index form raises
`TypeError` there for both read and write, so this PR turned a loud refusal into a quiet wrong
answer -- strictly worse than the gap it was closing. Root cause is NumPy's advanced-index axis
placement: when advanced indices are separated by a slice the broadcast dimension moves to the
front of the result; contiguous ones stay in place.

Resolved by **detecting and refusing** the separated case with a named `NotImplementedError`,
rather than approximating the rule. Contiguous fancy tuples remain supported and NumPy-matching.
Correctly implementing axis relocation needs a different execution model -- the same reason
multi-dimensional masks are refused -- and a partial version pretending to be complete is exactly
what produced this finding.

The other three, all verified fixed:

| finding | before | now |
|---|---|---|
| self-assignment `p[[1,0,3,2]] = p` | `RuntimeError: Already mutably borrowed` | `[1, 0, 3, 2]`, matching NumPy; RHS snapshotted before the mutable borrow |
| Pauli/PauliString array-like RHS | rejected via `__array_interface__` | native Pauli `Array` and NumPy object arrays both assign correctly |
| nested-list multidimensional mask | `'list' object cannot be interpreted as an integer` | the same accurate multidimensional-mask `NotImplementedError` native masks get |

Reported clean by the same pass: 3-D/4-D last-axis and first-axis fancy writes, boolean-mask
assignment in 3-D/4-D, empty and zero-sized selections, selection-based aliasing and overlaps, 52
dtype comparisons across all four assignment forms, non-contiguous transpose and stepped-slice
targets, and the multiple-fancy-axes limit.

Verification: clippy `-D warnings` clean; fmt clean; pecos-rslib **1591** passed; qec **1494**;
docs **271** passed, 0 failed; pre-commit two passes exit 0; 0 behind dev.

PR #491, opened 2026-08-13
Implement NumPy's advanced-indexing execution model on Array (#488)

### LINKED ISSUE #488: Array indexing cannot express NumPy's advanced-index execution model
## Summary

`Array` indexing walks axes sequentially, which cannot represent NumPy's advanced-indexing
execution model. Three capabilities are consequently refused with `NotImplementedError`:

| refused form | example | NumPy |
|---|---|---|
| advanced indices separated by a slice | `a[1, 0:2, [0, 2]]` | broadcast axis moves to the front of the result |
| multiple advanced indices in one tuple | `a[[0, 1], [2, 3]]` | broadcasts the index arrays together |
| multi-dimensional boolean masks | `a[mask_2d]` | flattens to elementwise selection |

These are **one root cause, not three**. All three need the same thing: resolve the full index
tuple into a broadcast selection first, then execute, rather than applying axes one at a time.

## Why this is a gap to close rather than a documented limit

The refusals are safe -- they were introduced in #486 specifically to replace a *silently wrong*
write, and refusing loudly is strictly better than corrupting quietly. But they are workarounds
for an unimplemented execution model, not principled boundaries: NumPy's advanced-indexing rules
are fully specified and implementable, so "we do not support this" is a statement about PECOS's
internals rather than about the semantics.

Contrast with `ravel`/`reshape` always copying (#470): that one *is* structural -- `Array` owns
its buffer and there is no view type -- and is documented as a deliberate divergence.

## Not currently blocking

`decode.py`, the largest remaining #458 migration target, uses `[:, q]` and `[:, 1::2]` forms
that already work. Nothing in the tree needs the refused forms today. The risk is the usual one:
the last three layer gaps (#481, #482, #483) were all "not blocking" until real code met them.

## Scope

One execution model covering all three:

1. Parse the whole index tuple into basic (int/slice) and advanced (integer-array/bool-mask)
   components.
2. Broadcast the advanced components together; apply NumPy's result-axis placement rule --
   broadcast dimensions go to the front when advanced indices are separated by a slice, and stay
   in position when contiguous.
3. Execute reads and writes through that resolved selection, sharing one path (as #486 already
   does for the sequential case).

NumPy is the oracle throughout. The acceptance test is the set of forms currently refused, plus
everything #486's review verified as already matching, so the change cannot regress what works.

ciaranra:
## Summary

Closes #488. Replaces `Array`'s sequential axis-by-axis indexing with NumPy's advanced-indexing
execution model: parse the whole index tuple, broadcast the advanced components together, apply
the result-axis placement rule, then execute reads and writes through one resolved selection.

This removes three `NotImplementedError` refusals that were **workarounds, not principled
limits** -- the semantics were fully specified, PECOS simply could not express them:

| previously refused | now |
|---|---|
| advanced indices separated by a slice, `a[1, 0:2, [0, 2]]` | matches NumPy, read and write |
| multiple advanced indices, `a[[0, 1], [2, 0]]` | matches NumPy |
| multi-dimensional boolean masks, `a[mask_2d]` | matches NumPy |

#486 introduced the first refusal deliberately, to replace a *silently wrong* write its review
caught. Refusing was the right stopgap; this is the actual fix.

## The rule, made falsifiable

NumPy moves broadcast result dimensions to the **front** when advanced indices are separated by a
slice, and leaves them **in place** when adjacent. That distinction is asserted on result
*shapes*, not just contents, because shape is where the rule shows:

- adjacent: `(2, 3, 5)`
- separated: `(3, 2, 4)`
- the unequal-length case `a[1, 0:3, [0, 2]]` that #486 wrongly rejected: `(2, 3)`

Verified independently across reads and writes, 3-D and 4-D, with scalar and correctly-shaped
array right-hand sides.

## Still unsupported, and honestly so

`Ellipsis`, `newaxis`/`None`, and zero-dimensional boolean indices raise explicit
`NotImplementedError`. These are *shape-expansion* semantics -- a different feature from advanced
indexing -- and are refused rather than guessed at. Non-scalar assignment RHS must still match the
selection shape exactly.

## Regression bar

Everything #486's independent pass verified must still match, and does: 1-D through 4-D
assignment with scalar and sequence RHS, negative indices, duplicate indices (last-write-wins),
1-D mask reads and assignment, empty index lists, empty and all-false masks, zero-length arrays,
shapes `(0, 3)` and `(3, 0)`, selection-based aliasing and overlap, direct self-assignment, 52
dtype comparisons including Pauli and PauliString with native-Array and NumPy-object RHS, and
non-contiguous transpose / stepped-slice targets.

## Verification

clippy `-D warnings` clean; fmt clean; pecos-rslib **1631** passed; qec **1494**; docs **271**
passed, 0 failed; pre-commit two passes exit 0; 0 behind dev. Mutation table in the
implementation record covers the adjacent/separated decision, the broadcast-shape check, and
bounds.

### DIFF
--- a/python/pecos-rslib/src/pecos_array.rs
+++ b/python/pecos-rslib/src/pecos_array.rs
@@ -67,11 +67,18 @@ pub enum ArrayData {
     PauliString(ArrayD<PauliString>),
 }
 
-/// Represents an indexing operation: either an integer index or a slice
-#[derive(Debug, Clone, Copy)]
+/// Represents an indexing operation.
+#[derive(Debug, Clone)]
 enum IndexOp {
     Integer(isize),
     Slice(isize, isize, isize),
+    Fancy(Vec<usize>),
+}
+
+struct AssignmentSelection {
+    ranges: Vec<Vec<usize>>,
+    integer_axes: Vec<usize>,
+    shape: Vec<usize>,
 }
 
 impl ArrayData {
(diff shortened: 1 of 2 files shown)


ciaranra:
Independent pass complete: **1,149 differential checks against NumPy, zero correctness
mismatches.** No high or medium findings.

The pass was weighted toward regression rather than the new capability, because this PR replaces
machinery that previously-working forms also use. All of it still matches NumPy:

- 840/840 checks across all 15 dtype variants (bool, every signed/unsigned width, float32/64,
  complex64/128, Pauli, PauliString), for scalar, native-`Array`, and NumPy right-hand sides.
- Negative and duplicate indices including last-write-wins; empty lists; all-false masks;
  zero-length arrays; shapes `(0, 3)` and `(3, 0)`; selection overlap both directions; direct
  self-assignment.
- 3-D to 5-D axis placement with advanced indices first, last, middle, adjacent, separated, with
  scalar integers between them, and with unequal-but-compatible broadcast shapes -- shapes as
  well as contents.
- Transposed, stepped, reversed, and transpose-plus-step targets across every dtype.
- The deliberate refusals raise accurately on reads and writes: `Ellipsis indexing is not
  supported`, `newaxis indexing is not supported`, `zero-dimensional boolean indices are not
  supported`.

## The one Low finding is pre-existing and structural

Transposes and slices return arrays that own their data, so writes through them do not reach the
parent, where NumPy's views do. Verified on `dev`: this predates the advanced-indexing work and
was surfaced only because this PR's review swept non-contiguous targets. It shares its root cause
with `ravel`/`reshape` always copying (documented divergence, #470) -- `Array` owns its buffer and
there is no view type. Filed as #500 with the three options; the genuine defect there is that the
divergence is undocumented, not that copies are wrong.

Nothing outstanding on this PR.

PR #492, opened 2026-08-13
Reject archive entry names that filesystem normalization would redirect

ciaranra:
Closes a path-traversal bypass in the archive containment guard added by #472. Found by an independent adversarial review of that already-merged code, and reproduced before fixing.

## The bypass

`contained_entry_path` rejected `..` by matching `Component::ParentDir`. An entry named `".. /nvcc"` — `..` followed by a **trailing space** — does not produce a `ParentDir`; Rust parses `".. "` as an ordinary `Component::Normal`, so it passed validation. Windows then strips trailing spaces and dots while resolving the path, turning it back into `../nvcc` at the moment of the write.

Reproduced against the shipped function before changing it:

```
ALLOW .. /nvcc         -> /out/cuda/.. /nvcc
ALLOW ...              -> /out/cuda/...
ALLOW nvcc/.. /x       -> /out/cuda/nvcc/.. /x
ALLOW nvcc/NUL.txt     -> /out/cuda/nvcc/NUL.txt
ALLOW nvcc/COM1        -> /out/cuda/nvcc/COM1
BLOCK a./../b
```

So the guard stopped the obvious spelling and let the normalized one through — on the one platform whose extractor needs it, since this guard exists for the Windows `.exe` (7z) path.

## What changed

- **Any component with a trailing dot or space is rejected.** This closes the class without trying to reproduce Win32 normalization rules in PECOS: such names cannot be represented on Windows anyway, and do not occur in the archives this reads. Rejecting is strictly safer than modelling.
- **Windows reserved device names are rejected** (`CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9`), matched case-insensitively on the component's stem, because Windows resolves them as devices in *every* directory — `nvcc/NUL.txt` writes to the null device rather than creating a file. Names that merely contain those letters (`console.dll`, `nulled.txt`, `comment.h`, `auxiliary/x`) are still accepted, and a test pins that.
- **The doc comment now states that containment is lexical** and cannot stop a write travelling through a symlink or junction that already exists under the destination. That is a real residual limitation, previously implied to be covered by "route every entry name through this function".

## Verification

- Three new tests: the trailing dot/space class (six spellings including the `.. ` bypass and nested `nvcc/... /x`), the device-name class (six spellings, mixed case), and the must-still-work case for ordinary names containing device words.
- **Both new guards mutation-tested**: disabling the trailing-dot/space check fails its test; disabling the device-name check fails its test. Each restored byte-identical, confirmed by hash.
- `cargo test -p pecos-build`, `cargo clippy --locked -p pecos-build --all-targets -- -D warnings`, `cargo fmt --all --check`, and `just lint` all clean.

## Scope

Deliberately only this bypass, so it can land quickly and independently. Related findings from the same review are being handled separately: cuTensor downloads and extracts entirely unverified even though NVIDIA publishes its SHA256, and #479 has a cache-recovery regression of its own. Neither touches this file.


ciaranra:
Independent cross review found more, including one finding larger than the bug this PR opened with. All addressed in the latest push.

## The zip path had the same bypass, unbounded

The doc comment this PR inherited claimed `zip`'s `enclosed_name` already handled this class. **It does not.** Reading zip 8.6.0's `path.rs`: `ParentDir` matches only the exact bytes `".."`, so a component like `".. "` classifies as `Normal`, is pushed verbatim, and — because `Normal` *increments* the depth counter while `ParentDir` decrements it — there is no depth budget to exhaust. An entry named `.. /.. /.. /.. /Windows/System32/evil.dll` survives `enclosed_name` intact.

That claim was load-bearing in the worst way: it is the sentence that told an earlier reviewer this area was already covered. Zip entries now route through `contained_entry_path` as well, and the comment has been rewritten to say what is actually true of each reader:

- **tar** is safe by a *stronger* mechanism than lexical filtering — it canonicalizes each entry's parent and rejects anything landing outside the destination, so the OS collapses `.. ` before the comparison. It never needed this guard.
- **zip** balances `..` lexically and nothing else.
- **7z** validates nothing at all.

Practical exposure for the zip path is limited because that archive is SHA256-verified before extraction, so a hostile zip cannot be delivered without breaking the digest. It is a defence-in-depth gap rather than an open door — but the misleading comment was a real one.

## Data streams and device aliases

`:` is now rejected outright. It opens an NTFS alternate data stream (`nvcc.exe:payload` writes a hidden stream onto a legitimate binary, invisible to `dir` and to the installation check) or names a device (`nul:`). The device stem is now cut at the first `.` **or** `:`, with trailing spaces trimmed, so `NUL::$DATA` and `nul:` are caught where previously only `NUL.txt` was. `CONIN$` and `CONOUT$` were added to the reserved list.

## The test gap that mattered most

The review observed that **every test called `contained_entry_path` directly**, so replacing the call site with a bare `dest.join` left all nine passing. Confirmed by mutation: the guard could be silently disconnected and nothing failed. The guard's logic was tested; its *wiring* was not.

There are now three end-to-end tests that build a real 7z archive with a hostile entry name and drive `extract_windows_exe`, asserting both the rejection and that nothing was written beside the destination. Mutation-verified the right way round: bypassing the call site fails two of them, where it previously failed none. This required enabling sevenz-rust2's `compress` feature as a **dev-dependency only** — flagged explicitly since it touches dependency configuration.

## Kept unconditional, deliberately

The Windows-specific rules stay ungated rather than `cfg(windows)`. A gated guard is one CI never exercises — which is exactly how the first bypass survived — and an archive extracted on one platform may be consumed on another. False rejections are loud and name the offending entry; false accepts write files.

## Recommended follow-up, not done here

Both bypasses in this function came from the same root: it models what the filesystem *will do* instead of asking it. The durable fix is tar's approach — create the parent, canonicalize it, verify it is still under a canonicalized destination — as a second gate beneath the lexical one, which would close the symlink case this PR's doc comment now disclaims. Not done here because the only caller is the Windows-only 7z path that CI cannot exercise, and an untestable change to a security guard is a poor trade. It deserves its own change with a way to test it.

PR #495, opened 2026-08-13
Verify cuTensor archives and surface build-script install failures

ciaranra:
Closes the last two integrity gaps found by an independent adversarial review of #479. Both concern the same code path — the `build.rs` auto-install that an ordinary `cargo build` triggers — which is why they are together.

## 1. cuTensor was downloaded and extracted with no verification whatsoever

`ensure_cutensor()` sits three lines below `ensure_cuquantum()` in `crates/pecos-cuquantum/build.rs` and fetches a ~478 MB archive, pipes it into `tar -xf`, and loads the result at runtime. There was no digest anywhere in `cutensor.rs`.

NVIDIA publishes one. The manifest at `.../compute/cutensor/redist/redistrib_2.4.1.json` carries a SHA256 for every platform and CUDA variant, exactly as the cuQuantum manifest does. So #479 hardened one archive on that line and left its mandatory sibling — described in `cutensor.rs` as *"a required runtime dependency of cuTensorNet"* — completely open.

This now mirrors the pattern #479 established, deliberately rather than by coincidence:

- Four published digests recorded (linux-x86_64 and linux-sbsa, CUDA 12 and 13), with the manifest URL and refresh procedure in a doc comment, and a note that values must come from the manifest and never from a local download.
- Unsupported platform/CUDA combinations are rejected by `checksum_for` with the supported list in the message.
- Verification runs on **every** install, cached or fresh; a cached archive that fails is discarded and re-fetched once, so a corrupt file cannot wedge every later build.
- Downloads land on a `.part` sibling and are renamed only after the digest matches, so nothing unverified is ever published at the cache path.
- Hashing is incremental (64 KiB chunks) rather than reading half a gigabyte into memory on each install.

## 2. `build.rs` discarded install failures, including failed integrity checks

```rust
if let Ok(cuquantum_path) = pecos_build::cuquantum::ensure_cuquantum()
```

A `Sha256Mismatch` was dropped on the floor here. Worse, the installers' own progress output goes to cargo's captured stdout, which cargo hides unless `-vv` — so on the one path a normal `cargo build` uses, a tampered or truncated archive produced **no error and no visible message at all**. The build simply continued without the rpath, and the user met an unrelated-looking link or runtime failure later.

Both auto-install sites now report failures through `cargo:warning=`, the channel cargo always displays, and which this same file already uses for its ordinary progress messages. The build still proceeds — auto-install remains best-effort — but it can no longer fail silently.

## Verification

- The `linux-x86_64` / CUDA 12 digest was checked against the real artifact: 477,937,640 bytes downloaded, computing `032904fb8bba341e24aa45a8cc7b5afc63e4c28e22474530ccc97cfa546d0442`, matching both the manifest and the constant.
- Five unit tests: table lookup, rejection of unpublished combinations, a well-formedness guard (64-char lowercase hex) to catch transcription damage on future refreshes, a duplicate-entry guard, and a **known-answer** test pinning `file_sha256` and `verify_archive` against the published SHA256 of `"abc"` — the table tests alone would pass even with the hash function wired up wrongly.
- **Mutation-tested**: forcing the digest comparison to succeed fails the known-answer test; making the lookup ignore the CUDA major fails the unpublished-combination test. Restored byte-identical after each.
- `cargo test -p pecos-build`, `cargo clippy --locked -p pecos-build -p pecos-cuquantum --all-targets -- -D warnings`, `cargo fmt --all --check`, and `just lint` all clean.

## Remaining from that review, not addressed here

`extract_windows_exe` appears unable to produce a valid CUDA install — it writes entries at their archive-relative paths while the Linux path deliberately flattens component directories via `copy_cuda_components`, so validation looking for `dest/bin/nvcc.exe` would never find it. Pre-existing and unrelated to integrity, but it means that code path has likely never succeeded, and it is the only caller of the 7z containment guard. Worth confirming against a real installer before deciding whether to fix or remove.


ciaranra:
Independent cross review (two arms, different model families, blind to each other) — both confirmed the core security design and converged on the same top finding.

**Confirmed sound, by both arms and an independent manual check:** all four SHA256 digests match NVIDIA's live `redistrib_2.4.1.json`; the URL/filename each platform builds is the same artifact whose digest it verifies (no crossed wire, including `aarch64 -> linux-sbsa`); verification runs on both cached and fresh paths; and extraction is gated behind a passing verify, so a tampered archive is rejected and never reaches `tar`. The incremental hash was proven equivalent to whole-file hashing. No traversal, no path to an unverified extraction, no regression versus the previous behaviour.

**Fixed (both arms' #1 finding):** the `.part` download name was deterministic (`archive.tar.part`), shared across every process using the cache. Two `cargo build`s in two worktrees share `~/.pecos`, so build B could truncate build A's partial *between A's verify and A's rename* — meaning A would rename bytes it never verified into the cache and extract them, defeating the verify-before-extract guarantee. Security impact is bounded (B downloads the same legitimate URL, so injection would still require breaking TLS), but it is a real correctness hole. The partial name now carries `std::process::id()`, so each build's verify-then-rename operates on bytes no other build can touch.

**Also fixed:** the cached-fail `remove_file` now tolerates `NotFound`, so a concurrent build that already discarded the bad entry doesn't abort this one's refetch.

**Test honesty (both arms):** two tests overclaimed. The lookup test's literals are the same constants the table holds, so it cannot catch a wrong digit — only a wrong-row lookup; its comment now says exactly that, and it gained an assertion that the two CUDA-12 digests differ so a platform-dropping lookup is caught. The well-formedness test checks shape, not value; its comment no longer claims to guard "transcription damage." The known-answer test (sha256 of "abc") is the one that genuinely pins the hash wiring, and it stays. Validating the table against the manifest was done by hand (recorded in the table's doc comment with the refresh procedure); a hex-preserving typo in an unbumped digest has no automated net, which is now stated rather than implied.

**Same bug fixed in the sibling PR:** #479's cuQuantum installer had the identical deterministic `.part` pattern — and a comment explicitly claiming concurrent installs were safe. Fixed there too (per-process partial + honest comment), since fixing a pattern means fixing every instance.

Gates: `cargo test -p pecos-build`, clippy `-D warnings`, fmt, and `just lint` all clean on both worktrees. The duplication both arms noted (several near-identical streaming-hash/verify bodies) is real but left as a follow-up — a shared `verify_file_sha256(path, expected)` is worth adopting opportunistically, while a generic download/cache/retry helper would be premature given the installers' materially different flows.

Issue #481, opened 2026-08-11
pecos.random has no binomial, needed for bootstrap confidence intervals


ciaranra:
## Summary

`pecos.random` provides `random`, `randint`, `choice`, `seed`, and `RngPcg`, but no `binomial`.
The #458 migration hits this in `examples/surface/native_dem_threshold_sweep.py:428`, which
bootstraps confidence intervals for threshold fits:

```python
rng = np.random.default_rng(_stable_bootstrap_seed(ordered))
bootstrap_counts = rng.binomial(n=shot_counts, p=observed_rates,
                                size=(_FIT_BOOTSTRAP_SAMPLES, len(ordered)))
```

The call is vectorised over per-point `n` and `p` arrays. Bootstrapping logical-error-rate
confidence intervals is a normal thing to want in QEC analysis, so this belongs in the native
layer rather than being hand-rolled at a call site.

## Scope

`pecos.random.binomial(n, p, size=None)` matching NumPy's signature, accepting scalar or
array-like `n` and `p` with broadcasting, returning an `Array`. NumPy stays the test oracle for
distributional correctness (compare means/variances over many samples with a fixed seed, and
exercise the edge cases `p=0`, `p=1`, `n=0`).

Note the streams will not match NumPy's for the same seed -- PECOS's generator is its own. That
is expected and fine; consumers that need reproducibility pin PECOS's seed, and any test that
previously asserted exact NumPy-derived numbers has to be re-baselined against PECOS. See #458.

Blocks the `native_dem_threshold_sweep.py` portion of #458 step 3; the rest of that step is
unaffected.

PR #501, opened 2026-08-13
Migrate the threshold sweep off runtime NumPy (#458 step 3b)

ciaranra:
## Summary

#458 step 3b: `examples/surface/native_dem_threshold_sweep.py` off runtime NumPy -- 25 sites, now
zero. **`examples/surface/` has no allowlist entries left**, so the whole directory is NumPy-free
with the #460 ban enforcing it.

This file was blocked until `pecos.random.binomial` landed (#481/#487); its bootstrap uses the
vectorised `binomial(n, p, size=...)` form over per-point arrays.

## Two oracles, because one does not fit

**Deterministic paths kept exact equality**, captured before the change and asserted after:
stable bootstrap seed, fitted per-round rate, projected d-round rate, fit RMS, observed counts
and rates, round values, duration schedule, the native DEM smoke point, and both Wilson bound
vectors -- all bit-identical.

**Bootstrapped confidence intervals could not**, and that is by design: PECOS's RNG is its own
implementation and its stream deliberately differs from NumPy's for the same seed. Those
constants are re-baselined and labelled as such in the test, with the shifts being:

| CI value | NumPy | PECOS | shift |
|---|---:|---:|---:|
| per-round low | 0.004578116 | 0.004567746 | 0.23% |
| per-round high | 0.005504737 | 0.005495604 | 0.17% |
| projected low | 0.022475218 | 0.022425238 | 0.22% |
| projected high | 0.026924279 | 0.026880590 | 0.16% |

Sub-percent, consistent in direction, on 2000-resample statistics -- the signature of a different
stream rather than a defect. The decisive evidence is what did **not** move: point estimates and
Wilson bounds are exactly unchanged, so only bootstrap-derived values shifted.

Because a re-baselined constant and a regression produce the same red test, the intervals are
also pinned by stream-independent properties: each brackets its point estimate, each width sits
between 2x and 6x the propagated analytic binomial standard error, two pinned-seed runs are
exactly reproducible, and `_stable_bootstrap_seed` still yields the same seed for the same input
points.

## Verification

qec **1502** passed; rslib **1656**; docs **272** passed, 0 failed; the script runs end to end;
`ruff check` clean with the allowlist entry removed; pre-commit two passes exit 0; Rust untouched.

Remaining in #458: step 4, `decode.py` (113 sites), the last file.

### DIFF
--- a/docs/development/DEVELOPMENT.md
+++ b/docs/development/DEVELOPMENT.md
@@ -172,10 +172,8 @@ assert flat == [1, 0, 0, 1]
 - `fill()` does not coerce values by truthiness or parse numeric strings. Convert values
   explicitly instead of relying on behavior such as a non-empty string becoming `True`.
 
-Three more migration boundaries currently need local workarounds:
+Two more migration boundaries currently need local workarounds:
 
-- Advanced indexed assignment and boolean-mask reads are not available yet (#482). Use scalar
-  assignment or explicit selection locally, and collapse the workaround when indexed access lands.
 - Elementwise `Array ^ Array` is not available yet (#458). For arrays already known to contain only
   binary values, elementwise inequality has the same result; do not use that substitution for general integers.
 - Spell NumPy's `dtype=float` as `dtype=dtypes.float64`, especially with `asarray()`, to preserve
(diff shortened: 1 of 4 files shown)


ciaranra:
Closing evidence for the re-baseline, replacing reasoning with execution. The open question was
whether the CI shifts are stream noise or a hidden defect. The decisive test: hold the captured
points fixed, vary ONLY the bootstrap seed across 40 PECOS seeds, and ask whether the old NumPy
endpoints fall inside the resulting spread.

| endpoint | NumPy old | PECOS 40-seed range | inside |
|---|---:|---|---|
| per-round low | 0.004578116 | [0.004492311, 0.004671469] | yes |
| per-round high | 0.005504737 | [0.005406886, 0.005646427] | yes |
| projected low | 0.022475218 | [0.022061550, 0.022924951] | yes |
| projected high | 0.026924279 | [0.026456031, 0.027601650] | yes |

All four old values are ordinary members of the seed distribution -- the NumPy result is
statistically indistinguishable from one more PECOS seed. Combined with the deterministic half
being bit-identical (point estimates, Wilson bounds, seed, fit RMS), the re-baseline is proven
noise rather than asserted noise.

Issue #502, opened 2026-08-14
Array.__iter__ pays a per-element PyO3 round-trip, same pattern tolist had


ciaranra:
## Summary

`Array.__iter__` invokes `__getitem__` once per first-axis element (`pecos_array.rs:1226`) --
the same per-element PyO3 round-trip pattern that made `tolist()` ~843x slower than NumPy before
its single-pass rewrite. For 1-D arrays each step is a scalar round-trip; for N-D arrays each
step resolves and materializes a subarray selection (`:1609`).

Found during the `tolist` fix (reported there, deliberately not changed in that packet). Any
Python-level loop over an `Array` -- including comprehensions like the flatten workaround the
migration used before #476 -- pays this cost.

## Scope

Rework `__iter__` to iterate the underlying `ndarray` directly with per-dtype scalar conversion
inline, mirroring the `tolist` approach. 1-D yields scalars; N-D yields subarrays (decide whether
those can share one materialization path with slicing). Identity oracle: iteration order and
yielded values must equal `list(np.asarray(a))` semantics for every dtype and dimensionality.
Benchmark bar: within ~10x of NumPy iteration on the 8192x49 uint8 case, reported before/after.

Not urgent: `tolist` was the hot path in decode-batch handoffs and is fixed; this is the
remaining instance of the pattern.

PR #503, opened 2026-08-14
Make Array.tolist a single-pass native conversion (843x to 7.9x)

ciaranra:
## Summary

`Array.tolist()` recursed through Python-level `__getitem__` once per element -- a PyO3
round-trip and index-tuple construction for every scalar. Measured on an 8192x49 uint8 array:
**2.252 s vs NumPy's 0.003 s, ~843x.**

This surfaced as a hard blocker in #458 step 4: the migrated `decode.py` batch-decode path
regressed 7.69x, with 6.30 s of 7.57 s attributed to `tolist` alone. The step-4 packet's
performance gate STOPPed rather than ship it -- this PR is the layer fix that unblocks it.

## The fix

`tolist` is now a single-pass conversion in Rust: one match on the `ArrayData` variant, then
recursion by `outer_iter()` building nested `PyList`s with per-dtype scalar conversion inline.
No Python-level indexing, no per-element dispatch. All 15 dtype variants including Pauli and
PauliString; 0-d returns a scalar as NumPy does; empty and zero-axis shapes preserved.

**After: 0.0238 s on the same case -- 7.9x NumPy, from 843x** (independently re-measured on the
rebuilt tree; output verified identical to both the old implementation's captured output and
`np.asarray(a).tolist()`, including Python scalar types).

## Second layer fix in the same campaign: array_equal

The #458 step-4 review found a valid Tesseract configuration throwing
`TypeError: array_equal() arguments must be numpy arrays of bool, int, float, or complex` --
`pecos.array_equal` compared via a hand-enumerated dtype-pair table that predated the unsigned
work: native uint8 Arrays, NumPy uint8 arrays, and every mixed-dtype pair all raised.
Pre-existing on dev; decode.py only reaches it now that it stopped using `np.array_equal`.

Replaced with NumPy semantics through one path: shape check, then value comparison with dtype
promotion. Mixed uint8-vs-int64 and int-vs-whole-float compare True as NumPy does; shape
mismatch is False, not an exception; `equal_nan` matches NumPy in both settings; Pauli compares
against Pauli only, unequal (not an exception) against numerics. Verified against the NumPy
oracle per dtype, mixed pairs, and NaN cases; the motivating check-matrix case passes.

## Same pattern found elsewhere, deliberately not fixed here

`Array.__iter__` pays the identical per-element round-trip (`pecos_array.rs:1226`). Reported and
filed as #502 rather than folded in -- it is not on the hot path that blocked step 4.

## Verification

clippy `-D warnings` clean; fmt clean; pecos-rslib **1760** passed (tolist is used throughout the
gap-parity tests, so identity regressions surface loudly); qec **1502**; pre-commit two passes
exit 0. Mutation: swapping the nesting axis fails the identity tests.


ciaranra:
Closing evaluation on the one reserved aspect of this PR: `array_equal`'s mixed-dtype promotion
semantics, where a wrong `True`/`False` would be silent.

Two passes, both by execution:

**Targeted edges** (the known trap shapes): both directions of the uint64-max float-precision
trap, signed zero, complex-component NaN under both `equal_nan` settings, bool-vs-int
cross-pairs including truthy-but-unequal, `int64::MIN`, cross-width infinity, complex-vs-real.
All nine NumPy-identical -- including the case where NumPy's own answer is arguably surprising
(`uint64 MAX-1` vs the rounded float compares `True` under promotion), which PECOS mirrors
faithfully since NumPy semantics ARE the contract.

**Exhaustive sweep**: all 13x13 numeric dtype pairs x per-family value categories (zeros,
extremes at each width, NaN, +/-inf, -0.0, fractional, the float32 precision boundary
16777217, complex NaN/inf components) x both `equal_nan` settings --

```
swept 6050 cases across 13x13 dtype pairs
mismatches: 0   harness errors: 0
```

with `np.array_equal` computing every expected value. The enumerable surface is enumerated;
nothing left to an imagination-based review.

PR #526, opened 2026-08-16
Document native batch and parallel decoding in the Guppy DEM workflow

ciaranra:
## Summary

The Guppy DEM decoding workflow doc previously showed per-shot Python decode loops and mentioned `decode_count` in one sentence. This adds a section on the native `SampleBatch` decode paths:

- `decode_count`, `decode_count_parallel`, and `decode_count_batch`, with guidance on when each pays off.
- The configuration caveat that type strings construct decoders with default settings, including the `"pymatching"`-is-correlated vs `PyMatchingDecoder.from_dem`-is-uncorrelated distinction (the executable blocks use `"pymatching_uncorrelated"` so their equality assertions against the hand-built loop are exact).
- Wrapping simulated shots from `evaluate_result_columns()` in a `SampleBatch` so stage 4b shots ride the same native parallel path.
- A pointer to `decode_stats_parallel` for latency quantiles.

All new code blocks are tested: the doc-test suite (`just docs-test`) passes with the generated workflow test running all 11 blocks, and the equality assertions between the native paths and the hand-written loops execute as part of it.

## Verification

- `just docs-test`: 288 passed, 12 skipped (includes the 11 generated blocks for this doc).
- `just lint`: clean, no formatter rewrites.

### DIFF
--- a/docs/workflows/guppy-dem-decoding.md
+++ b/docs/workflows/guppy-dem-decoding.md
@@ -349,8 +349,54 @@ for syndrome, observable_mask in sim_shots:
 print(f"simulated shots, pymatching: {sim_errors}/{len(sim_shots)}")
 ```
 
-When you only need the count, `batch.decode_count(dem_text, "pymatching")` runs
-this same loop natively and returns the number of mismatches.
+The per-shot loops above are written for clarity, and they pay for it: every
+shot crosses the Python/Rust boundary once per decoder. When you only need the
+error count, `SampleBatch` runs the whole loop natively:
+
+- `decode_count(dem_text, decoder_type)` is the same loop with no per-shot
+  Python crossing.
+- `decode_count_parallel(dem_text, decoder_type, num_workers=...)` spreads the
+  shots across worker threads, each with its own decoder instance. Slow
+  decoders such as Tesseract and BP+OSD benefit most; for PyMatching at small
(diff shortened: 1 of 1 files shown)


ciaranra:
Superseded by #546 and #547, which rewrote this same section of `docs/workflows/guppy-dem-decoding.md` for the unified decode API.

This PR documented the legacy `decode_count` / `decode_count_parallel` / `decode_count_batch` menu. Those methods have since been removed: #546 migrated every consumer onto `SampleBatch.decode(...)` / `DemSampler.decode(...)`, and the follow-up deleted the eight legacy methods entirely. Merging this would conflict, and resolving it would reintroduce prose describing methods that no longer exist.

The content itself is not lost — dev now covers native, sequential, and parallel decoding through `workers=` and `execution_path`, plus a section on per-shot confidence with the experimental Frontier and BP-Trellis decoders. Closing rather than reworking, since nothing here is worth carrying forward separately.

ciaranra:
Correction to my previous comment: the follow-up that removed the legacy methods is on branch `unified-decode-p4b`, not #547 — #547 is unrelated work on stabilizer tensor networks. I cited the wrong number. The substance is unchanged: this PR is superseded by #546, and the legacy `decode_count` family it documents has since been deleted.

PR #545, opened 2026-08-17
feat(selene): add PECOS general noise error-model plugin

qciaran:
## Summary

- add a device-neutral Selene error-model plugin backed by PECOS GeneralNoiseModel
- expose immutable GeneralNoiseParameters with the full configurable Rust builder surface and noiseless defaults
- bridge scheduled Selene operations, idle timing, leakage, crosstalk, metrics, and state dumps into PECOS
- register the package in Cargo/uv workspaces and the cross-platform wheel matrix

## Validation

- 13 Python tests, including Stim and PECOS StateVec integration
- 2 Rust unit tests
- cargo clippy -D warnings
- cargo fmt and Ruff checks
- uv lock check
- wheel build with bundled native library


ciaranra:
## Review: multi-reviewer panel, findings verified by execution

Reviewed at head `a48162b62` with three independent reviewers (noise-model boundary fidelity, testing-claim verification, and a cross-model pass), each blind to the others. Every finding that drives a recommendation was re-verified locally; the commands and outputs are quoted. Recommendation: **REVISE**, which suits the draft status.

Scope note for anyone reading along: `crates/pecos-engines/src/noise/general.rs` is 4011 lines already on `dev`; this PR changes 25 lines there plus 148 in `builder.rs`. The new surface is the plugin — about 2200 lines of source against 3765 lines of tests.

### What is genuinely good

**The mutation audit holds up.** All ten defects in `tests/MUTATION_AUDIT.md` were re-applied at the current head and all ten are still rejected by the named tests, nine commits after the audit was written. No code had moved. The document is also honest about its own limits and records a mutation that initially survived along with the test added in response. That is unusual and worth saying plainly.

**The reference oracles are real.** `tests/qutrit_reference.py` is a NumPy three-level density-matrix model sharing no code with the Rust path, it is self-tested for trace/Hermiticity/positivity, and it demonstrably rejected four independently injected defects. `tests/basis_state_reference.py` is the same character. This is the right way to build an oracle.

Also verified clean: two-qubit operand order survives the round trip (`RZZGate{q1:3, q2:1}` with `pauli_model {"XI": 1.0}` puts the X on qubit 3); the nanosecond-to-second conversion uses `Duration::from_nanos(..).as_secs_f64()` with no integer truncation; emission replaces the original gate rather than adding to it; the leakage lifecycle is correct when `PZ` is not declared noiseless; global and local crosstalk payloads use opposite conventions and the plugin gets both right; and a default-constructed model is exactly noiseless.

### 1. The per-qubit clock is zeroed each shot, but the runtime clock is monotonic

`src/lib.rs:399` (`self.last_operation_end.fill(0);` in `shot_start`), consumed by `add_idle_before` at `src/lib.rs:341`.

Neither shipped runtime rewinds its clock. In the pinned Selene revision, `SimpleRuntime::shot_start` is `Ok(())` and `self.start` only ever advances (`+=` at lines 79 and 287); `soft_rz` is identical. So on shot N the first operation on each qubit arrives with `start` around `(N-1) * T_circuit` while `last_operation_end` has just been reset to zero, and the adapter fabricates an idle equal to the entire elapsed emulator time.

Executed, with `{"idle":{"linear_rate":1.0,"linear_model":{"X":1.0}}}`, shot 0 spanning `[0, 1e9)` ns and shot 1 starting at `1e9` ns (a physically zero gap):

```
[RXYGate { qubit_id: 0, theta: pi, phi: 0.0 }, Measure { qubit_id: 0, result_id: 0 }]
shot1 measure = true          // should be false
```

Two qualifications that matter. The spurious idle lands immediately before the qubit's first operation of the shot, so it is overwritten when that operation is a `Reset` — the usual Guppy `qubit()` pattern, which is why this does not show up in ordinary use. It is not overwritten when the first operation is a gate or a measurement, and `qalloc` emits no reset. Separately, the leak branch draws an RNG value either way, so the random stream for the whole shot depends on the shot index.

### 2. Measurement-crosstalk victims accumulate across shots

`src/lib.rs:405` (`shot_end` calls `model.reset()`) against `general.rs:1406`, where `reset_noise_model` clears `leaked_qubits` and `measured_qubits` but not `prepared_qubits`.

`prepared_qubits` is the global measurement-crosstalk victim pool. It is documented as "qubits prepared at any point in the program", but the plugin holds one model instance for the whole run, so "the program" becomes "every shot so far". Executed with `global_crosstalk_probability = 1.0`:

```
shot 0: Reset q1; Measure q0   -> q1 is a victim (correct)
shot 1: Measure q0 only        -> ops = [Measure q1, Measure q0, RXY(pi) q1]
```

Qubit 1 was collapsed and bit-flipped in a shot where it was never prepared. Shots are therefore not i.i.d.: shot 0 sees systematically less crosstalk than every later shot, so a logical error rate estimated over S shots is biased and depends on shot ordering. Two of the three reviewers found this independently.

Suggested fix for 1 and 2 together: reset all physics-affecting state in `shot_start` rather than `shot_end`, including `prepared_qubits`, and do not zero the per-qubit clock — or rebase it on the shot's start timestamp.

### 3. A noiseless reset leaves leakage state stale

`with_noiseless_gate` (`src/lib.rs:121`, `plugin.py:381`) bypasses the fault path, but the bridge still sends `PZ` to the simulator as `Reset` (`src/simulator.rs:148`). After a leak followed by a noiseless reset, the simulator is at `|0>` while the model still records the qubit as leaked, so a later `MeasureLeaked` disagrees with the physical state. Either run the leakage bookkeeping independently of the noiseless-fault bypass, or reject `PZ` as a noiseless-gate selection.

### 4. Two plausible defects survive the entire suite

Both were applied and run against 8 Rust tests, the 98-case fast lane, and the 776-case slow lane — all green in every configuration:

- `apply_prep_faults` no longer calls `mark_as_unleaked`, so a leaked qubit stays leaked forever through a reset. No test resets a leaked qubit mid-circuit; leakage is cleared per shot by `reset_noise_model()`, so the documented "leaked until a Prep" semantics is never exercised.
- Measurement-crosstalk victim selection, when the same defect is applied to production and to the in-file reference. See the next section for why that is possible.

### 5. `reference_message()` is a copy of the code it checks

`src/lib.rs:750`. It is a near-verbatim transcription of production `handle_operations`: the same `Duration::from_nanos(..).as_secs_f64()` idle computation, the same `!measured.contains(qubit)` victim filter, the same `rzz(theta, &[(first, second)])`.

It is the sole guard for three behaviours: sub-second nanosecond-to-second idle conversion, local crosstalk victim selection, and `MeasureLeaked` triggering the crosstalk payload. Putting one mistake on both sides leaves 8 Rust tests and 880 Python tests green, which is finding 4's second entry.

It also diverges from production in a way nothing tests: production emits the crosstalk payload inside the operation loop immediately before the first measurement, while the reference emits it before all operations. That only shows up in a batch mixing gates and measurements, which the randomized trace never generates — it emits one operation kind per batch.

This is worth contrasting with `qutrit_reference.py` in the same PR, which is a genuine independent oracle and does catch real defects. The distance between oracle and implementation is the whole point; a differential test against a transcription of the implementation cannot provide it.

### 6. The strongest tests are the least enforced

- 1014 of 3065 test lines sit in files carrying a file-level `pytestmark = pytest.mark.slow`: `test_generated_conformance.py`, `test_idle_matrix.py`, `test_layered_matrix.py`, and `test_coverage_contract.py`.
- The PR lane runs `-m "not slow"`; the `-m slow` step is gated on `github.event_name != 'pull_request'` (`selene-plugins.yml:145`). On this PR, 781 of 879 Python cases were deselected.
- `selene-general-noise-semantics.yml` is `schedule` plus `workflow_dispatch` only, so no PR triggers it. More importantly, its only test step pipes through `tee` (`:75`) with no `pipefail` and no `shell:` override, so under GitHub's default `bash -e` the step takes `tee`'s exit status. Verified: `bash -e -c 'false | tee /dev/null'` exits 0. **That workflow cannot fail on a test failure.**
- `test_coverage_contract.py` is itself slow-marked (`:12`), and it enforces nothing about behaviour. It reads `@pytest.mark.noise_channel` metadata gathered at collection, before deselection. Verified: `pytest tests -m slow -k coverage_contract` reports `1 passed, 880 deselected` — full channel coverage certified with zero contact with the plugin. In `test_generated_conformance.py` the marks are attached during `_sensitive_cases()`, a pure oracle-side computation, so the coverage numbers certify that the reference model predicts a distinguishable distribution, not that the plugin was run or agreed.
- Tolerance calibration: base rates scaled by 1.35 are caught (5 fast-lane failures), 1.15 survives the fast lane and is caught only by the extended matrix, and 1.08 survives both. Several "certain outcome" cases run at 64 shots with a single category, where the Hoeffding tolerance is 0.337 — a third of shots could come out wrong and still pass. One runs at 32 shots, tolerance 0.476.

Note how this interacts with the findings above: none of 1 through 4 is a test that failed to run. Each is a behaviour with no test at all. Every Rust test calls `shot_start(0, ..)`, and `test_idle_matrix.py` deliberately drives idle through `with_idle_after_2q` "without assuming a device schedule", so `add_idle_before` has no multi-shot coverage anywhere in the PR.

### Smaller items

- `src/simulator.rs:52` maps `GateType::H` to `RXY(pi/2, -pi/2)`, which is `RY(-pi/2)`. Against PECOS's standard Hadamard that differs by `diag(1,-1) = Z`, not by a global phase; `H` is not expressible as a single `RXY` and needs an accompanying `RZ(pi)`. Currently unreachable — no emitted or noise-injected gate produces `H` — but it sits in a table that reads as verified, and the first channel that emits `H` will silently produce a wrong Clifford. The other dead arms are correct up to global phase.
- Angles are round-tripped through `Angle64` rather than carried as the source `f64`, so the simulator executes a quantized version of the requested circuit. Absolute resolution is about 3.4e-19 rad and physically negligible, but the relative error grows for small rotations, and it cuts against the convention established by the controlled-rotation work of lowering from the source `f64` at every ingress. Separately, a non-finite angle panics in `Angle64` conversion; `handle_operations` is not wrapped in `catch_unwind` and neither is the plugin export path, so `rz(NaN)` unwinds out of an `extern "C"` function and aborts the emulator.
- `builder.rs:1225` extends `with_scale`/`with_idle_scale` to the sin-squared and coherent idle rates, where previously only the linear rate was scaled. This is almost certainly the intended correction, but it silently changes the noise magnitude for any existing consumer combining those setters, and it is a `pecos-engines` behaviour change arriving inside a plugin PR. It belongs in release notes.
- `two_qubit.idle_after_gate` has no documented unit — the docstring is "Set the idle duration applied after two-qubit gates" while idle rates elsewhere are in seconds, so a user passing nanoseconds gets a 1e9 error that no validator catches. It is also additional to the schedule-derived idle, so raising the runtime's `duration_ns_rzz` double-counts. Measurement, normally the longest operation on a trapped-ion device, has no analogous knob.

### Cheapest things that would move this forward

1. One multi-shot Rust test: `shot_start(0)`, run a circuit, `shot_start(1)`, run the same circuit, assert identical output. That kills findings 1 and 2 together and is a few lines.
2. Add `shell: bash` with `set -o pipefail`, or drop the `tee`, in `selene-general-noise-semantics.yml`. Right now that lane cannot report a failure.
3. Base the coverage contract on passed test reports rather than collection-time metadata, and run it in the PR lane.
4. Replace `reference_message()` with something structurally distant, or state in the file that it is a transcription and therefore cannot detect shared misconceptions.

### Findings that did not survive cross-checking

Reported for completeness, since both looked serious at first:

- "The branch does not compile." True at `1c3b6625a`, where the `dev` merge brought the `RXY1Q` rename without updating the plugin. Fixed nine minutes later by `a48162b62`, which is the actual head. `cargo check -p pecos-selene-general-noise --all-targets` exits 0 there.
- "The plugin's Rust unit tests do not run on PRs." True of `rust-test.yml`, whose PR path is `pecos rust test --help`, but `pr-core-rust` runs `just rstest debug` which is `cargo test --workspace`, and the plugin is on neither exclude list. The 8 Rust tests do run; verified locally, 8 passed.
- "The diff carries unrelated qudit-simulator work." An artifact of a stale local `dev` ref. `aeb193def` is an ancestor of both this branch and `origin/dev`.

### One caveat on this review

Every reference model used here was written by a reviewer or by the PR itself. Nothing checks this plugin against an external noise-model implementation, so a shared misconception about Selene's semantics — particularly around shot boundaries, which is where findings 1 and 2 live — would not be caught by any of it.

PR #547, opened 2026-08-17
Repair disentangling proofs and fix gauge-sensitive projection truncation (fixes #557)

### LINKED ISSUE #557: STN prob_bitstring disagrees with the simulator's own state for Clifford+T circuits
`StabMps::prob_bitstring` returns probabilities that disagree with the state the simulator actually holds, by up to 6e-3, for Clifford+T circuits. `test_prob_bitstring_honest_clifford_t_family_matches_dense_state_vector` catches this and has been failing on the macOS and Windows `rust-test` lanes on `dev` since #523 (be5a2ead4), the commit that added the test. Those lanes only run post-merge, so no PR has ever seen it.

## The fault is in `prob_bitstring`, not in the state

On a failing circuit (`n=6`, `t=4`, `circuit_seed=22604`), comparing all three readouts per outcome:

```
outcome=  0  dense=0.0312500000000000  amplitude^2=0.0312500000000000  prob_bitstring=0.0276291969490717
outcome=  4  dense=0.0312500000000000  amplitude^2=0.0312500000000000  prob_bitstring=0.0263537005417451

worst |amplitude^2 - dense|     = 0.000e0
worst |prob_bitstring - dense|  = 4.972e-3
```

`amplitude()` agrees with the dense state vector exactly, so the MPS + tableau state and the dense contraction are both correct. Both distributions sum to 1.0, so this is not a normalization slip: `prob_bitstring` redistributes mass between outcomes.

That isolates the defect to the only path `prob_bitstring` uses and `amplitude()` does not — the iterative forced projection of Liu-Clark 2412.17209 Algorithm 3, `measure::project_forced_z`, whose conditional probabilities are multiplied together in `exp/pecos-stab-tn/src/stab_mps.rs:1011`. The test's own comment names the suspected mechanism: sequential forced-projection frame errors that a diagonal-only circuit tail would hide.

## It is not platform-specific

The failure was first seen only on macOS and Windows, which suggested a platform floating-point issue. It is not. Sweeping the same honest family over 640 seeded circuits on Linux:

```
68 mismatches, worst delta 6.336e-3     (test tolerance is 1e-12)
n=4 t=6 seed=21406 outcome=0: prob=0.0624935946082910 dense=0.0625000000000000 delta=6.405e-6
n=6 t=4 seed=22604 outcome=0: prob=0.0276291969490717 dense=0.0312500000000000 delta=3.621e-3
```

The committed test covers 16 circuits, and its `n=5 t=6 seed=10506` case happens to come out at `4.163e-16` on Linux while landing on the wrong side of the defect on macOS and Windows. The platform split is which circuits trip the bug, not whether the bug exists.

## Reproducer

Sweep `circuit_seed` in `test_prob_bitstring_honest_clifford_t_family_matches_dense_state_vector` (same builder settings: `merge_rz(false)`, `max_truncation_error(0.0)`, `flush()` before reading). `n=6 t=4 seed=22604` fails in well under a second in release mode.

## Scope

`prob_bitstring` is exposed to Python through `pecos-rslib-exp` (`python/pecos-rslib-exp/src/stab_mps_bindings.rs:643`), which is not among the published distributions (`pecos-rslib`, `pecos-rslib-llvm`, `quantum-pecos`), so it currently reaches source builds only.

## Acceptance criteria

- `prob_bitstring` matches the dense state vector to 1e-12 across the honest Clifford+T family for a wide seed sweep, not just the 16 committed circuits.
- The `#[ignore]` added to `test_prob_bitstring_honest_clifford_t_family_matches_dense_state_vector` to unblock the release lane is removed, and the seed coverage is widened so a passing seed cannot mask the defect again.

ciaranra:
Closes #557

Fixes silent state corruption on `dev` when a measurement or projection is followed by a
non-Clifford rotation that takes the exact-disentangling fast path.

## The defect

The exact-disentangling fast path in `non_clifford.rs` treats a `Some(SiteEigenstate::Z(false))`
entry in `disent_flags` as a proof that the coefficient-MPS site is `|0>`, and absorbs the
rotation into the tableau on that basis. After a projection, however:

- the site that actually collapses is the projection's flip site, which is **not in general the
  measured physical qubit index**, and
- the compensated CNOTs applied during phase-complete pre-reduction can modify **other** virtual
  sites,

while the callers marked `disent_flags[physical qubit] = Z(outcome)` and left every other flag
untouched. `Mast::project_all` repaired no flags at all. A later `rz` could therefore select the
fast path against a stale proof and silently corrupt the state.

Reproduced against `dev` (`be5a2ead4`) with an independent dense-state-vector oracle:

| path | fidelity vs oracle |
|---|---|
| StabMps eager measurement | 0.9335 |
| StabMps lazy measurement | 0.8902 (max bitstring-probability error 5.8e-2) |
| MAST `project_all` | 0.8278 (max probability error 1.4e-1) |
| MAST exact `mz` | 0.9335 |

Two of the four are phase-sensitive: probabilities stay correct while the state is wrong, so
probability-only tests cannot see them. Behavior is identical with numerical flag redetection on
or off. Independently re-confirmed through the Python bindings on a hand-written oracle: the same
5-qubit circuit gives fidelity 0.9662 before this change and 1.000000000000 after.

## The fix

- The projection helpers now report the virtual site they actually collapsed and every site the
  compensation touched (`measure.rs`).
- One shared repair marks only the collapsed virtual site `Z(false)` and conservatively clears the
  touched proofs (`stab_mps.rs`), applied to eager and lazy `StabMps` measurement, `reset_qubit`,
  `Mast::project_all`, and MAST exact data measurement.
- Existing public return types are preserved through wrappers.

`prob_bitstring` and `amplitude_iterative` operate on cloned state and never mutated live flags,
so they were unaffected.

## Verification

- New dense-oracle continuation regressions covering eager/lazy x redetection-on/off (fidelity
  1.0000000000000004, max bitstring error 5.6e-17) and the MAST paths (release-gated for its dense
  eight-site reconstruction).
- `cargo test -p pecos-stab-tn`: 267 unit, 91 verification, 3 doc tests pass; default lane ~74s.
- `cargo clippy -p pecos-stab-tn --all-targets -- -D warnings`: clean.

## Blast radius

The wrong-flag assumption is **original to the crate**, not a regression from #523: both
assignments are present in `0ab008fa8` (PR #295, which introduced `pecos-stab-tn`) and unchanged
through `be5a2ead4^`. #523 did not introduce it — it enlarged the exposure by adding exact
projection routes. Release notes should scope this to all released versions of `pecos-stab-tn`.

Mitigating: the measurement path wrote `Z(result.outcome)`, and only `Z(false)` is matched by the
consumer, so just the `outcome == 0` case planted a live false proof; pure-Clifford and
trivial-MPS workloads keep valid flags throughout.

## Performance impact (measured, not a defect)

Repairing the proofs removes fast-path firings that were previously taken against invalid proofs,
so the exact-disentangling fast path fires less often. Measured across two independent probes on
identical seeds:

- STN `multi_disent` 45 -> 38, fast-path rate 2.821% -> 2.369%; MAST post-projection
  `multi_disent` 7 -> 5 (committed `examples/disent_measurement_probe.rs`, trajectory hash
  identical throughout, so no behavioral drift)
- an earlier probe over a different workload mix measured a much larger swing; the figures above
  are the ones the committed probe reproduces and are the ones to trust

The stock `examples/disent_firing_rate.rs` takes no measurements in its non-MAST scenarios and
cannot observe this; a measurement-bearing probe is required to reproduce.

Follow-up (not in this PR): `is_numerical_product_zero_site` requires `bond_dim == 1` on both
bonds, while the sufficient condition is a zero sigma=1 block at any bond dimension. Generalizing
redetection to a canonical-form marginal test would soundly recover much of the lost firing rate.

---

## Added since the original description

This PR grew while chasing the same machinery. It now also contains:

**Fixes #557 — gauge-sensitive SVD truncation in exact forced-projection compensation.**
`prob_bitstring` disagreed with the simulator's own state by up to 6.3e-3 on Clifford+T
circuits while `amplitude()` matched dense exactly. Root cause: a forced projection leaves the
coefficient MPS in an arbitrary gauge, and the SWAP-chain long-range CNOT split each adjacent
pair with a local SVD whose singular values are not physical Schmidt weights in that gauge, so
the cutoff discarded significant weight and could even raise the contracted norm through lost
interference. The compensation path now maintains mixed-canonical form around the active bond
(left-absorb inward SWAPs, right-absorb the CNOT and return SWAPs) so every SVD happens at the
orthogonality center. Measured identically on merged dev and on this branch pre-fix (4/640
circuits, 68 outcomes, worst 6.336e-3), so it was a distinct defect from the flag repair.
After: worst 1.193e-15 (`prob_bitstring`) and 1.305e-15 (`|amplitude_iterative|^2`) over a
640-circuit sweep, now a permanent release-gated test beside the fast 16-circuit lane.

Note for merge ordering: #558 quarantines the honest-family test that this fixes. Whichever
lands second should drop the `#[ignore]`.

**Verified proof installation, replacing conservative clearing.** Proofs are installed by
testing the real invariant — the stored-frame marginal `<psi|P_1|psi>/<psi|psi>`, valid at any
bond dimension — instead of the previous `bond_dim == 1` product-site proxy. On the committed
probe this recovers 6 of the 7 lost STN fast-path firings (85.7%; `multi_disent` 38 -> 44
against a pre-PR baseline of 45), with identical trajectory hashes. The MAST leg's two lost
firings are not recovered by it.

**Batched environment verification.** Candidate sites are checked against left/right
environments computed once per pass rather than re-contracted per candidate: fallback kernel
11.53 us -> 2.34 us per candidate (4.93x), with the per-candidate implementation retained as a
test-only reference and equivalence asserted to 1e-12. Combined with the canonical-form change,
the MAST projection-heavy workload moved from +26.7% versus merged dev to **-39.4%**.

**Also fixed here (found by adversarial review of the first commit):** the lazy path installed a
proof valid only in the effective `V * stored` frame while flags are stored-frame; and
`flush_pauli_frame_to_state` applied X to MPS sites without clearing the corresponding proofs.

## SVD stability hardening (final three commits)

Adversarial review of the branch surfaced latent SVD-robustness defects at 8-qubit MPS shapes; the
last three commits fix them and were each re-reviewed:

- **Phase-aligned QR repair of fallback factors** (`0e6fa5a81`). Factors derived as `A v / s` in the
  Gram and realified SVD fallbacks accumulate condition-dependent orthogonality error; a
  phase-aligned thin QR restores orthonormality (isometry defects ~1e-2 -> ~1e-15 on graded and
  clustered spectra, verified against the nalgebra oracle). Also restores the load-bearing
  expectation-endpoint snap with a regression, returns exact `0+0i` for structurally zero iterative
  amplitudes, and raises the fuzz lane to 8 qubits. A simplification experiment confirmed both
  fallbacks are needed: removing the Gram path surfaces 72 errors over a 1,200-circuit census.
- **Loud failure over silent continuation** (`8e923a0a2`). An intermediate design stashed SVD
  failures and continued executing; review showed reads then returned unit-norm states up to 35%
  wrong with nothing consulting the stashed error. That machinery is removed: an unrecoverable
  numerical state panics with the failing operation named. Also fixes `Mast::mz` returning fewer
  results than qubits on one path, stops a speculative disentangler candidate failure from aborting
  the sweep, and documents the empirical SVD tolerances.
- **Dimension-scaled fallback tolerance** (`ae4928f18`). The fallback acceptance gate used a
  dimension-free max-element bound while its independent validator used the Frobenius-norm,
  dimension-scaled backward-error scale; accurate factorizations on ordinary 8-qubit states could
  pass validation yet be rejected, panicking at census seeds 766 and 1698 (~1 in 900 circuits at
  8q/80 gates). All fallback tolerances now share one helper computing
  `||A||_F * 512 * max(nrows, ncols) * eps`. This also rescales `fallback_zero_threshold`, the
  deflation threshold passed to the realified fallback, making the code match its own documented
  claim of using the validator's scale. Both seeds are default-lane regressions (verified to fail
  on the parent commit), and `--stability-census-wide` extends the 8-qubit lane through seed 1999.

Final review verdict: mergeable. Independent re-verification measured 0 panics over the shipped
1,200-circuit census and the 4,800-circuit wide census; a 2,224-matrix ground-truth battery over
graded/clustered/rank-deficient spectra with forced truncation found no acceptance-gate error
(worst reconstruction residual 4.0e-14 relative); the widened acceptance window admitted exactly
the two previously panicking factorizations out of 17,779 fallback acceptances; and an
end-to-end dense-oracle replay over 253 eight-qubit circuits shows worst probability disagreement
6.7e-6, attributable to default truncation.

Out of scope, filed separately: #562 (`amplitude_iterative` loses the per-bitstring Clifford frame
phase; predates this branch; magnitudes and all probability paths unaffected).


ciaranra:
Marking this NOT ready to merge pending fixes from an adversarial review of the delta since the first commit. Two findings are substantive and both were reproduced by execution:

**The #557 root cause is only fixed on one path.** `apply_long_range_two_site_gate` still splits each SWAP/gate pair with a local SVD in whatever gauge it receives, and it is used by the lazy/deferred measurement path and by the multi-site RZ std cascade — which is the dominant branch (multi_std 1555 vs multi_disent 44 on the committed probe). A gauge-invariance probe presenting the same physical state in two gauge-equivalent forms gives worst pair fidelity 1.000000000000 on the repaired exact route but 0.3139 on the legacy route, and 0.2195 for `apply_long_range_two_site_gate` directly. The doc comment at mps.rs:495-497 also still asserts the belief #557 falsified.

**`Mps::compress()` truncates off the orthogonality center.** The right-to-left sweep writes S*Vt back to site q and pushes bare U left, so singular values stay at q and site q never becomes right-canonical; from the second bond onward every truncation is measured against a non-identity right environment. Over 40 seeds a textbook canonical sweep beat it 40/40, worst fidelity 0.7666 vs 0.8589. This is on the exact-projection path and the hot RZ paths, and it also means `truncation_error`/`bond_cap_hits` currently report discarded weight in the wrong metric.

Also fixing: the proof-install tolerance accepts a probability marginal of 1e-12 where the test it generalizes used an amplitude-relative 1e-12 (probability 1e-24), with 99 installs measured within 1.2x of the cutoff and the soundness assertion set looser (1e-10) than the installer so it cannot constrain it; and an unreproducible `lazy && seed == 5` test carve-out that passes with the skip removed.

Corrected the performance figures in the description — the earlier "43.3% -> 25.2%" and "~91%" came from a different probe than the one committed here. The committed probe shows STN multi_disent 45 -> 38 -> 44 (85.7% recovery) with an unchanged trajectory hash.

Review also confirmed the parts that are sound: canonical form genuinely holds on the new exact path for adjacent, distant, boundary and reversed-order cases (worst amplitude error 2.74e-15 vs dense over all 20 ordered pairs); the cached-environment marginal is correct and norm-independent, and cache staleness is structurally impossible via borrowck; the 640-circuit sweep is a genuine falsifier that fails when the fix is reverted.

PR #550, opened 2026-08-17
Compute Pauli product quarter-phase with word-parallel popcounts

ciaranra:
## What

`PauliBitmaskGeneric::multiply_with_phase` computed the quarter-phase of a Pauli product with a
per-qubit loop over a 4x4 lookup table, iterating once per qubit up to the highest set bit. It now
computes the same exponent with six masked popcounts per storage word.

The phase of a symplectic Pauli product decomposes into six mutually exclusive per-site cases
(XY, YX, XZ, ZX, YZ, ZY), each contributing `i^1` or `i^3`. Each case is a boolean function of the
four X/Z plane bits, so a whole 64- or 128-bit word is handled by six mask-and-popcount operations
instead of up to 64 or 128 bit extractions.

The computation lives behind a new required `BitmaskStorage::mul_phase_exponent`, implemented for
all three backends (`u128`, `Vec<u64>`, `SmallVec<[u64; 8]>`). The two word-based backends share
one `phase_terms_u64` helper so the derived algebra exists in a single place. Their loop is bounded
by the shorter operand's plane length, since a term can only fire where both operands have stored
bits; this also makes the common empty-times-long multiplication O(1).

The public signature and semantics of `multiply_with_phase` are unchanged: it still returns
`(product, k)` with the true product equal to `i^k` times the phaseless symplectic product,
`k` in `0..4`.

## Breaking change

`BitmaskStorage` is publicly re-exported, and `mul_phase_exponent` is a required method with no
default body, so any out-of-tree backend implementation must add it. This is intentional. A default
per-bit implementation would silently hand external backends the slow path this PR exists to
remove; a compile error is the honest signal. There are no implementors outside this module in the
workspace, and every other method on the trait is likewise required.

## Correctness

The previous per-qubit implementation is preserved verbatim in the test module as
`reference_phase_exponent` and is the oracle for the new tests:

- exhaustive 16-pair single-qubit table, asserted against hardcoded expected exponents, run on all
  three backends;
- seeded randomized comparison against the oracle across all three backends, over qubit counts
  spanning word boundaries (63/64/65, 127/128/129, 191/192/193, 511/512/520) and mixed dense/sparse
  operators;
- deterministic fixtures where the four X/Z planes hold pairwise-distinct stored lengths, in both
  operand orders, including support beyond the inline `SmallVec` capacity;
- a debug assertion in `multiply_with_phase` that the phase parity equals the symplectic product,
  cross-checking the new formula against the independently written `commutes_with` on every
  debug-mode call.

Mutation-tested: flipping any coefficient or truncating the word loop in any backend fails the
suite.

## Performance

Measured with the benchmark module added in #553, same harness and byte-identical operands on both
branches, criterion baseline comparison. Selected results:

| shape | backend | change vs dev |
|---|---|---|
| dense 128q | u128 | -96.2% |
| dense 1024q | `Vec<u64>` | -94.1% |
| dense 1024q | `SmallVec` | -92.8% |
| sparse low-index 1024q | `Vec<u64>` | -27.1% |
| sparse high-index 1024q | `Vec<u64>` | -97.9% |
| identity times dense 1024q | `Vec<u64>` | -97.2% |

Phase-free `multiply`, whose code this PR does not touch, moved by under 1.4% in the same runs,
which sets the noise floor for these numbers.

The first benchmark run exposed a genuine regression rather than confirming a win everywhere:
sparse low-index operands on a wide register were 87% SLOWER, because the previous implementation
stopped at the highest set bit while the word-parallel one walked every stored word, most of them
all-identity padding. The follow-up commit skips words where either operand is entirely identity,
which turns that case into a 27% improvement while leaving the dense cases unchanged. Every
benchmarked shape is now faster than dev.

Callers of this method are the fault-analysis and Heisenberg-propagation paths in `exp/pecos-eeg`;
their suites pass unchanged.


ciaranra:
Post-panel review of the skip-guard commit (82c78fca9), which landed after the original three-arm review: two fresh independent arms both confirmed the skip is exact — every phase term is a bitwise subset of (lx|lz) & (rx|rz), so all-identity words provably cannot contribute — and one arm killed all six semantically wrong guard mutations against the suite (three survivors are strictly weaker but still exact skips). Follow-up commit addresses the three non-blocking findings: the distinct-plane-length fixture now expects a non-zero phase (its assertion was vacuous at 0), the shared word bound is named common_len instead of the misleading max_len, and a new fixture puts the only contributing overlap in word 10 so an early-stopping word walk fails a deterministic test (both previously-surviving cap mutations now die).

PR #552, opened 2026-08-18
Add operator product, commutator and anticommutator to PauliSum

ciaranra:
## What

`PauliSum` is the sparse complex sum of Pauli operators used to express channels and Hamiltonian-like operators. It supported addition and scalar multiplication but had no operator product, so expressing `A*B`, `[A,B]` or `{A,B}` meant dropping to dense matrices.

This adds:

- `try_mul` — the operator product, expanding term by term. Each pair multiplies its Pauli keys through `PauliBitmaskGeneric::multiply_with_phase`, folds the resulting `i^k` into the coefficient product, and accumulates through the existing `add_term`, which already merges equal keys and drops near-zero coefficients.
- `impl Mul for PauliSum`, delegating to `try_mul` and panicking on qubit-count mismatch, mirroring how `Add` delegates to `try_add`.
- `commutator` (`A*B - B*A`) and `anticommutator` (`A*B + B*A`), built from the product and the existing addition and scalar-multiplication paths.

Cancellation to the empty sum is a correct result, not an error: the commutator of two commuting sums is legitimately zero.

## Tests

Single-term algebra against hand-computed values (`XY = iZ`, `YX = -iZ`, `ZX = iY`, `XX = I`); complex coefficient propagation; multi-term expansion where distinct term pairs merge onto one output Pauli, and where they cancel exactly and the term is dropped; `[X,Y] = 2iZ`, `[X,X] = 0`, and a two-qubit commuting pair; `{X,X} = 2I` and `{X,Y} = 0`; qubit-count mismatch from all three entry points plus the panicking operator; identity is multiplicative identity in both orders; associativity on three multi-term operands.

Mutation-tested: flipping the sign of the `i` phase or of the commutator's negation fails the suite.


ciaranra:
Post-review fix, from an independent fresh-context review of this PR's diff (verdict REVISE, now addressed):

The review found one real defect: try_mul accumulated through add_term, whose zero-dropping tolerance tests each pair contribution in isolation — so contributions that should sum to a significant coefficient were silently discarded before they could accumulate. Demonstrated concretely: squaring a 200-term sum of 1e-6 X coefficients has an exact identity coefficient of 2e-10 (two hundred times the drop tolerance) and returned the empty sum. The lost mass grows with the product of the term counts. The fix accumulates raw into a local map and applies the tolerance once per final coefficient via the existing try_new, which is sum-then-drop; commutator and anticommutator inherit the correction unchanged. A regression test pins the 200-term case.

Also from the review: the exponent-2 entry of the phase table had no test able to detect a sign flip (the only even-weight product in the suite asserted emptiness, where the error cancels), and the Mul operator's operand order was untested (the associativity check is invariant under a global swap). Both now have dedicated tests, and all three mutations were verified to fail the suite against the committed fix. Doc comments on try_mul, commutator, and anticommutator now state the tolerance semantics and the term-count cost bound, and the associativity test is renamed to scope its claim to well-scaled coefficients.

PR #554, opened 2026-08-18
Add Pauli action on computational basis states to PauliBitmask

ciaranra:
Rebased onto dev after #550 merged; the diff is one method plus tests.

## What

A Pauli operator maps a computational basis state to exactly one other basis state, up to a quarter phase. `PauliBitmaskGeneric` could multiply Paulis but had no way to apply one to a basis state, so callers wanting `P|b>` had to loop qubit by qubit themselves.

`apply_to_basis_state(&self, state: &B) -> (B, u8)` returns the image basis state and the exponent `k` with `P|state> = i^k |image>`. The image is `state XOR x_bits`; the phase comes from three per-word masked popcounts, mirroring how the product phase is computed:

- `Y|0> = i|1>` contributes 1, `Y|1> = -i|0>` contributes 3, `Z|1> = -|1>` contributes 2;
- `I`, `X`, and `Z|0>` contribute nothing.

For the growable backends the walk is bounded by the Z plane's length narrowed to where an X bit or a set state bit can exist, since every term needs a Z bit and either an X bit or a state bit. Zero-padding inside that bound is meaningful rather than merely harmless: an absent state word means those qubits are in state zero, and an absent X word means Z rather than Y.

Like the product phase, this is a required method on the publicly re-exported `BitmaskStorage` trait. It is deliberately landing in the same release window as #550's trait change so external backends absorb one break rather than two.

## Tests

A per-qubit oracle independent of the word-level formula backs randomized comparison across all three backends over qubit counts spanning word boundaries and the `SmallVec` spill threshold, plus deterministic unequal-length fixtures exercising both zero-padding directions. The exhaustive single-qubit test pins all eight (Pauli, state bit) combinations against hardcoded expected images and phases on every backend.

The strongest check shares no oracle with either implementation: applying `p` then `q` to a basis state must agree, in both image and total phase, with applying the product `q*p` computed by `multiply_with_phase`. That ties the basis-state phase convention to the operator-product one.

Mutation-tested: flipping any of the three coefficients in either the word helper or the u128 implementation, or narrowing either backend's loop bound, fails the suite.

### DIFF
--- a/crates/pecos-core/src/pauli/pauli_bitmask.rs
+++ b/crates/pecos-core/src/pauli/pauli_bitmask.rs
@@ -26,6 +26,16 @@ fn phase_terms_u64(lx: u64, lz: u64, rx: u64, rz: u64) -> u64 {
         + 3 * u64::from((!lx & lz & rx & rz).count_ones()) // Z*Y = -iX  (+3)
 }
 
+/// Per-word quarter-phase contribution of a Pauli acting on a computational
+/// basis state, for the `u64`-word backends. `x`/`z` are the Pauli X/Z plane
+/// words, `r` the basis-state word (set bit = that qubit is in state one).
+#[inline]
+fn basis_state_phase_terms_u64(x: u64, z: u64, r: u64) -> u64 {
+    u64::from((x & z & !r).count_ones()) // Y|0> = i|1>   (+1)
+        + 3 * u64::from((x & z & r).count_ones()) // Y|1> = -i|0>  (+3)
+        + 2 * u64::from((!x & z & r).count_ones()) // Z|1> = -|1>   (+2)
+}
+
 /// Trait for bitmask storage backends.
 ///
 /// Enables `PauliBitmaskGeneric<B>` to work with different widths:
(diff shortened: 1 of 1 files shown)


ciaranra:
Independent two-arm review of this commit (it was the one piece of the batch written and verified by the same party, so it got fresh eyes): both arms returned SHIP with no must-fixes.

Highlights: the Y = iXZ convention was re-derived from the merged multiplication table and shown to be forced, with no residual convention freedom; the three phase coefficients and the absence of a fourth case were confirmed from first principles by both arms independently; the loop bound z_len.min(x_len.max(state_len)) was proven to be exactly the contribution condition; and a deliberate attempt to slip a conjugated phase convention (k -> 4-k) past the suite failed — the composition test catches it whenever the two Paulis anticommute, and the hardcoded single-qubit tables block the compensating double-reversal. A 17-mutant campaign killed 15, with the 2 survivors proven behavior-equivalent.

Follow-up commit applies the two test-strengthening findings: the unequal-length fixtures now assert the image (previously phase-only, flagged by both arms), and the wide state gains a set bit at qubit 640 so a pure-Z contribution sits in a word past the X plane's stored length — the one bound mutation that previously survived the deterministic fixtures now fails all five tests.

PR #579, opened 2026-08-22
Remove the singular per-shot sampler

ciaranra:
Removes `StabMps::sample_bitstring(num_shots)` from the Rust API and Python binding, resolving a naming hazard: it shared an identical signature with `sample_bitstrings(num_shots)` while differing by one letter — but the singular cloned the full simulator per shot and followed the measurement mode, while the plural shares projection work via the prefix tree, is always exact, and is tens-to-hundreds-fold faster. A one-letter typo silently selected the slow path.

With exact measurement now the default (#567), the singular's only remaining distinction was being slower and mode-following. That intent is better expressed explicitly: code that requires each shot to follow the configured `MeasurementMode` should clone a fresh simulator per shot and measure with `mz` — which is exactly how the falsifier matrix's per-shot statistical surface is now implemented (`per_shot_mz_on_clone`), preserving all nine surfaces of measurement-statistics coverage in both Exact and Lazy matrices. The `sampling_methods` example keeps its performance comparison via the same explicit-loop leg.

Breaking change: `sample_bitstring` (singular) is removed; use `sample_bitstrings` for sampling, or an explicit per-shot `mz` loop for mode-following semantics.

Verification: whole-repo grep clean of the removed symbol; full suites debug (436) and release (454) green; the release-gated falsifier lane passes with all 20 matrices/stress tests; clippy -D warnings and fmt clean on both crates; the updated Python exposure test passes against a rebuilt extension.

### DIFF
--- a/docs/user-guide/stabilizer-tensor-networks.md
+++ b/docs/user-guide/stabilizer-tensor-networks.md
@@ -10,7 +10,7 @@ All public bitstrings use qubit-index order: `bits[q]` is the bit for qubit `q`.
 
 ## `StabMps` quickstart
 
-Use `sample_bitstrings`, plural, for shot workloads. It shares each distinct measurement-prefix projection between shots and is always exact by construction, independent of the singular-measurement policy. `sample_bitstring` clones and collapses the entire simulator once per shot and follows that policy. The two methods do not share an RNG stream, so seeded outputs should not be compared shot for shot across methods.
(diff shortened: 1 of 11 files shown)


ciaranra:
Review follow-up (commit just pushed): the replacement per-shot recipe had a real defect the light review caught — `set_rng` reseeded only the main RNG while cloned simulators shared the parent's tableau RNG (which the trivial-MPS fast path draws from), so clone+`set_seed` sampling produced fully correlated shots in Lazy and Pragmatic modes (400 shots, one distinct bitstring; Exact unaffected). `set_rng` now derives the tableau stream from the supplied RNG (one documented draw), matching what the removed method did by hand. Bound by 400-shot stabilizer-only regressions in both modes, mutation-verified (removing the tableau reseed collapses both to a single bitstring). Also: fragment-built positive control added to the exposure test, and terminology nits fixed.

Issue #585, opened 2026-08-24
Add qutrit state-vector and density-matrix simulators for noise verification


qciaran:
## Summary

Add production-quality qutrit state-vector and density-matrix simulators to PECOS, then use them as independent references for verifying leakage-aware noise models, including `GeneralNoiseModel` and its Selene plugin.

The initial physical basis is:

```text
|0>, |1>, |L>
```

The implementation should remain device-neutral. A generic qudit core is preferable if it does not compromise a clear qutrit API or delay the initial leakage use case.

## Motivation

PECOS currently has two complementary pieces, but no production qutrit simulator:

- The Rust density-matrix simulator is qubit-specific. It represents an N-qubit density matrix using a 2N-qubit state vector and assumes a local dimension of two throughout its indexing, gate traits, measurement, and state-access APIs.
- `GeneralNoiseModel` represents leakage efficiently as stochastic classical bookkeeping: leaked qubits are tracked separately, ordinary gates are suppressed, and seepage returns a qubit to the computational subspace.
- PR #545 adds a small test-only NumPy qutrit density-matrix oracle to verify the device-neutral Selene general-noise plugin.

The existing stochastic model is appropriate for scalable QEC studies, but an independent physical qutrit simulation would let us verify that its ensemble behavior is correct and explore coherent leakage effects that classical leakage bookkeeping cannot represent.

### Qutrit state vector

Store `3^N` complex amplitudes and support shot-based quantum trajectories.

Required capabilities:

- arbitrary one- and two-site qutrit unitaries;
- computational-subspace qubit gates embedded as `U + |L><L|`;
- native coherent coupling between computational and leakage levels;
- computational and full `{0, 1, L}` measurements;
- reset and preparation;
- sampled Kraus channels with outcome probability, application, and renormalization;
- deterministic seeding;
- state/probability inspection suitable for tests.

This should be the more scalable physical-leakage backend: `O(3^N)` state storage, with mixed channels represented through trajectories across shots.

### Qutrit density matrix

Store an exact `3^N x 3^N` density operator.

Required capabilities:

- the same unitary, measurement, reset, and preparation semantics as the state-vector simulator;
- exact Kraus-channel application;
- reduced density matrices and outcome distributions;
- density-operator inspection;
- validation helpers for trace preservation, Hermiticity, and positive semidefiniteness.

This is expected to be a small-system reference backend because storage scales as `O(9^N)`.

## Architecture

[...]

PR #588, opened 2026-08-24
Add qutrit state-vector and density-matrix reference simulators

qciaran:
## Summary

- add shared uniform-local-dimension QuditStateVec and QuditDensityMatrix reference backends with qutrit aliases for |0>, |1>, |L>
- support arbitrary multi-site unitaries, embedded qubit gates, preparation, reset, full and strict computational measurements, and exact or trajectory Kraus channels
- expose Kraus branch probabilities, joint basis outcomes, coarse-grained projective partitions, and generalized measurement instruments with exact mixed conditional states
- provide pure- and mixed-state reduced density matrices, density purity, and trace, Hermiticity, and positive-semidefiniteness diagnostics
- validate imported density operators and unitary/channel completeness; use fallible dense allocation and expose memory estimates
- include device-neutral qutrit leakage and seepage channel constructors
- document radix ordering, API semantics, and O(d^N) / O(d^(2N)) scaling

## Verification

- 19 focused qudit tests, including randomized d=2 through d=4 circuits, exact-versus-trajectory convergence, multi-site channels, generalized measurements, and backend cross-validation
- complete pecos-simulators test suite
- strict Clippy with warnings denied
- cargo formatting
- all pre-commit hooks

A rustdoc run with all existing warnings promoted to errors reaches unrelated pre-existing broken intra-doc links outside the new qudit module; the normal crate doctest suite passes.

This is the native simulator foundation from issue #585. GeneralNoiseModel and Selene boundary verification can follow after PR #545 lands without coupling the core simulator API to the plugin branch.

Progresses #585.


ciaranra:
## Review: multi-reviewer panel, findings verified by execution

Reviewed with three independent reviewers (fresh-context numerics, API/test-quality, and a cross-model pass), each blind to the others. Every finding below that drives a recommendation was then re-verified locally against this branch; the commands and outputs are quoted.

### The mathematics is correct

This is the strongest result of the review, and it is worth recording explicitly. Three independent oracles were built against this branch and none of them broke the implementation:

- NumPy tensor oracle at `d=3, N=3`, targets `[2, 0]`: unitary application, exact Kraus evolution, both reduced-density-matrix orders, joint probabilities, purity, and mixed conditional instruments agreed to `8.4e-17`.
- NumPy Haar oracle at `D=27` and `D=81` with non-adjacent and reversed target lists (`[2,0]`, `[0,2]`, `[3,1,0]`, `[2,3,0]`): twelve separate checks, all within `~1e-16`. Trajectory sampler unbiased (chi-squared 7.93 on 8 dof over 60,000 `measure_joint` shots); 40,000-shot Kraus ensemble matched exact `sum K rho K^dag` to `6.3e-4` statistical.
- Analytic probes: Bell-state partial trace, asymmetric embedded unitary orientation, instrument post-state normalization, incomplete-partition rejection.

The radix convention (`global = sum digit[s] * d^s`, `local = sum digit[targets[j]] * d^j`, so `targets[0]` is the least-significant local digit) is used consistently by `extract_local_index`, `replace_local_index`, `traced_digits_match`, and every caller. That was the highest-risk defect class for this module and it is not present.

One suspicion raised during review was that `diagnostics()` uses `nalgebra::symmetric_eigen` on a `Complex64` matrix, which could silently treat a Hermitian matrix as complex-symmetric. Checked directly against nalgebra 0.35: `[[0.5, 0.6i], [-0.6i, 0.5]]` returns eigenvalues `[1.1, -0.1]` and `[[0.6, 0.5i, 0], [-0.5i, 0.6, 0], [0, 0, -0.2]]` returns `[1.1, 0.1, -0.2]`. The Hermitian path is correct. Concern withdrawn.

### 1. The test suite does not falsify the implementation (highest priority)

A mutation sweep over the implementation left 10 of 29 mutants alive. Five of the most consequential were re-run locally to confirm. Baseline is `21 passed`, and each mutation below still produces `21 passed; 0 failed`:

| Mutation | Result |
|---|---|
| `traced_digits_match` (`qudit.rs:2188`) short-circuited so the partial trace sums *every* matrix element regardless of environment digits | 21 passed |
| `measure_instrument` (`qudit.rs:679`) collapses into `branches.swap_remove(0)` instead of `swap_remove(selected)` | 21 passed |
| `embedded_qubit_unitary` (`qudit.rs:1704-1705`) with the 2x2 block transposed | 21 passed |
| `purity` (`qudit.rs:1210-1211`) computing `sum rho^2` instead of `sum abs(rho)^2` | 21 passed |
| `validate_measurement_partition` (`qudit.rs:1944-1948`) with the coverage check removed | 21 passed |

The root cause is structural, not a matter of adding a few asserts. `state_vector_and_density_matrix_agree_for_coherent_qutrit_evolution` and `randomized_multilevel_unitaries_cross_validate_backends` derive the density expectation from the state-vector backend, and both reduced-density-matrix call sites go through the same `traced_digits_match` helper. A shared defect therefore agrees with itself. The remaining RDM tests use states whose density matrices are diagonal, where the partial-trace mutation is a no-op.

Two further consequences worth calling out:

- `embedded_qubit_gates_leave_the_leakage_level_unchanged` asserts `probability(2) == 1.0`, which is phase-blind by construction. Replacing the `|L><L|` identity block with `-1` passes the test named for that contract.
- The state-vector `measure_instrument` post-state is never inspected; `generalized_measurement_instruments_preserve_outcome_semantics` checks only the returned `InstrumentSample` fields and the density backend's post-state. The trajectory backend's core documented behaviour has no falsifying coverage.

Suggested fix: commit an independent oracle test (fixed complex state, `d=3`, `N>=3`, non-adjacent and reversed targets, expected amplitudes/matrices as literals derived outside this module) covering unitary application, both RDM orders, exact Kraus, purity, and both instrument post-states. That single test kills all five mutations above.

Claims from the PR summary with no falsifying test today: arbitrary multi-site unitaries beyond 2 sites; pure and mixed coherent reduced density matrices; density purity off the diagonal; the coherent action of the leakage/seepage constructors (only populations are checked); state-vector instrument post-states; partition completeness; density-import length/trace/finiteness validation; `AllocationFailed`.

### 2. Tolerance asymmetry: a state this simulator produces cannot be re-imported

`OPERATOR_TOLERANCE` is `1e-10` (`qudit.rs:28`) and gates are accepted at that bound, but `validate_state` checks the imported norm at `PROBABILITY_TOLERANCE` `1e-12` (`qudit.rs:1900`), and no code path renormalizes the state vector.

Reproduced on this branch. Applying the exact ternary Fourier matrix repeatedly to one qutrit and attempting a round trip through `from_state` at every 1000 gates:

```
FIRST from_state REJECTION: (4000, 1.077360423096252e-12, "NotNormalized { norm: 1.0000000000010774 }")
```

That is 4000 gate applications before the simulator's own output stops being valid input to its own constructor.

A note on the shape of the drift, since it was reported differently by two reviewers. In this configuration the drift plateaus at `norm - 1 = 1.5991e-11` and stays there through 40,000,000 applications, so `apply_kraus(qutrit_leakage_channel(0.0))` -- the identity channel -- is never rejected here. A different configuration (three-site operators on four qutrits) showed continued growth and did reach channel rejection. The drift shape is operator-dependent; the import-side rejection is the robust and cheaply-reachable failure.

Suggested fix: pick one tolerance policy for norm-like quantities and apply it at both ends, or restore normalization explicitly. Separately, `QuditError::NotNormalized` is currently emitted for three distinct causes (state-norm drift, channel completeness, density trace), so a user whose channel is exact and whose state has drifted is told their channel is not normalized.

### 3. Sub-tolerance negative probabilities reach the sampler

`from_density_matrix` accepts a minimum eigenvalue down to `-1e-11` (`qudit.rs:1108`, `qudit.rs:1266`). Reproduced:

```
input:  diag(-5e-12, 0.5, 0.500000000005)   (Hermitian, trace 1)
accepted by from_density_matrix
outcome_probabilities(0) = [-5e-12, 0.5, 0.500000000005]
```

`sample_distribution` (`qudit.rs:2162`) neither validates nor clamps individual entries; it subtracts each from a running threshold, so a negative entry *increases* the threshold and shifts weight onto later branches. The same unguarded path is reached from `instrument_probabilities`.

Suggested fix: sanitize the probability simplex at the sampling boundary -- reject materially negative entries, clamp tolerance-scale negatives, recompute the total from the sanitized entries.

### 4. Empty target lists are silently accepted

`validate_targets` (`qudit.rs:1916`) has no emptiness check, so `d^0 = 1`. Reproduced:

```
apply_operator(&[], &[i])  -> Ok            (applies a global phase)
measure_joint(&[])         -> Ok(MeasurementSample { outcome: 0, probability: 1.0 })
```

The doc comment on `apply_operator` says "one or more sites", and the module is otherwise strictly fail-loud (it rejects duplicate targets, out-of-range targets, non-unitary operators, and leakage population). A caller who computes an empty target list from upstream data gets a silent no-op measurement reporting outcome 0. Worth deciding the contract explicitly: reject for local operations and measurements, or document the zero-site semantics.

### 5. `RngManageable` is not implemented, and `reset` will become a breaking rename

`RngManageable` has 18 implementations across 14 files in this crate (`DensityMatrix`, `StateVecSoA`, `SparseStateVecSoA`, every stabilizer variant). `QuditStateVec` and `QuditDensityMatrix` hold an `rng: R` field but implement neither it nor `QuantumSimulator`, so there is no way to reseed or inspect the RNG of a constructed simulator without discarding and rebuilding it.

This is worth settling before merge rather than after, because `reset(&mut self, target: usize)` (`qudit.rs:787`, `qudit.rs:1594`) collides in name with the trait's `reset(&mut self)`. Adopting the trait later is a breaking rename of a public method.

### Smaller items

- `prelude.rs:25-28` exports the simulator types, sample structs, and channel constructors, but not `QuditError`, which every method on those types returns, nor `DensityMatrixDiagnostics`, `basis_swap`, or `embedded_qubit_unitary`. `lib.rs:100-104` exports all of them; the two lists disagree with no stated rationale.
- The README states a fallible dense-allocation guarantee, but the transient buffers at `qudit.rs:481`, `541`, `711`, `1294`, `1459`, and `1518` use infallible `vec![...]` and will abort rather than return `AllocationFailed`. Either use fallible reservation throughout or narrow the documented claim.
- `validate_kraus_slices` (`qudit.rs:2021`) initialises `local_size = 0` and returns `Ok(0)` for an empty operator slice, certifying an empty channel as trace preserving. All three current callers guard against this, so it is a latent trap rather than a live defect.
- `measure_computational` (`qudit.rs:774`, `qudit.rs:1581`) passes a hard-coded `total = 1.0` to `sample_distribution` where every other call site passes the actual sum. With the sub-tolerance leakage the method deliberately admits, the sampling interval no longer matches the mass in the slice and the draw falls through to the `last_nonzero` fallback. The magnitude is currently unobservable (60,000 shots of `(|0> + |1>)/sqrt(2)` gave a one-fraction of 0.50343, z = 1.68), but the contract is inconsistent.
- `collapse_local_partition` (`qudit.rs:855`, `qudit.rs:1666`) rejects a selected outcome with `probability <= 1e-12` as `ZeroNorm`, while `apply_kraus` and both `measure_instrument` implementations divide by `sqrt(p)` or `p` with no lower guard at all. Neither is reachable at realistic shot counts, but the asymmetry is unintended.
- `QuditStateVec::reset` is implemented as measure-then-swap, so it draws from the RNG and can return `ZeroNorm`; `QuditDensityMatrix::reset` is the deterministic `{|0><i|}` channel. Both are individually correct, but the tests at `qudit_simulator_tests.rs:406` and `:459` assert seed-matched outcome equality across backends, which a `reset` inserted before such a comparison would desynchronise. The doc for the state-vector `reset` ("Reset a site to local basis state zero") does not disclose that it measures.
- `QutritStateVec` / `QutritDensityMatrix` are newtypes over the general types that add only `d = 3` and expose the parent API through `Deref`/`DerefMut`. `qutrit()`, `qutrit_with_seed()`, `with_rng`, and `into_inner` are called zero times in the test suite.
- `docs/user-guide/simulators.md:83-84` lists these under Requirements "Rust API", where every other row carries a build or runtime dependency and every other listed simulator is reachable from the Python `sim(...).quantum(...)` builder.

### Recommendation

REVISE. The implementation is correct and the design is sound; the gap is that nothing in the repository would catch it becoming incorrect, which matters more than usual because this module is intended as the reference everything else is validated against.

Blocking before merge:

1. An independent-oracle regression test that kills the five mutations in section 1.
2. A single coherent tolerance policy for norm-like quantities (section 2).
3. Probability sanitisation at the sampling boundary (section 3).
4. An explicit empty-target contract (section 4).
5. A decision on `RngManageable` and the `reset` signature, because deferring it makes it a breaking change (section 5).

One gap no reviewer covered: every oracle used here was written for this review. Nothing checks these backends against the `GeneralNoiseModel` or Selene boundary that #585 is ultimately for, so if that boundary carries its own digit-ordering convention, the choice made here is load-bearing in a way this review did not test.

qciaran:
Implemented the review follow-up in f6d4bff8 and merged current dev in 2b423687a.

Addressed the blocking findings:
- Added independent literal-oracle coverage for coherent partial traces and reversed target order, asymmetric embedded-unitary orientation, off-diagonal purity, state-vector and density-matrix instrument post-states, complete measurement partitions, three-site reversed-target operations, and exact leakage-channel coherences.
- Re-ran the five previously surviving mutations individually; every one is now killed by a focused test.
- Unified imported-state and density normalization tolerance with the accepted operator tolerance, with a 10,000-operation round-trip regression.
- Sanitized probability weights at the sampling boundary: reject material negatives, clamp tolerance-scale negatives, and recompute totals after clamping.
- Made empty local target lists fail with QuditError::EmptyTargets.
- Implemented RngManageable for all generalized and qutrit wrapper types.
- Renamed the inherent site reset API to reset_site before merge and documented trajectory RNG consumption versus exact density reset.
- Completed the qudit prelude exports and narrowed the dense-memory documentation to match the actual allocation guarantees.

Post-merge local verification is clean:
- cargo fmt --all --check
- cargo test -p pecos-simulators --locked (all unit, integration, and doc tests pass; qudit suite is now 27 tests)
- cargo clippy -p pecos-simulators --all-targets --locked --offline -- -D warnings

CI has restarted against the updated head.

ciaranra:
## Follow-up review of f6d4bff8: all five blocking items verified closed

Re-verified each item by execution against `2b423687a`, rather than by reading the change. The REVISE recommendation is withdrawn.

### Blocking items

**1. Independent-oracle regression test.** All five mutations from the previous comment are now killed. Baseline is `27 passed`; each mutation below fails a named test:

| Mutation | Killed by |
|---|---|
| `traced_digits_match` short-circuited so the partial trace ignores environment digits | `reduced_density_matrices_match_literal_multisite_oracles` (`:691`) |
| `measure_instrument` collapsing into `swap_remove(0)` | `instrument_post_states_and_partition_completeness_have_literal_oracles` (`:804`) |
| `embedded_qubit_unitary` 2x2 block transposed | `embedded_unitary_and_purity_match_literal_complex_oracles` (`:743`) |
| `purity` computing `sum rho^2` | `embedded_unitary_and_purity_match_literal_complex_oracles` (`:743`) |
| `validate_measurement_partition` coverage check removed | `instrument_post_states_and_partition_completeness_have_literal_oracles` (`:804`) |

**2. Tolerance policy.** `NORMALIZATION_TOLERANCE` is now applied at state import, density trace, and physicality validation. Re-ran the drift reproduction with the same ternary Fourier matrix, extended to 5,000,000 applications: worst `abs(norm - 1)` is `1.5991e-11` and `from_state` never rejects. The previous head rejected its own output after 4,000 applications.

**3. Probability sanitisation.** `diag(-5e-12, ...)` and `diag(-5e-11, ...)` are now accepted at import and clamped, with `outcome_probabilities(0)` returning `[0.0, 0.5, ...]`; `diag(-5e-10, ...)` is rejected at import as `NotPositiveSemidefinite { minimum_eigenvalue: -5e-10 }`. Sanitisation happens inside `sample_distribution`, so every sampling path is covered rather than each call site individually.

**4. Empty targets.** `EmptyTargets` is returned from `apply_operator`, `measure_joint`, and `reduced_density_matrix` on the state-vector backend and from the density backend's `measure_joint`.

**5. `RngManageable` and the `reset` signature.** The trait is implemented on all four types and reseeding round-trips. Renaming the inherent method to `reset_site` before merge is the right call -- it removes the breaking rename that deferring would have caused. The trajectory-versus-exact reset asymmetry is now documented in both the README and the user guide.

`measure_computational` also dropped the hard-coded `total = 1.0` and recomputes from the sanitised slice, which is a better fix than the one suggested.

### Verification reproduced locally

- `cargo fmt --all --check` -- exit 0
- `cargo clippy --locked -p pecos-simulators --all-targets -- -D warnings` -- exit 0, run with `qudit.rs` touched first so the cache could not skip the changed code
- `cargo test --locked -p pecos-simulators` -- exit 0 across all lanes (755 library tests; qudit suite 27)
- CI on this head: 83 successful, 8 skipped, 0 failures

### Two new guards are themselves untested (non-blocking)

Applying the same mutation discipline to the code added in this commit, two guards survive the full 27-test suite:

- Removing the material-negative rejection in `sanitize_probability_weights` (`qudit.rs:2078`) leaves all 27 tests passing. Reachability was checked: the guard does fire, via `from_density_matrix_unchecked` with `diag(-0.2, 0.6, 0.6)`, giving `InvalidProbability(-0.2)`. With the guard removed the same input still errors, as `NotNormalized { norm: 1.2 }`, because clamping only ever raises the total. So this is a diagnostic-quality gap rather than a safety hole.
- Mutating `set_rng` to a no-op also leaves all 27 passing. The `reseed` helper (`qudit_simulator_tests.rs:52`) calls `set_seed(131)` and asserts nothing, so it pins that the trait is implemented but not that reseeding controls the stream. One assertion -- reseed two simulators to the same seed and compare measurement sequences on a state with a non-trivial distribution -- would close it.

Two items from the previous comment's smaller-items list are unchanged and remain reasonable to leave: `validate_kraus_slices` (`qudit.rs:2140`) still returns `Ok(0)` for an empty operator slice, which stays latent because all three callers guard it; and `qutrit()`, `with_rng`, and `into_inner` are still referenced zero times by the test suite.

Neither of the above is worth holding the PR for. The blocking objection is withdrawn.

qciaran:
Closed the two non-blocking mutation gaps from the follow-up in 00d8bbdf.

- The RNG contract test now verifies the first post-reseed draw for all four generalized/qutrit state-vector and density-matrix types against PecosRng seeded identically. A no-op QuditStateVec::set_rng mutation now fails the focused test.
- Added a checked diagnostic regression for an unchecked trace-one density with a material negative weight. It requires InvalidProbability(-0.2) from both probability inspection and measurement. Removing the material-negative guard now fails with the distinguishable NotNormalized { norm: 1.2 } result.

Mutation-audited both regressions individually, restored the implementation, then verified:
- qudit simulator suite: 28 passed
- cargo fmt --all --check
- strict all-target pecos-simulators Clippy

I left the explicitly latent empty-Kraus helper and zero-use convenience-constructor observations unchanged, as recommended.

qciaran:
Closed the remaining smaller review items in 2fe2c7df:

- `validate_kraus_slices` now rejects an empty channel itself; the duplicated outer guards were removed, so both state-vector and density-matrix public paths exercise the centralized invariant.
- Added explicit empty-channel coverage for both backends. As a mutation check, deleting the centralized guard makes the test fail with `ZeroNorm` instead of the required `EmptyKrausChannel`.
- Added behavioral coverage for all four `qutrit()` convenience constructors.
- Added wrapper `with_rng` / `into_inner` coverage that verifies dimension, evolved state, and continuation of the injected deterministic RNG stream.

Local verification is clean:

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p pecos-simulators --locked --offline` (including 29 qudit integration tests and 90 doctests)
- `cargo clippy -p pecos-simulators --all-targets --locked --offline -- -D warnings`

PR #590, opened 2026-08-24
Expose the factor model and maxlog metric through the Python Frontier bindings

ciaranra:
## Summary

Stacked on #589 (which stacks on #587); retarget as the stack merges. Exposes the Phase A/B engine surface to Python.

- `FrontierDecoder.from_dem` and `FrontierCommitteeDecoder.from_dem` gain `metric_mode` (upstream's full alias set: `logsumexp_float`/`float`/`exact` and `frontier_lite`/`maxlog_int`/`viterbi_int` variants) and `int_metric_scale` keyword arguments. The committee's max-log rejection surfaces as a clean `RuntimeError` from the engine.
- New `FrontierDecoder.from_factors(factors, num_detectors, num_observables, ...)`: a list of factors, each a list of `(probability, detectors, observables)` outcome tuples, validated by the engine; named column orders resolve through the factor-model ordering functions, never by laundering the model through a DEM.
- Error-type convention held file-wide: binding-level argument parsing raises `ValueError`; every engine `DecoderError` maps to `RuntimeError`.
- `FactorModel::new` now bounds the declared widths at the `u32` outcome-index range — an adversarial probe showed `from_factors(..., 10**15, 1)` previously aborted the interpreter with a failed 8 PB allocation instead of raising.
- Docs: factor-model section with a tested example fence, `int_metric_scale` documented, and the maxlog clauses updated (decoder accepts it from Python; committee remains float-only).

## Verification

- 28 Python tests in `tests/qec/test_frontier_decoder.py`, including: hand-derived degeneracy masses through `from_factors`; the metric-flip case (float predicts label 0, `maxlog_int` predicts 1); exact `==` pins of `int_metric_scale` behavior at scales 1 and 1024; delegation equality vs `from_dem` parametrized over all three named column orders; and error paths for every guard (unknown metric, committee+maxlog, infinite delta under maxlog, BP+N-ary, bad probability sums, oversized widths).
- Doc example fences regenerated and executed by the doc-test framework (Rust crate and generated Python docs both green).
- An independent adversarial review ran against the diff and upstream's alias table; every finding was verified and closed (the interpreter-abort guard and the error-type unification came from it). New guards mutation-tested.
- `cargo clippy --all-targets -- -D warnings` (including `pecos-rslib-exp`), `cargo fmt --check`, the three-crate engine test battery, `maturin develop` + full pytest file, and the repo-wide pre-commit lane all pass.

### DIFF
--- a/docs/experimental/decoders.md
+++ b/docs/experimental/decoders.md
@@ -14,11 +14,12 @@ Two capabilities live here that the production decoders do not offer:
 
 ## Frontier and BP-Trellis
 
-Both take a Stim-format DEM string and expose the same result surface.
-`FrontierDecoder` and `BpTrellisDecoder` are each a single trellis decoder over
-the same machinery (BP-Trellis adds BP-informed pruning scores); the separate
-`FrontierCommitteeDecoder` runs a two-leg forward/backward committee over them.
-Start with whichever you like and compare.
+`FrontierDecoder.from_dem` and `BpTrellisDecoder.from_dem` take a Stim-format
+DEM string and expose the same result surface. `FrontierDecoder.from_factors`
+also accepts the factor model described below. Each is a single trellis decoder
+over the same machinery (BP-Trellis adds BP-informed pruning scores); the
+separate `FrontierCommitteeDecoder` runs a two-leg forward/backward committee
+over them. Start with whichever you like and compare.
 
 <!--test-name: experimental_decoders_gap-->
 ```python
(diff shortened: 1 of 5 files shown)


ciaranra:
Review-panel follow-up pushed: the factor-model width bound was off by one — index u32::MAX exists, so width u32::MAX+1 is addressable, and the DEM parser already accepts D4294967295 with exactly that width. FactorModel::new now matches that contract (a maximal-width model with a u32::MAX index constructs; one wider is rejected), so TryFrom<&SparseDem> can no longer reject a DEM the parser accepts. Guard mutation-tested.

ciaranra:
Post-panel delta review fixes: the width-guard error rename had left the Python test pinning the old string (masked locally by a stale maturin-installed module — the test file now matches, verified against a freshly rebuilt module), and the width comparison now happens in u64 so the u32::MAX+1 constant cannot const-overflow on 32-bit targets, matching the DEM parser's own platform handling.

Issue #575, opened 2026-08-22
DEM two-qubit noise gate list is duplicated three times and omits U2q, RXXRYYRZZ, and CRZ


ciaranra:
`is_two_qubit_noise_gate` (`crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs:3234`) enumerates the two-qubit gates that receive two-qubit noise during DEM construction:

```
CX | CZ | CY | SZZ | SZZdg | SXX | SXXdg | SYY | SYYdg | SWAP | RXX | RYY | RZZ
```

Three two-qubit gate types in `GateType` are missing from that list: `U2q`, `RXXRYYRZZ`, and `CRZ`.

## Consequence

The predicate is used in two places, and a gate absent from it is skipped by both:

- `builder.rs:1529` -- the source-tracked fault loop has an arm `gate_type if is_two_qubit_noise_gate(gate_type) && !loc.before => {}`, and the match terminates in `_ => {}` (line 1574). A `U2q` / `RXXRYYRZZ` / `CRZ` location therefore reaches the catch-all and contributes nothing.
- `two_qubit_after_location_pairs` (`builder.rs:3253`) groups locations for two-qubit fault processing using the same predicate, so those gates are never paired.

The result is a DEM built from a circuit containing these gates that silently omits their two-qubit noise. There is no error and no warning; the DEM is simply missing mechanisms, which makes any threshold or logical-error-rate number computed from it optimistic.

`RXXRYYRZZ` is the case most likely to be hit in practice, since it is the native two-qubit interaction for trapped-ion hardware and is already a first-class `GateType` (`crates/pecos-core/src/gate_type.rs:81`).

## Suggested scope

Decide per gate whether it should receive two-qubit noise and add it to the predicate, or be explicitly rejected. Silently contributing nothing is the one option that should not remain. If a gate is deliberately unsupported by the DEM builder, it should fail loudly rather than produce a quietly incomplete model.

A regression test that builds a DEM from a circuit using `RXXRYYRZZ` and asserts the expected two-qubit mechanisms are present would cover the fix.

Related: #549 records the same silent-identity shape in Pauli propagation for an overlapping set of gates.

ciaranra:
Correction to the scope above, found in a follow-up review: the omission is not confined to `builder.rs`, and the other two sites do not go through `is_two_qubit_noise_gate` at all. Each carries its own inline copy of the same thirteen-gate list, with the same three gates missing:

[...]

Issue #549, opened 2026-08-17
Unhandled non-Clifford gates and QFree are silent identity in Pauli propagation


ciaranra:
Surfaced by the adversarial review of #541; both items are pre-existing and neither is changed by that PR.

## Silent identity for gates the propagator cannot model

`apply_gate` (`crates/pecos-qec/src/fault_tolerance/propagator/pauli.rs`) ends in a `_ => {}` arm, so `T`, `Tdg`, `U`, `CCX`, `CRZ`, `CH`, `RXXRYYRZZ`, `U2q`, `Channel`, and `Custom` propagate as identity. The rotation family is handled only at Clifford angles (`try_simplify_rotation` / `half_turn_decomposition`); other angles fall through the same way.

A Pauli frame genuinely cannot represent a non-Clifford conjugation, so identity is not a fixable approximation -- but the consequence is that a fault-tolerance checker handed such a circuit returns a confident verdict computed from a model that silently ignored those gates. Failing loud (or requiring an explicit opt-in acknowledging the gates were skipped) would keep the verdict honest.

## `QFree` does not clear its wire

`GateType::QFree` is absent from `apply_gate_flip_ledger`'s clearing arm (`crates/pecos-qec/src/fault_tolerance/pauli_prop_checker.rs`), which clears on `PX`/`PZ`/`QAlloc`/`MPZ`. A ledger entry therefore survives on a freed qubit and can be read back later, exactly the phantom-flip shape the clearing arm exists to prevent. If a freed qubit can be re-allocated, the stale entry outlives its wire.

## Suggested scope

Two independent changes: a decision about how the checker should react to gates it cannot model, and adding `QFree` to the clear arm with a regression test. The second is small and self-contained.

ciaranra:
Correcting the `QFree` half of this issue -- I checked it by execution and my original description was partly wrong, and the fix is not as safe as I implied.

Probing `apply_gate_flip_ledger` directly:

```
after QFree:                    x = true     (entry survives -- as filed)
after QFree then QAlloc:        x = false    (re-allocation clears it)
QFree then CX(freed, other):    other x = true  (spreads)
```

**What I got wrong.** I wrote that "if a freed qubit can be re-allocated, the stale entry outlives its wire". It does not: `QAlloc` is already in the ledger's clear arm, so reuse re-clears. That failure mode does not exist.

[...]

PR #600, opened 2026-08-26
Fail loudly when DEM construction meets a gate Pauli propagation cannot represent

### LINKED ISSUE #594: DEM builders accept bare T/Tdg as fault locations while Pauli propagation ignores them
## Summary

The DEM-side builders classify bare `T` / `Tdg` (and non-Clifford `RX` / `RY` / `RZ` / `U`) as ordinary one-qubit noisy fault locations, while the Pauli propagator has no arm for `T` / `Tdg` and leaves a propagated Pauli unchanged when it meets one. The result is a detector error model in which faults are inserted around a gate that propagation treats as the identity. Nothing errors; the DEM is silently wrong.

This is the builder-side half of #549, which tracks the propagator silently ignoring non-Clifford gates. Fixing the propagator alone cannot help (T maps Paulis outside the Pauli group), so the builders themselves must reject the input.

## Locations

- Fault-location classification including `T` / `Tdg` / `RX` / `RY` / `RZ` / `U`:
  - `crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs:1529-1559`
  - `crates/pecos-qec/src/fault_tolerance/dem_builder/mem_builder.rs:122`
  - `crates/pecos-qec/src/fault_tolerance/dem_builder/dem_sampler.rs:2290`
- Propagator with no `T` arm, falling through unchanged: `crates/pecos-qec/src/fault_tolerance/propagator/pauli.rs:78-111`
- The fault-table path already rejects `T` even at zero noise: `crates/pecos-qec/src/fault_tolerance/fault_sampler.rs:2586` (`test_unsupported_gate_rejected_even_with_zero_noise`)

## Proposed resolution

Hard-error in all three DEM-side builders on any gate the propagator cannot conjugate, using one shared predicate rather than three hand-maintained gate lists (the same three copies were the cause of #575). Non-Clifford logical operations must be lowered to Clifford-plus-measurement gadgets (for example T injection) before DEM construction.

ciaranra:
Closes #594. Extends #549 with the builder-side half.

## Problem

The DEM builders classified bare `T` / `Tdg` -- and rotations with non-Clifford angles -- as ordinary one-qubit noisy fault locations, while the Pauli propagator had no arm for them and silently left the propagated Pauli unchanged. Faults were inserted around a gate that propagation treated as absent, producing a silently wrong detector error model with no error and no warning.

This cannot be fixed by teaching the propagator to conjugate `T`: `T` maps Pauli operators outside the Pauli group, so no conjugation rule exists. The input has to be rejected.

## Approach

`apply_gate` (`crates/pecos-qec/src/fault_tolerance/propagator/pauli.rs`) becomes the single source of truth, returning `PauliPropagationOutcome::{Propagated, Unsupported}`. `Propagated` covers gates actually conjugated plus those transparent to Pauli propagation by design; `Unsupported` covers everything the old code silently skipped.

The transparent set is not a new hand-written list. `is_supported_prep_gate` (`PX`, `PZ`, `QAlloc`) and `is_supported_noop_or_metadata_gate` (`QFree`, `I`, `Idle`, `TrackedPauliMeta`, both measurement-crosstalk payloads) were promoted into `fault_tolerance/propagator.rs`, and the classification, `validate_tick_circuit`, and the tick propagator all read that one definition. Maintaining several copies of one gate list is what produced #575.

Every DEM entry point converts `Unsupported` into a structured error naming the gate and its location, including the independently supplied exact-replay circuit accepted by `with_exact_branch_replay_context` (and exposed to Python), which was previously unguarded. Tick inputs are validated before DAG conversion so the error reports `Tick { tick, gate_in_tick }` rather than a node index in a derived DAG.

Validation of gate payloads happens once per circuit at the entry points; the repeated per-measurement propagation walks use a crate-private non-validating dispatcher that remains bounds-safe.

## Scope notes

Two behaviours are documented rather than changed, because altering them belongs in separate PRs:

- `R1XY` retains its existing `1e-9`-turn snapping tolerance while axis rotations require exact angle matching.
- `InfluenceBuilder` accepts rotations only at exactly zero angle and rejects every `R1XY`, including `R1XY(0, phi)`. This errs conservatively, so it is not silent wrongness; the accepted subset is now stated in its rustdoc.

Non-DEM consumers (Pauli-frame path, propagation checker, generic propagation drivers) keep their previous permissive behaviour explicitly.

## Verification

- `cargo test -p pecos-qec` -- 991 passed, 0 failed
- `cargo check --workspace --all-targets` -- clean
- `cargo clippy -p pecos-qec --all-targets -- -D warnings` -- clean, run cold
- Rust doc-test crate -- 176 passed
- Regression tests cover bare `T`, both eighth-turn `RZ` signs, non-Clifford `RZ`, malformed angle arity, the replay-context path, and `QAlloc` / `QFree` circuits building successfully through every DEM entry point
- Each new guard was mutation-tested: disabling it fails the intended tests


ciaranra:
## Merge-risk check: does this reject anything the repository actually builds?

This change makes previously-succeeding DEM construction fail on gates Pauli propagation cannot represent, so the question that matters for merging is whether any real workflow now errors. Answered two ways.

### The Python surface pipeline: complete gate inventory

Rather than run a sample of examples and hope they cover the paths, I enumerated every gate the Python surface generators can emit and checked each against the new classification.

Gate-emitting calls in `python/quantum-pecos/src/pecos/qec/surface/*.py`:

`cx`, `h`, `mz`, `mz_free`, `pz`, `qalloc`, `sz`, `szdg`, `szz`, `szzdg`, `x`, `z`, `tracked_pauli`

plus the generic `add_gate`, which is used at exactly two sites (`circuit_builder.py:2026` and `:2029`) with `GateType.SX` and `GateType.SXdg`.

That is the complete vocabulary: `CX`, `H`, `MZ`, `MeasureFree`, `PZ`, `QAlloc`, `SZ`, `SZdg`, `SZZ`, `SZZdg`, `X`, `Z`, `TrackedPauliMeta`, `SX`, `SXdg`. Every one of them classifies as `Propagated` -- the Cliffords and measurements through `apply_named_gate`, and `PZ` / `QAlloc` / `TrackedPauliMeta` through the shared transparent predicates. **No surface-code workflow is rejected by this change.**

`QAlloc` is worth calling out: an earlier revision of this PR omitted it from the transparent set, which would have rejected every circuit these generators produce. The Rust test suite did not catch it because Rust test circuits allocate with `pz()` rather than `qalloc()`. That is why this inventory is by enumeration rather than by sampling, and why the current tests include `qalloc` and `qfree` circuits building successfully through every DEM entry point.

### Rust examples

`influence_builder_example` runs clean.

`fault_distance_timing` fails both on this branch and on a clean `dev` checkout, with the same underlying error (`missing detector id`); it is pre-existing and unrelated. The only difference is where it surfaces: on `dev` it panics inside `dem_builder/builder.rs:194`, and on this branch the now-fallible API returns the error and the example's own `.unwrap()` panics on it. Same outcome, better mechanism -- the example simply does not handle the `Result`.

PR #603, opened 2026-08-26
Add the pecos-synth crate: exact rings and Matsumoto-Amano synthesis

ciaranra:
Phase 1 of the Clifford+T synthesis plan. Adds a new crate, changes no existing behaviour.

## What this is

`crates/pecos-synth` provides exact number-theoretic arithmetic and exact one-qubit Clifford+T synthesis. It is the foundation for Ross-Selinger `Rz` approximation; the grid problem, the norm equation and factoring are deliberately NOT here and are later work.

Everything is exact. There is no floating point anywhere in the crate, and no tolerances: downstream code will rely on exact equality, exact unitarity and canonical forms.

## Contents

**Rings.** `Z[sqrt2]`, `Z[omega]` with `omega = e^{i pi/4}`, and `D[omega] = Z[omega][1/sqrt2]`.

`Z[omega]` elements are four `BigInt` coordinates in the basis `(1, omega, omega^2, omega^3)`, constant term first -- the basis Selinger's `newsynth` uses, so future oracle fixtures compare coordinate-wise without a change of basis. Both Galois automorphisms are provided under names that say which is which, since they are easy to confuse: `sqrt2_conjugate` (`sigma_5`, fixes `i`, negates `sqrt2`) and `conjugate` (`sigma_7`, complex conjugation, negates `i`, fixes `sqrt2`).

Divisibility by `sqrt2` uses a parity criterion rather than trial multiplication. For `y = [y0, y1, y2, y3]`, `sqrt2 | y` iff `y0 = y2 (mod 2)` and `y1 = y3 (mod 2)`; this follows from `x * sqrt2 = [x1 - x3, x0 + x2, x1 + x3, x2 - x0]`, whose coordinates 0 and 2 share parity, as do 1 and 3. The exact quotient inverts that map directly. The derivations are reproduced as comments beside the code.

`D[omega]` is stored as `(u, k)` meaning `u / sqrt2^k` in least-denominator form, with equality and hashing defined on the canonical form rather than by cross-multiplication. The exponent is a fixed-width `u32`, never `usize`, so results are reproducible across platforms; the type documents the resulting capacity limit.

**Matrices and synthesis.** Exact 2x2 matrices over `D[omega]` with an exact unitarity test (`U^dagger U == I` in the ring), and Matsumoto-Amano normal form via `exact_synthesize`, returning a structured error rather than panicking on non-unitary input.

Words are in execution order: index 0 is applied first, so the matrix product is the reverse of the vector order. This is stated in the module docs because the opposite convention is equally common and would silently transpose results. Generators are the conventional `H`, `S = diag(1, i)`, `T = diag(1, omega)`, matching what PECOS's tokens execute after #597. The `omega^j` scalar is carried as a separate field, never as a gate token, and exact synthesis emits `T` and never `Tdg`. A generated 24-entry table gives each projective Clifford class a canonical shortest-then-lexicographically-first word.

## The oracle

The number of distinct one-qubit Clifford+T unitaries with T-count at most `n` is `192 (3 * 2^n - 2)` (Matsumoto-Amano). The enumeration test builds every normal form up to `n = 4` and checks that count.

It is worth being precise about why this is evidence rather than a tautology: the test keys a map **by matrix**, asserts every insertion is new, and takes its per-level count from *distinct matrices*, never from words enumerated. So two distinct normal forms colliding to one matrix fails immediately, and the count matching then shows the enumeration is both complete and non-redundant. Each enumerated matrix is additionally re-synthesised and must return exactly the word and scalar it came from.

    T-count  distinct  cumulative  192(3*2^n - 2)
          0       192         192             192
          1       576         768             768
          2     1,152       1,920           1,920
          3     2,304       4,224           4,224
          4     4,608       8,832           8,832

## Convention fixtures, and why they are not relational

`generator_convention_is_pinned_by_absolute_coordinates` asserts absolute coordinates for `omega`, `i`, `sqrt2`, the `S` and `T` generators, and one hand-computed asymmetric product (`[T, H]` in execution order is `H * T = (1/sqrt2) [[1, omega], [1, -omega]]`).

This exists because every other test in the crate is relational, and the coordinated relabelling `omega -> omega^{-1}`, `i -> -i` preserves all of them: `omega^2 == i`, `T^2 == S`, the automorphism identities, the round trip, and the enumeration count. That mirror image is a different convention -- it makes `T` the conjugate `diag(1, e^{-i pi/4})`, which is PECOS's `Tdg`. Applying that relabelling to the source fails exactly one test, this one, and leaves the other 22 green. Please do not rewrite these as relational checks.

## Verification

- `cargo test -p pecos-synth` -- 23 passed, 0 failed
- `cargo clippy --locked --all-targets -- -D warnings` from inside the crate directory, matching the CI recipe -- clean
- `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, pre-commit -- clean
- Mutation-tested: typo the `sigma_7` coordinate formula, drop the odd-`k` sign flip in the lifted `bullet`, change the `T` generator to `diag(1, i)`, or mirror the convention -- each fails tests


ciaranra:
## API review guide

The mathematics here has been carried by machinery -- the `192 (3 * 2^n - 2)` enumeration oracle over distinct matrices, six mutation tests, an adversarial review round, and a pin against `GateType::canonical_1q_matrix()` -- so the remaining review that matters is **API taste**: phases 2 and 3 build on these names and shapes. The full public surface is ~50 items; the decision points are below. Everything else is `Add`/`Mul`/`Neg`/`Zero`/`One` plumbing.

```
pecos_synth::ring::{ZSqrt2, ZOmega, DOmega}
pecos_synth::matrix::{Gate, Matrix}
pecos_synth::synthesis::{exact_synthesize, ExactSynthesis, SynthError}
```

### Decisions worth an opinion

1. **`Gate` vs `pecos_core::Gate`.** This crate's `Gate` is a 9-variant token alphabet (`I X Y Z H S Sdg T Tdg`), not core's payload-carrying struct. Any file importing both must qualify one. Alternatives: `Token`, `SynthGate`, or live with qualification. Note it spells `S`/`Sdg` while core spells `SZ`/`SZdg` -- deliberate (matches the synthesis literature), but it is a second spelling of the same gate in one codebase.

2. **The scalar is a bare `u8`.** `ExactSynthesis::scalar()` and `Matrix::with_global_phase(exponent: u8)` both mean "exponent `j` in `omega^j`, mod 8", but the type says none of that. A `newtype OmegaExponent(u8)` (or at least renaming to `scalar_exponent`) would make misuse (passing a degree, a tick count) a compile error rather than a silent wrong phase. Phase 2 will thread this value through the approximation path, so now is the cheap moment.

3. **`t_count() -> usize`.** The crate's own determinism rule keeps observable quantities at fixed widths (`u32` denominator exponent, never `usize`). T-count is equally observable and lands in fixture files. `u32` would be consistent; `usize` is defensible as "a length". Pick one on purpose.

4. **`exact_synthesize(&Matrix)` is a free function** returning `ExactSynthesis`. Fine as-is; the alternative (`Matrix::synthesize`) reads better at call sites but puts synthesis policy on the matrix type. Phase 2 will add `synthesize_rz(angle, opts)` beside it, so the free-function namespace is arguably the right pattern.

5. **`ExactSynthesis` as a name** describes the operation, not the result. The result is a Matsumoto-Amano normal form; `NormalForm` or `SynthesizedWord` would say what you hold. Renaming after phase 2 exists is more churn than doing it now.

None of these block merging -- they are one-word renames today and breaking changes later. Items 2 and 3 are the two I would actually change.

ciaranra:
## API decisions applied

Following the guide above, the five decision points were settled -- the two mechanical ones by the crate's own rules, the two naming forks via an independent design review, and one left as it was.

| # | decision | outcome |
|---|---|---|
| 1 | `Gate` name + `S` spelling | **Renamed `GateToken`, variants now `SZ`/`SZdg`** (design review recommendation, adopted in full). Kills the `pecos_core::Gate` collision and makes token-to-`GateType` conversion one-to-one; the module docs still name the literature's `S` where the Matsumoto-Amano grammar is stated. |
| 2 | bare `u8` scalar | **`OmegaExponent` newtype**, wrapping constructor so every held value is canonical `0..8`; threads through `Matrix::with_global_phase`, `NormalForm::scalar`, and the enumeration tests. |
| 3 | `t_count() -> usize` | **Now `u32`**, per the crate's own fixed-width rule for observable values. |
| 4 | `exact_synthesize` free function | Kept: phase 2 adds `synthesize_rz` beside it, so the free-function namespace is the pattern. |
| 5 | `ExactSynthesis` name | **Renamed `NormalForm`.** The design review recommended `MatsumotoAmanoNormalForm`; the substance is adopted (name the thing held; keep phase 2's approximate result a deliberately distinct type), with the shorter spelling since the crate context and the type's doc line supply the "Matsumoto-Amano". Stated as a deviation from the review, on purpose. |

Also in this commit: a `# Panics` section on `exact_synthesize` documenting the physically-unreachable `u32` T-count bound, and the ordering comments updated to the new spellings.

Verification unchanged and green: 24 tests (including the enumeration oracle and the canonical-table pin), in-crate `--locked` clippy with `-D warnings`, workspace check, fmt, pre-commit.

Issue #637, opened 2026-08-30
qelib1.inc: ry computes RY(-theta), cry inherits the sign error, and crx is not controlled


ciaranra:
## Summary

Three gate macros in `crates/pecos-qasm/includes/qelib1.inc` compute the wrong unitary. These are the definitions user QASM resolves to, because lowercase gate names are never treated as native (`crates/pecos-qasm/src/parser/native_gates.rs:3-4,37-43` -- "Only uppercase names are considered native gates"), so every lowercase `ry`, `cry`, or `crx` in a user program takes the macro path.

### 1. `ry(theta)` computes `RY(-theta)` (`qelib1.inc:38-42`)

```
gate ry(theta) a {
  rx(-pi/2) a;
  rz(theta) a;
  rx(pi/2) a;
}
```

Circuit order means the matrix is `RX(pi/2) . RZ(theta) . RX(-pi/2)`. Conjugating `Z` by `RX(pi/2)` gives `-Y`, so this is `exp(+i theta Y / 2) = RY(-theta)`. The two `rx` calls are swapped. PECOS's `RY(theta)` is `exp(-i theta Y / 2)` (`gate_matrix_def.py:381-389`), so this is a genuine sign error, not a convention difference.

Numerically (max abs entry difference): `||ry_macro(0.7) - RY(+0.7)|| = 6.9e-01`, `||ry_macro(0.7) - RY(-0.7)|| = 1.2e-16`.

### 2. `cry(theta)` computes `CRY(-theta)` (`qelib1.inc:103-108`)

The macro's own structure (`ry(theta/2); cx; ry(-theta/2); cx`) is the standard one and would be correct given a correct `ry`; it inherits the sign error above. `||cry_macro(0.7) - CRY(+0.7)|| = 6.9e-01`, `||cry_macro(0.7) - CRY(-0.7)|| = 3.3e-16`.

### 3. `crx(theta)` is not a controlled gate at all (`qelib1.inc:94-100`)

```
gate crx(theta) a,b {
  ry(pi/2) b;
  cx a,b;
  ry(theta) b;
  cx a,b;
  ry(-pi/2) b;
}
```

The two `cx` gates cancel on the control=0 block, leaving `ry(-pi/2) . ry(theta) . ry(pi/2) = RY(+-theta)` acting on the target when the control is |0>. It also rotates by the full `theta` rather than `theta/2` per branch. This is broken independently of bug 1.

[...]

Issue #635, opened 2026-08-30
SLR-to-QIR lowering emits controlled-phase for CRZ (and CRX), not controlled-RZ


ciaranra:
## Summary

The SLR-to-QIR lowering of `CRZ` (`python/quantum-pecos/src/pecos/slr/ast/codegen/qir.py:206-215`) emits `RZZ(-theta/2)` plus `RZ(theta/2)` on BOTH qubits. That product is the controlled-phase gate `diag(1, 1, 1, e^{i theta})` up to global phase, not the textbook controlled-RZ `diag(1, 1, e^{-i theta/2}, e^{i theta/2})` that the SLR `CRZGate` docstring promises and that `gate_matrix_def.CRZ` defines. The two differ by a phase on the control qubit, which is relative (observable whenever the control is in superposition), not global.

Checked against PECOS's own oracle matrices (`python/quantum-pecos/tests/pecos/integration/state_sim_tests/gate_matrix_def.py`, theta = 0.7):

```
(RZ(t/2) x RZ(t/2)) . RZZ(-t/2)  == oracle CRZ (up to global phase): False
(I       x RZ(t/2)) . RZZ(-t/2)  == oracle CRZ (up to global phase): True
(RZ(t/2) x RZ(t/2)) . RZZ(-t/2)  == diag(1,1,1,e^{i t})              : True
```

The comment at that site claims verification against `gate_matrix_def.CRZ` for random angles; that claim does not hold with the current symmetric `RZ` convention (#601). `CRX` in the same table is derived from `CRZ` by `H` conjugation and inherits the error. The existing QIR/Guppy `CRZ` tests (`tests/slr_tests/ast_guppy/test_qir_corpus_executable.py`, `tests/guppy/test_crz_angle_arithmetic.py`) evidently exercise only definite-control inputs.

## Fix

Emit the target-only decomposition `RZ(theta/2)` on the target and `RZZ(-theta/2)`; add a test with the control in superposition (`H` on the control first) comparing the executed state against `gate_matrix_def.CRZ` for angles including `-pi`, `pi`, `2pi`; same for `CRX`.

## Context

Found while deciding #626. `CRZ` in PECOS is an interop gate (Guppy/HUGR `CRz`, QIS `crz`, PHIR `CPhase`, SLR `CRZGate`), not a QEC primitive; the proposal under #626 is to accept it only at boundaries and lower it exactly into `RZ`/`RZZ`, which would make this lowering the single place the decomposition lives.

[...]

Issue #622, opened 2026-08-29
HUGR parser maps CRz to RZZ and defaults a missing rotation angle to zero


ciaranra:
## Summary

The HUGR-to-PHIR gate mapping translates the HUGR `CRz` operation to `QuantumOp::RZZ`. Controlled-Rz and RZZ are different gates.

`crates/pecos-phir/src/hugr_parser.rs`:

```rust
"CRz" => Some(QuantumOp::RZZ(Angle64::from_radians(angle.unwrap_or(0.0)))),
```

## Why it matters

`CRZ(theta) = RZ(theta/2)_target . CX . RZ(-theta/2)_target . CX` differs from `RZZ(theta) = exp(-i theta/2 Z (x) Z)` by a local `RZ` on the target (and a global phase). Any HUGR program that uses `CRz` currently simulates a different unitary than it declares. `CRZ` already exists as a `GateType` and as a `QuantumOp` consumer path elsewhere in the engine (`crates/pecos-engines/src/quantum.rs` handles `GateType::CRZ` via `try_crz`, and `crates/pecos-simulators/src/clifford_rotation.rs` implements the decomposition), so the fix is to map to the CRZ op, not RZZ.

## Also on the same lines

`angle.unwrap_or(0.0)` silently substitutes a zero angle when the HUGR op carries no angle for `Rx`/`Ry`/`Rz`/`CRz`. A missing angle on a parameterized gate is a malformed program and should be an error, consistent with the fail-loud policy adopted elsewhere.

## Acceptance

- `CRz` maps to a controlled-Rz operation; a test compares the resulting circuit against the explicit `RZ/CX/RZ/CX` decomposition for a Clifford angle (`pi`, which must equal `CZ`) and a non-Clifford angle.
- Missing angle on a rotation op is a parse error, with a test.

Related: #621

ciaranra:
Fixed by #638 (cdaa9aaad): the PHIR HUGR parser mapped `CRz` to a plain `RZZ`. It now lowers exactly from the source angle, pinned by a full-matrix regression at five angles including the 2pi and 3pi wrap-crossing cases.

PR #638, opened 2026-08-30
Lower controlled rotations from their source angle at every boundary and remove GateType::CRZ

### LINKED ISSUE #626: CRZ is 4pi-periodic but carried as an Angle64 (mod 2pi), dropping Z on the control across the boundary
## Summary

`CRZ(theta)` is 4-pi-periodic in `theta`, but PECOS carries its angle as `Angle64`, which is normalized modulo 2-pi. Every `CRZ` path therefore computes `CRZ(theta mod 2pi)`, which differs from `CRZ(theta)` by a `Z` on the control whenever the reduction crosses a 2-pi boundary.

```
CRZ(theta) = |0><0| (x) I + |1><1| (x) RZ(theta)
CRZ(theta + 2pi) = |0><0| (x) I - |1><1| (x) RZ(theta) = (Z_control) . CRZ(theta)
```

Concretely `CRZ(-pi) = diag(1, 1, i, -i) = (SZ (x) I) . CZ` while `CRZ(pi) = diag(1, 1, -i, i) = (SZdg (x) I) . CZ`; with `Angle64`, `-pi` and `pi` are the same value, so one of the two is executed for both. `CRZ(2pi)` should be `Z_control` but executes as identity.

Affected: `crates/pecos-simulators/src/clifford_rotation.rs` `try_crz` (Clifford path, `RZ(theta/2) . CX . RZ(-theta/2) . CX`), the arbitrary `crz` in `crates/pecos-simulators/src/arbitrary_rotation_gateable.rs` and its implementors, the engine dispatchers in `crates/pecos-engines/src/quantum.rs`, and the Python bindings. The doc line on `try_crz` also claimed `CRZ(pi) = CZ`; that was corrected to `(SZdg (x) I) . CZ` in the Clifford-rebase PR, implementation untouched.

The same 4-pi structure applies to any controlled rotation whose target rotation is 2-pi-periodic only up to sign (`CRX`, `CRY` if added), and to the `U` gate's phase bookkeeping tracked under the arbitrary-gate-set note (#575/#576/#577).

## Options

1. Define PECOS `CRZ(theta)` as the controlled-phase-style gate that IS 2-pi-periodic (`diag(1, 1, 1, e^{i theta})` up to a documented convention) and document that it is not the textbook controlled-RZ. Simplest, but a convention break for anyone reading `CRZ` as controlled-RZ.
2. Keep textbook controlled-RZ and carry the extra sign bit: represent the CRZ parameter as `(Angle64 half_angle, bool wind)` or take `theta/2` as the stored angle (`RZ(theta/2)` is what the decomposition consumes, and `theta/2 mod 2pi` preserves `theta mod 4pi`). Storing the half-angle is the smallest change that keeps the textbook gate exact.

Recommendation: option 2 with the stored half-angle, since it keeps the decomposition exact and the user-facing gate textbook.

## Acceptance

- Matrix-oracle tests for `CRZ` at `pi`, `-pi`, `2pi`, `3pi` on the state-vector simulator and tableau tests on the Clifford path, on control-superposition inputs (an `|1>` control alone cannot see the `Z_control` difference).
- One convention statement in the gate docs.

Related: #621

ciaranra:
## Summary

Closes #626, #622, #635. `CRZ` (and `CRX`/`CRY`, and PHIR `CPhase`) are interop spellings, not PECOS gates: they arrive from Guppy/HUGR (`CRz`), the QIS ABI (`__quantum__qis__crz__body`), PHIR (`CPhase`), SLR (`CRZGate`), QASM, and the Python API, and no QEC path needs them. Selene's own simulator-plugin interface receives only `RXY`/`RZ`/`RZZ` because the compiler lowers `crz` before any simulator sees it, and Guppy's `angle` never wraps modulo full turns. PECOS now does the same: every ingress halves the source `f64` angle and emits native rotations, and `GateType::CRZ` leaves the executed IR.

## Why

PECOS stores angles as `Angle64`, reduced mod 2pi at ingestion. Textbook controlled-RZ is 4pi-periodic (`CRZ(theta + 2pi) = (Z (x) I) . CRZ(theta)`), so every executor that halved the *stored* angle was wrong whenever the reduction had crossed 2pi: `CRZ(-pi)` ran as `CRZ(+pi)` (off by `Z` on the control), `CRZ(2pi)` as identity. Halving the `f64` before normalization and emitting

```
CRZ(theta) = RZZ(-theta/2) . (I (x) RZ(theta/2))
```

is exact for every theta up to a global phase of +-1 (the same global phase every stored PECOS rotation already carries, since `Angle64` reduces mod 2pi while `RZ(theta + 2pi) = -RZ(theta)`); every relative phase, including the `Z_control` factor, is exact. Both factors are ordinary PECOS rotations. Tests compare the full 4x4 matrix under one shared global phase, never per column, so a controlled-phase-shaped error cannot hide. Because of this defect, #621/#628/#631 had deliberately withheld `CRZ` from the Python-facing simulator arms and the neo fallback.

Two live bugs are fixed by the same lowering table: the SLR-to-QIR codegen lowered `CRZ` with `RZ(theta/2)` on both qubits, which is controlled-phase, not controlled-RZ (#635; `CRX` inherited it); the PHIR HUGR parser mapped `CRz` to `RZZ` (#622). PHIR `CPhase(lambda)` was executed as `crz(lambda)`, dropping a control-side phase; it is now lowered exactly as `(U(0,0,lambda/2) (x) RZ(lambda/2)) . RZZ(-lambda/2)` (`U(0,0,x) = e^{ix/2} RZ(x)` carries the `e^{i lambda/4}`, so no scalar term is needed).

## Changes

- `pecos-core`: one lowering module (`lower_crz` / `lower_crx` / `lower_cry` / `lower_cphase`) taking radians as `f64`, documented with the identities and the reason halving must precede normalization.
- Noise and scheduling semantics change deliberately: `CRZ` was one two-qubit noise location; the lowered form is one `RZZ` plus one `RZ` in consecutive ticks, which is what hardware runs. Gate counts and traces change accordingly.
- Every ingress calls it: HUGR (`hugr_convert`, and the PHIR HUGR parser, which also stops defaulting a missing rotation angle to zero), the QIS conversion in `ccengine` (ABI symbol unchanged; `crz` previously fell through to "Unsupported operation" there), QIS trace replay (trace files unchanged), PHIR `CPhase` execution, SLR codegen (`QuantumCircuit`, QIR, QASM; the Guppy target keeps emitting Guppy's own `crz`); SLR's `CRXGate`/`CRYGate`/`CRZGate` take their angle as a real number in radians and reject a typed `Angle` with a `TypeError`, because a typed `Angle` has already been reduced modulo 2pi and controlled rotations are 4pi-periodic -- the same rule the Rust constructors follow, QASM conversion where it mapped `crz`, the Python API and pyo3 constructors (`crz(theta_radians: float)`), the neo Python binding, `hugr_to_dag.py` (`CRx`/`CRy`/`CRz`/`CU1`), and the `zlup` parser (which had the same `crz -> RZZ` mapping as #622).
- Removed: `GateType::CRZ` (numeric id 70 retired, not reused; a stale ByteMessage payload with id 70 now fails to decode), `CliffordRotation::try_crz`, the `ArbitraryRotationGateable::crz`/`crx`/`cry` defaults, the engine/executor/neo arms, the noise-group entries, the DEM builder's zero-angle special case, and the `CRX`/`CRY`/`CRZ` entries in the simulator spelling table (the spellings remain accepted where they were, via lowering).
- Docs: naming conventions state the rule (controlled rotations are boundary spellings lowered exactly to the rotation family); gate reference updated; `gate_matrix_def.CRZ` stays as the test oracle.
- The QIR corpus test that asserted a deterministic outcome for `H(c); CRZ(pi); H(c)` pinned the controlled-phase behavior; it is rewritten against the textbook amplitudes.
- Tests: every ingress is pinned by a control-superposition oracle test at theta in `{-pi, pi/3, pi, 2pi, 3pi}` (a definite control cannot see the `Z_control` difference, which is why #635 went unnoticed); Clifford-angle behavior on stabilizer simulators; #622 and #635 regressions; gate-id-70 decode rejection.

- Test-driven surface additions: `StateVectorEngine::{simulator, simulator_mut}` (amplitude-level Rust boundary tests), `pecos_rslib_exp.neo_fallback_native_gates` (exposes the legacy-to-neo fallback circuit), and `zluppy.compile_to_phir_json` (parser-to-PHIR regression). `zlup` stays dependency-free, so it halves the source angle locally rather than calling `pecos-core`; the same five-angle matrix regression pins it.
- The HUGR-to-DAG conversion now wires each gate to the previous gate on its qubits (`add_gate_auto_wire`) instead of a separate HUGR-wire edge pass; the traversal is already topological (`check_predecessors_ready`), so the edges are identical, and a staggered-chain test pins that.

## Not in this PR

- A first-class `Phase(gamma)` gate (one rule: `e^{i gamma}` on the all-ones subspace of its operands; `{}` = global phase, `{q}` = `U(0,0,gamma)`, `{c,t}` = controlled-phase), which would let `T` and `CPhase` definitions carry their global phase explicitly. Follow-up; design in the vault note `design/crz-parameter-representation.md`.

## Not covered: the QASM macro path

Lowercase gate names never resolve as native (`crates/pecos-qasm/src/parser/native_gates.rs:3-4`), so user `crz`/`crx`/`cry` in QASM expand through `qelib1.inc` rather than through this lowering. `crz` and `cphase` are exact there (verified up to a global sign at theta in {pi/3, pi, 2pi, 3pi, 6pi}), but reviewing this branch turned up three independent pre-existing defects in that include -- `ry` computes `RY(-theta)`, `cry` inherits the sign, and `crx` is not a controlled gate at all -- filed as #637 and deliberately not fixed here (different defect class, its own tests and review). The gate documentation names this exception rather than claiming universal lowering.

## Merge order

This branch and the `zluppy` test-revival branch both touch `exp/zlup/src/{ast.rs,parser.rs,codegen/slr.rs}` and `exp/zluppy/{src/lib.rs,tests/test_zluppy.py}`. They do not conflict semantically -- this PR makes `crz` its own `GateKind::CRZ` boundary spelling, that PR makes `GateKind::ALL`/`keyword()`/`ordinal()` the single canonical spelling table -- but whichever lands second needs a rebase. If this one lands first, the other adds `CRZ` to the three tables (keyword `"crz"`) and drops its `crz -> RZZ` alias special case; `gate_name_to_mapping("crz")` already returns `TketOp::CRz`, so that PR's mapping-drift guard still passes.

## Verification

Cold in a fresh worktree: `cargo test` over the twelve touched crates, the doc-test crate, workspace clippy `-D warnings` (files touched first), `cargo fmt --check`, the fast Python lane plus the Guppy and SLR suites, `pecos-rslib` tests, `pre-commit run --all-files`; a classified grep receipt of every remaining `CRZ`/`crz` spelling (ingress / ABI carrier / test oracle / accepted-and-lowered spelling). Reviewed by a read-only pass of the task packet against the tree, hunk review, and a fresh-context review of the finished tree.


ciaranra:
The optional-dependency lane (`just pytest-dep`, which carries this PR's QIR corpus probes for #635) only runs post-merge, so PR checks cannot show it. Dispatched it on this branch: https://github.com/PECOS-packages/PECOS/actions/runs/33340785239 -- the whole run, including `python-slow-postmerge`, succeeded. That run predates the final commit by two documentation files only (`git diff 703b4dada c9ee63284` = `docs/development/parallel-blocks-and-optimization.md`, `docs/user-guide/parallel-execution.md`), so it is valid evidence for the test code.

Those two docs were the one thing PR CI caught that local verification missed: they passed a typed `Angle` to `CRZ`, which the new parameter guard correctly rejects. Fixed by passing radians directly; `just docs-test` now reports 305 passed, 0 failed.

ciaranra:
Second independent review, targeting the round-7 work and the rebase, returned REVISE. Both must-fixes and four follow-ups are in.

**The good news first:** the reviewer verified that #632 (landed on `dev` mid-work, moving Clifford snapping to explicit entry points) does not break this lowering. Every lowered rotation reaches a `*_snapped` entry point; old and new snap on the *identical* source window, because both snap the halved angle; and `lower_crz`/`lower_cphase` land exactly on Clifford fractions for pi multiples, so #632's exact-matching consumers see them as Clifford too. The rebase conflict resolution lost nothing from #632. The QIR probes killed 9 of 12 mutations, the three survivors being semantically equivalent mutants (symmetric RZZ operands, a +2pi target shift, and a mod-4pi source reduction).

Fixed:

- **The pyo3 `TickHandle.crz` was not atomic.** It re-implemented the lowering instead of calling the Rust `try_crz` added in the previous round, inserting the `RZZ` before attempting the `RZ` in the next tick -- so a conflict raised but left a stray `RZZ`, i.e. an `RZZ` where a `CRZ` was requested. Reproduced, then fixed by delegating; the regression test asserts both that the error is raised and that the first tick is untouched.
- **`docs/user-guide/gates.md` prose split the two-qubit gate table**, so every row from `SWAP` down would have rendered as literal pipe-delimited text. Table is contiguous again with the prose after it.
- Lowered controlled rotations had no stabilizer-simulator coverage through the ByteMessage engine dispatch after the old `try_crz` tests were removed; added.
- The analysis DAGs (`SimpleHugr` and `hugr_to_dag.py`) attached a lowered `CRz`'s successors to the target-only `RZ`, giving a gate on the *control* qubit a false dependency and losing a parallel-scheduling opportunity. Reproduced before fixing; both are now port-aware, which removes the same defect for `CU1`.
- Dead `node_to_dag_idx` state and a node-zero placeholder left by the auto-wire rewrite; removed.
- A dynamic-angle `crz` reported `RZZ at Node(..)`, naming a gate the user never wrote; it now uses the source spelling.

Cold verification: 10/10 steps exit 0 -- 16-crate `cargo test`, doc-test crate, workspace clippy `-D warnings`, fmt, 2831 + 937 + 2277 pytest, pre-commit, `just docs-test` (305 passed), and the slow+optional QIR corpus (31 passed).

ciaranra:
Third review returned SHIP -- no blockers, no majors. It verified the riskiest part properly: all four Python port tables checked against the actual gate order in `controlled_rotations.rs` (including `CU1`, where the `U` sits on the control and the `RZ` on the target -- the easy one to get backwards, and it is right), and six circuits executed through `hugr_to_dag` confirming every dependency edge lands on the correct gate per qubit. Execution order is unchanged; the port-awareness fix is analysis-only.

Its four minors are now fixed:

- **A behavior regression the previous round introduced.** Delegating to `try_crz` dropped the empty-pairs guard, so `crz(theta, [])` clobbered `last_gate_idx` and a following `.meta()` silently became tick-level instead of gate-level. Restored by assigning only when an index exists; regression test added. Verified: the attr now lands on the gate, with nothing at tick level.
- The `.meta()`-targets-the-`RZZ` invariant is now pinned by a test, so a future refactor routing the `RZ` through a `TickHandle` cannot silently corrupt metadata on an unrelated batch.
- A `wire_to_qubit` miss silently dropped a dependency edge. Fixed at the source -- the fallback branch now records its minted `QubitId` like the linked branch does -- rather than tolerating the miss, with a dangling-port regression.
- The new engine test killed the qubit-swap mutation only about 3 times in 4 against an entropy-seeded RNG, i.e. as a flake rather than a failure. The control is now prepared in `|1>` so it is deterministic; confirmed by mutation, both the moved-`RZ` and flipped-`RZZ`-sign cases now fail cleanly.

Cold verification: 10/10 steps exit 0 -- 16-crate `cargo test` (8,379 passed), doc-test crate, workspace clippy `-D warnings`, fmt, 2831 + 937 + 2279 pytest, pre-commit, `just docs-test` (305 passed), slow+optional QIR corpus (31 passed).

Issue #636, opened 2026-08-30
zlup HUGR backend rejects gate-target forms and comptime angles that the SLR backend accepts


ciaranra:
## Summary

The zlup HUGR backend (`exp/zlup/src/codegen/hugr.rs`) lowers a strict subset of the gate-target forms the SLR backend accepts. Each of the following compiles to SLR and fails `compile_to_hugr` (reproduced on branch `zluppy-tests-current-syntax`, which is where current-syntax gate statements first reach the HUGR backend at all):

| Source | SLR | HUGR |
|---|---|---|
| `h q;` (allocator broadcast; only `pz q;` is special-cased) | ok | `HUGR error: unsupported expression in codegen` |
| `h [q[0], q[1]];` (bracket-array batch; `is_batch_literal` accepts only `{...}` and `&[...]`) | ok | `HUGR error: unsupported expression in codegen` |
| `crz(pi_2) (q[1], q[0]);` with `pi_2` a named comptime constant (`extract_angle` resolves only `pi`/`tau` identifiers) | ok | `HUGR error: rotation angle must be a numeric literal` |

The third case means the repository's own `exp/zlup/examples/qft_3qubit.zlp` fails HUGR codegen while every other example compiles.

## Expected

The HUGR backend accepts every target form and angle expression the SLR backend accepts (broadcast to a whole allocator, bracket-array batches, comptime constant resolution), or the language documentation states the HUGR-specific restriction and the example is changed to conform. The first reading is the consistent one: both backends lower the same AST.

## Notes

- Before the zluppy suite revival these programs did not error: the HUGR backend ignored `Stmt::Gate`/`Stmt::Prepare` entirely and compiled them to an identity circuit. Failing loudly is the improvement; this issue is about closing the remaining gap.
- Codegen-level tests that assert the emitted operation list (`collect_program` + `operations`) are the right place to pin each form; byte-length assertions on the HUGR blob cannot distinguish a lowered circuit from an empty one.

PR #639, opened 2026-08-30
Restore the zluppy test suite to current zlup syntax and give it a CI lane

ciaranra:
## Summary

The `zluppy` Python test suite (`exp/zluppy/tests/test_zluppy.py`) was 41 failed / 29 passed on `dev`, and nothing ran it: no workflow or Justfile recipe invokes it (the Python workflow's path filters do not include `exp/**`), so the suite rotted as the zlup language moved on. This PR brings it back to green and gives it a post-merge CI lane.

## What was wrong

- 26 tests wrote zlup source in a syntax the language no longer accepts (`var q = qalloc(2); h(q[0]); cx(q[0], q[1]);`). Current zlup uses `q := qalloc(2);`, DSL gate statements (`h q[0];`, `cx (q[0], q[1]);`), `pz q;`, `result: u1 = mz(u1) q[0];`, and `pub fn main() -> unit { ...; return unit; }` (see `exp/zlup/examples/bell_state.zlp`).
- 6 tests asserted the old rendering of `ZlupProgram.to_source()`; the pretty-printer has rendered current syntax for a long time.
- 9 tests failed on a real binding bug: `ZlupProgram.build_ast()` emitted `main` with a `unit` return type and no return statement, which zlup's semantic analysis rejects (`missing return statement in function 'main'`).

Reviving the tests then exposed two `zlup` backend bugs that the old syntax had hidden: the HUGR backend did not lower current-syntax gate statements (`h q[0];`, `pz q;`) at all, so such programs compiled to identity, and the SLR backend only accepted two-element qubit tuples, so the documented `ccx (q[0], q[1], q[2]);` failed.

The HUGR backend now also fails loudly on a gate it has no mapping for instead of dropping it; the remaining forms it does not lower (allocator broadcast for non-`pz` gates, bracket-array batches, named comptime angle constants -- which is why `exp/zlup/examples/qft_3qubit.zlp` fails HUGR codegen) are tracked in #636.

## Changes

- `exp/zlup/src/codegen/hugr.rs`: `Stmt::Gate`, `Stmt::Prepare`, slot references, and angle literals lower through the existing named-gate path, with codegen tests asserting the emitted operation list for each form (byte-length checks on the HUGR blob cannot tell a lowered circuit from an empty one); `GateKind::keyword()`/`GateKind::ALL` are the one canonical spelling shared by parser, pretty-printer, suggestions, and HUGR codegen, with a round-trip test; `exp/zlup/src/codegen/slr.rs`: tuple lowering generalized to the gate's arity.
- `exp/zluppy/Cargo.toml` / `pyproject.toml`: pyo3 extension-module linking moved behind a maturin-enabled feature (the `pecos-rslib` pattern) so plain `cargo test` works for the crate's unit tests.
- `exp/zluppy/src/lib.rs`: the builder appends the explicit unit return that zlup requires, using the AST the parser produces for `return unit;` (the pretty-printer keeps rendering it as the `return;` shorthand); Rust test on the builder path.
- `exp/zluppy/tests/test_zluppy.py`: sources rewritten in current zlup; rendering expectations updated to the pretty-printer's output; assertions kept as specific as before.
- `Justfile`: `pytest-zluppy` recipe (syncs the `exp/zluppy` project env and runs its suite).
- `.github/workflows/python-test.yml`: `zluppy-postmerge` job patterned on `python-slow-postmerge` (post-merge push and `workflow_dispatch`), `exp/zlup/**` and `exp/zluppy/**` added to the path filters, and `uv lock --check --project exp/zluppy` on the PR fast path.

Judgment call: `zluppy` is its own uv project on its own version train, so its environment rebuilds `quantum-pecos` and the Rust extensions. Running that only post-merge (with a dispatch escape hatch) matches how the slow lane is treated and keeps the PR path unchanged; making it a PR blocker would be a separate decision.

## Merge order

This branch and the CRZ boundary-lowering branch (#626) both touch `exp/zlup/src/{ast.rs,parser.rs,codegen/slr.rs}` and this package. They do not conflict semantically, but whichever lands second needs a rebase. If #626 lands first, this branch adds its `GateKind::CRZ` to `ALL`/`keyword()` (`"crz"`)/`ordinal()` and drops the `crz -> RZZ` alias special case here; the mapping-drift guard still passes because `gate_name_to_mapping("crz")` already returns `TketOp::CRz`.

## Verification

Full zluppy suite green; `cargo test`/`clippy -D warnings`/`fmt` on `zluppy` and `zlup`; `uv lock --check --project exp/zluppy`; `pre-commit run --all-files`.

## Also included: wasmtime 46.0.2 -> 46.0.3

RUSTSEC-2026-0268 and RUSTSEC-2026-0269 (Wasmtime) were published on 2026-08-31, after `dev`'s last clean `cargo-deny` run, and both require `>=46.0.3`. They are not caused by this PR, but this PR touches `exp/zluppy/Cargo.toml`, so `cargo-deny` gates it -- and the repo's policy is to fix a pre-existing advisory on the current PR only when it blocks that PR's checks. `dev` is affected identically and will need this bump regardless.

`wasmtime` is reached only through `crates/pecos-wasm`. The plain `cargo update` also re-normalized 25 unrelated dependency edges among duplicate versions (`windows-sys`, `unicode-width`, `hashbrown`, `indexmap`), so the lockfile change here was rebuilt as version+checksum substitutions only -- zero non-version lines differ. Verified: `cargo tree --locked` resolves, `cargo check --locked -p pecos-wasm` builds, and `just cargo-deny` reports advisories, bans, and sources all ok.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -23,6 +23,8 @@ on:
       - 'python/**'
       - 'examples/**'
       - 'crates/**'
+      - 'exp/zlup/**'
+      - 'exp/zluppy/**'
       - 'Cargo.toml'
       - 'Cargo.lock'
       - 'pyproject.toml'
@@ -40,6 +42,8 @@ on:
       - 'python/**'
       - 'examples/**'
       - 'crates/**'
+      - 'exp/zlup/**'
+      - 'exp/zluppy/**'
       - 'Cargo.toml'
       - 'Cargo.lock'
       - 'pyproject.toml'
@@ -228,7 +232,9 @@ jobs:
 
     - name: Validate Python lockfile (PR fast path)
       if: github.event_name == 'pull_request'
-      run: uv lock --check --project .
+      run: |
+        uv lock --check --project .
+        uv lock --check --project exp/zluppy
 
   python-compat-smoke-build:
     runs-on: ${{ matrix.os }}
@@ -669,6 +675,87 @@ jobs:
 
     - name: Run slow Python tests
       run: just pytest-slow
+
+  zluppy-postmerge:
+    if: (github.event_name == 'push' && contains(fromJSON('["main", "master", "development", "dev"]'), github.ref_name)) || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    env:
+      PECOS_BUILD_MWPF: "1"
+    steps:
+    - name: Harden the runner (egress audit)
+      uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+      with:
+        egress-policy: audit
+
+    - name: Free Disk Space (Ubuntu)
+      uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be # v1.3.1
+      with:
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: false
+        docker-images: true
+        tool-cache: false
+        swap-storage: true
+
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      with:
+        persist-credentials: false
+
+    - name: Set up Python
+      uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+      with:
+        python-version: "3.14"
+
+    - name: Install the latest version of uv
+      uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+      with:
+        version-file: ".github/uv.toml"
+        enable-cache: true
+        save-cache: ${{ github.event_name == 'push' && contains(fromJSON('["main", "master", "development", "dev"]'), github.ref_name) }}
+
+    - name: Set up Rust
+      run: |
+        bash scripts/ci/ensure-rust.sh stable minimal
+        export PATH="$HOME/.cargo/bin:$PATH"
+        rustup show
+
+    - name: Verify cmake is available (MWPF decoder build)
+      run: cmake --version
+
+    - name: Install just
+      uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4
+
+    - name: Cache Rust
+      uses: Swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+      with:
+        cache-bin: false
+        prefix-key: v1-rust-no-bin
+        save-if: ${{ github.event_name == 'push' && contains(fromJSON('["main", "master", "development", "dev"]'), github.ref_name) }}
+        workspaces: |
+          python/pecos-rslib
+          python/pecos-rslib-llvm
+
+    - name: Cache LLVM ${{ env.LLVM_VERSION }}
+      id: cache-llvm-zluppy
+      uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+      with:
+        path: ~/.pecos/deps/llvm-21.1
+        key: llvm-${{ env.LLVM_RELEASE_VERSION }}-${{ runner.os }}-${{ runner.arch }}-v3
+
+    - name: Ensure LLVM ${{ env.LLVM_VERSION }}
+      run: just ci-env
+
+    - name: Save LLVM ${{ env.LLVM_VERSION }} cache
+      if: steps.cache-llvm-zluppy.outputs.cache-hit != 'true' && github.event_name == 'push' && contains(fromJSON('["main", "master", "development", "dev"]'), github.ref_name)
+      uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
+      with:
+        path: ~/.pecos/deps/llvm-21.1
+        key: ${{ steps.cache-llvm-zluppy.outputs.cache-primary-key }}
+
+    - name: Run zluppy Python tests
+      run: just pytest-zluppy
 
   windows-vanilla-cargo:
     name: Windows vanilla cargo (crates.io consumer contract)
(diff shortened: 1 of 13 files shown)


ciaranra:
The `zluppy-postmerge` job skips on pull requests by design (it mirrors `python-slow-postmerge`), so this PR's own checks cannot demonstrate the lane it adds. Dispatched it manually on this branch to get real CI evidence: https://github.com/PECOS-packages/PECOS/actions/runs/33339683630 -- `zluppy-postmerge` succeeded, so the recipe and the revived suite both work on a clean CI checkout, not only locally.

ciaranra:
Second independent review of the round 2-4 work returned REVISE; all four must-fixes are in, and two are mutation-proven rather than asserted.

- **Workspace Rust lane regression (blocker).** Dropping `test = false` put `zluppy-python`'s lib test into `cargo test --workspace` (what `pecos rust test` runs), where it links libpython and fails on any runner without a real Python -- e.g. the post-merge macOS lane, which has no `setup-python` step. Restored `test = false` and removed the redundant Rust unit test (the Python suite already covers that fix through `SemanticAnalyzer`). Verified in a clean environment: `env -u PYO3_PYTHON -u LD_LIBRARY_PATH cargo test -p zluppy-python --no-run` exits 0 with nothing linked. The `extension-module` feature gating stays: without it, cargo feature unification leaks `extension-module` into `pecos-rslib`'s test targets.
- **The `GateKind::ALL` guard was inverted.** It anchored on `GateKind::PZ.ordinal() + 1`; with `PZ.ordinal() == 33` and `ALL.len() == 34`, adding a variant after `PZ` and forgetting `ALL` left both sides at 34 (test passes, gate silently unparseable), while adding it correctly made them 35 vs 34 (test fails on a correct change). Replaced with a `gate_kinds!` macro that generates the enum, `ALL`, and `keyword()` from one list, so omission is a compile error; `ordinal()` is deleted.
- **Reversing multi-qubit target order survived all 41 HUGR tests**, so a builder-built `CX` could have emitted `CX(q1, q0)` silently. Added exact ordered `QubitRef` assertions; the mutation now fails.
- **`HugrError::UnknownGate` was unreachable for gates** and only ever fired on user function calls, reporting `unknown gate 'helper'`. Split into an accurate error for the non-gate-call path.

Also strengthened: the Bell assertion (`shot[0] == shot[1]` was satisfied by the all-zero state), and `collect_prepare_op`'s whole-allocator and out-of-bounds branches, both mutation-tested.

Cold verification: 10/10 steps exit 0 -- clean-environment `zluppy-python` and lane-equivalent workspace builds, 70 zluppy tests, `cargo test -p zlup --features cli` and the HUGR codegen suite, clippy `-D warnings`, fmt, `uv lock --check`, pre-commit.

ciaranra:
Third review found that the round-5 macro fix closed only one of several hand-maintained gate-keyword lists. Fixed.

The `gate_kinds!` macro made the enum, `ALL`, and `keyword()` impossible to desynchronize -- but `exp/zlup/src/zluppy.pest` hardcodes `param_gate_keyword`/`simple_gate_keyword` independently, and the guard test only exercised `parse_gate_keyword` (a string lookup over `ALL`), never pest. So adding a variant to the macro would update `ALL`, `keyword()`, and prompt the HUGR-mapping test, while real source using the new gate still failed to parse -- the same hazard, one layer down. `semantic::get_gate_info` and `slr::get_gate_info` had the same blind spot.

The round-trip guard now parses every `keyword()` through the grammar and asserts full-input consumption, and both `get_gate_info` tables have `ALL`-driven exhaustiveness assertions. Verified independently from the other direction: mutating the grammar (`"iswap"` -> `"iswapZZ"`) makes the test fail with `grammar rejected gate keyword 'iswap'`, and the tree restores byte-identically.

Also: `test_controlled_rotation` asserted only `num_nodes() >= 3`, and the round-4 syntax rewrite had silently retargeted it from `TketOp::CRz` to the `rzz` decomposition without any assertion noticing. It now asserts the emitted operation list, and a comment records that #638 makes `crz` a distinct `GateKind::CRZ`, so the expected update at rebase time is obvious.

Cold verification after the change: 10/10 steps exit 0 (clean-environment `zluppy-python` and lane-equivalent workspace builds, 70 zluppy tests, `cargo test -p zlup --features cli` 919 passed, HUGR codegen 45 passed, clippy `-D warnings`, fmt, `uv lock --check`, pre-commit).

ciaranra:
Rebased onto `dev` now that #638 has merged (`cdaa9aaad`). Both conflicts were where the merge-order note predicted, in `exp/zlup/src/{ast.rs,parser.rs}`:

- `ast.rs`: kept the `gate_kinds!` macro form and added `CRZ => "crz"` in dev's enum position.
- `parser.rs`: dropped **both** sides of the keyword-lookup conflict -- dev's hand-written match and this branch's `crz -> RZZ` alias -- since `CRZ` is now a canonical variant and the `ALL`-driven lookup resolves it. Kept **both** tests from the second conflict: dev's `crz`/`rzz` distinctness check and this branch's grammar-aware round-trip guard.

The interesting part is what happened next. The resolution compiled cleanly, but two tests failed:

1. `test_all_gate_kinds_have_hugr_mappings` -> `missing HUGR mapping for crz`. Round 6 had removed `"crz" => Direct(TketOp::CRz)` as provably dead code, correctly: nothing could reach it while `crz` aliased to `rzz`. Merging #638 made it live again. Restored, matching dev's mapping (arity 2, parameterized).
2. `test_crz_alias_lowers_as_rzz_decomposition` failed, carrying the comment round 6 was asked to leave: *"PR #638 gives CRZ its own GateKind and exact lowering. Until those branches meet..."*. It is now `test_crz_lowers_to_tket_crz`, asserting the single `TketOp::CRz` emission.

Without the `ALL`-driven exhaustiveness guard added in round 6, `crz` would have silently stopped lowering in the HUGR backend -- the same class of identity-circuit bug this branch was opened to fix, reintroduced by a merge rather than an edit.

Cold verification after the rebase: 10/10 steps exit 0 (72 zluppy tests, clean-environment `zluppy-python` and lane-equivalent workspace builds, `cargo test -p zlup --features cli`, HUGR codegen, clippy `-D warnings`, fmt, `uv lock --check`, pre-commit). One `cargo fmt` miss from the hand resolution -- a stray blank line -- was caught by the fmt gate and fixed before pushing.

Issue #566, opened 2026-08-20
Unpin macOS python-compat-smoke Rust toolchain once the rustc 1.98 aarch64-darwin segfault is fixed


ciaranra:
rustc 1.98.0 (aarch64-apple-darwin) segfaults deterministically in LLVM target initialization (`initialize_available_targets`) when maturin probes target information with `rustc - --print=cfg ... -C link-arg=-undefined -C link-arg=dynamic_lookup`. Reproduced twice on macOS-latest runners in the `python-compat-smoke-build` job (e.g. https://github.com/PECOS-packages/PECOS/actions/runs/32411062129/job/96561348226). Clean A/B: the same job passed at 2026-08-20 15:12 UTC on rustc 1.97.1 and began failing at 19:52 UTC when runners picked up 1.98.0. Plain cargo builds on the same runners are unaffected; only the maturin probe invocation crashes.

The macOS smoke lane is pinned to 1.97.1 as a stopgap (this repo otherwise deliberately floats latest stable). Remove the pin once a fixed rustc (1.98.x or later) no longer segfaults under the maturin probe on aarch64-apple-darwin.

ciaranra:
Correction to the original report: the crash is **nondeterministic**, not deterministic. Observed crashing twice (2026-08-20 19:52 and ~20:39 UTC) and then surviving on a third runner with the **identical runner image version (20260707.563)** at 2026-08-21 00:11 UTC. The workflow's canary step now runs the probe five times per job and reports the crash count; unpin only after the zero-crash warning has appeared consistently across several PRs.

ciaranra:
The unpin attempt (PR #574, first commit) was falsified immediately: the segfault returned on the first real unpinned macOS build, while the canary had reported survived-5/5 on four consecutive runs. Root cause of the false signal: the canary hand-replicated the rustc invocation, but the real crash occurs in cargo's target-info probe under maturin's link-arg RUSTFLAGS — the replica differed in argv and environment, and 20/20 replica survivals predicted nothing. PR #574 now keeps the pin and replaces the canary with a cargo-driven probe (`cargo +stable check` under the same RUSTFLAGS on a throwaway crate). New unpin criterion: persistent zero-crash warnings from the faithful canary AND a passing unpinned build on a draft PR.

ciaranra:
[...]

PR #640, opened 2026-08-31
Pin the rust-test macOS lane to Rust 1.97.1 with a self-expiring canary

ciaranra:
## Summary

The `rust-test` macOS matrix lane fails on current stable: rustc 1.98.0 on `aarch64-apple-darwin` segfaults compiling ordinary dependency build scripts (`libc`, `proc-macro2`, `quote`) before any workspace crate builds. This is the same 1.98 aarch64-darwin crash that already pinned the `python-compat-smoke` macOS lane (issue #566); the `rust-test` lane was never pinned because its macOS job only runs on dispatch and pre-merge matrix runs.

Reproductions (identical signature, two dispatch runs on one branch plus `dev` itself):
- https://github.com/PECOS-packages/PECOS/actions/runs/33217804443 (branch)
- https://github.com/PECOS-packages/PECOS/actions/runs/33259733605 (branch, re-run)
- `dev` matrix runs on 2026-08-30 failed the same job the same way.

## Change

- The Unix Rust setup step selects 1.97.1 on `macos-latest` and stable elsewhere (expression-level conditional; a matrix `include` would add a phantom macOS job on pull requests, where the matrix is ubuntu-only).
- A self-expiring canary step (macOS, non-PR events only, never fails the job) probes latest stable with five plain `cargo check` runs — this lane's actual crashing invocation, unlike the smoke lane's maturin-link-args probe — and emits a workflow warning when the pin looks obsolete, mirroring the smoke-lane pattern and its unpin discipline (issue #566).

The canary block intentionally duplicates the smoke lane's shape rather than extracting a shared script: the two probes differ in the invocation under test, and this keeps the change to one workflow.

## Verification

- Workflow YAML parses; repo pre-commit hooks pass on the file.
- A `workflow_dispatch` matrix run on this branch is the live proof: the macOS job should install 1.97.1 and pass where the two runs above segfaulted. Run: https://github.com/PECOS-packages/PECOS/actions/runs/33347969025

Unpin path is unchanged and tracked in #566: remove the pin only after the canary warning persists across several PRs and an unpinned build passes on a draft PR.


ciaranra:
Live proof complete: the workflow_dispatch matrix run on this branch is fully green — rust-test (macos-latest) passes with the 1.97.1 pin (https://github.com/PECOS-packages/PECOS/actions/runs/33347969025), where the two unpinned runs cited in the description segfaulted in dependency build scripts. ubuntu and windows lanes unaffected and green.

ciaranra:
The pin itself is well justified and the conditional is the right shape. One must-fix in the canary.

## The canary cannot observe the crash it is probing for

The probe builds a bare project:

```
cargo +stable init --lib --name pin_probe "$probe_dir"
cargo +stable check --manifest-path "$probe_dir/Cargo.toml"
```

`cargo init --lib` emits a manifest with an empty `[dependencies]` section. I ran it: `cargo check` on that project compiles exactly **one** unit, the empty crate itself, and **zero build scripts**.

The failure this pin exists for is, in the PR's own words, `rustc 1.98.0 on aarch64-apple-darwin` segfaulting **compiling ordinary dependency build scripts (`libc`, `proc-macro2`, `quote`)**. The probe compiles none of those, so it cannot reproduce the crash under any toolchain.

The consequence is the opposite of the stated intent. The step will record 0 crashes out of 5 on its very first run and emit:

> Rust toolchain pin may be obsolete: rustc ... survived 5/5 cargo check probes on this runner.

while the pin is still genuinely required. A canary meant to stop the pin going stale silently instead argues for removing it from day one, and it will keep doing so no matter how broken stable is. The five repetitions and the `rm -rf target` between them do not help: five compiles of nothing are still nothing.

Worth noting because the "nondeterministic, so probe repeatedly" reasoning in the comment is sound, and it is what makes the empty probe look convincing rather than obviously wrong.

## Fix

The probe needs to compile the crates whose build scripts crash. Either give the temporary project those dependencies:

```toml
[dependencies]
libc = "0.2"
proc-macro2 = "1"
quote = "1"
```

or point the probe at the real workspace with the pinned toolchain overridden, for example `cargo +stable check -p pecos-core`, which exercises the actual dependency graph and needs no synthetic manifest at all. The second is closer to "this lane's actual crashing invocation", which is what the PR says it is aiming for.

Whichever is chosen, the check worth doing before merge is the one I could not do from here: confirm on a macOS runner that the probe **does** report crashes against 1.98 stable. A canary that has never been seen to trip is indistinguishable from one that cannot.

## The rest reads correctly

- Selecting the toolchain by expression rather than a matrix `include` is right, and the reason given (a phantom macOS job on pull requests, where the matrix is ubuntu-only) is a real hazard.
- Gating the canary on `github.event_name != 'pull_request'` and never failing the job is appropriate for an advisory probe.
- Duplicating the smoke lane's shape rather than extracting a shared script is defensible when the invocations under test genuinely differ, which they do. That difference is exactly the point of this lane's probe being a plain `cargo check` rather than a maturin link-args probe -- which makes it more important, not less, that the plain `cargo check` actually builds something.

qciaran:
Closing as superseded by #783, which fixes the macOS LLVM library search-path conflict underlying the rustc crash and adds a blocking regression check. The regression was demonstrated red in CI, then the probe and both Python extension builds passed with Rust 1.98.1 after the fix. Final validation of #783, including the full macOS Rust-test lane, remains outstanding.

ciaranra:
Superseded by #783.

The macOS `rust-test` failures were not a toolchain bug. PECOS's `DYLD_LIBRARY_PATH` export made rustc load Homebrew LLVM 21 instead of its bundled LLVM 22. #783 switched to `DYLD_FALLBACK_LIBRARY_PATH`, removed the 1.97.1 workaround, and replaced the warning-only canary with a blocking stable Cargo probe that runs in the real build environment.

Dev now passes the macOS lane on stable 1.98.1 with no pin in `rust-test.yml` (runs 35155378810 and 35163052129), so this pin and its canary are no longer needed. The open review item here, that the canary compiled no build scripts and therefore could not observe the crash it probed for, is moot for the same reason.

Issue #641, opened 2026-08-31
StateVecSoA::g operates on stale amplitudes: it never flushes the pending single-qubit gate queue


ciaranra:
## Summary

`StateVecSoA::g` (crates/pecos-simulators/src/state_vec_soa.rs, ~line 3999) reads and writes `self.real`/`self.imag` directly but never calls `flush_two_qubit`, unlike every sibling two-qubit gate (`cx` ~2963, `cz` ~3169, `iswap` ~3918, `sxx` ~3268). When single-qubit gates are queued, `g` silently operates on stale amplitudes and produces a wrong state with no error.

## Reproduction (executed)

2 qubits, `h(0); rz(0); ry(0)` then `g(0,1)`:

```
g with pending queue : [0.228-0.046i, 0.654+0.133i, 0.654-0.133i, -0.228-0.046i]
g after forced flush : [0.228-0.133i, 0.654-0.046i, 0.228+0.133i,  0.654+0.046i]
cz-h-h-cz decomposition (reference): matches the flushed result exactly
```

With a fresh (unqueued) state, native `g` matches the `cz-h-h-cz` decomposition to machine precision — the gate matrix is right; only the flush is missing.

## Scope

Found by an independent review probe while auditing the density-matrix conjugation fix; unrelated to that change (`DensityMatrix::g` inherits the trait `cz-h-h-cz` decomposition and is correct). The failure mode is the silent-wrong-state class: any circuit interleaving queued single-qubit gates with a native `g` call gets corrupted amplitudes.

## Proposed resolution

Add the same `flush` call the sibling two-qubit gates perform at entry, plus a regression test applying `g` after queued single-qubit gates and comparing against the `cz-h-h-cz` decomposition.

Issue #608, opened 2026-08-26
GPU and CUDA paths are never executed by CI, so GPU changes merge on inspection alone


ciaranra:
## Summary

The GPU and CUDA code paths cannot be executed by CI, so changes to them are merged on inspection alone.

- `cuda-build-check` runs `cargo check --locked -p pecos-cuquantum-sys` and `-p pecos-cuquantum` in stub mode on `ubuntu-latest`. It proves the code compiles; it never runs a kernel.
- The `rust-test` CUDA step installs a toolkit but the runner has no GPU.
- `crates/pecos-gpu-sims/tests/gate_fuzz.rs` is a genuine CPU-versus-GPU amplitude comparison covering `T` and `Tdg`, and it requires hardware. In an environment without a GPU it reports `Device type Cpu is not a hardware GPU` and does not run.

## Why it matters now

The T-gate convention work (#597) changed the GPU constant tables, the auto-dispatch, and the density-matrix system/environment handling. Those diffs were reviewed by inspection and are consistent with the CPU result, but no GPU kernel was executed at any point in that work. The same is true of the Python cuStateVec and MPS simulators, which need CUDA.

This is a structural gap rather than an oversight: no amount of review or CI configuration closes it without hardware.

## Proposed resolution

Establish a way to run the GPU suites before merging GPU-affecting changes. Options, roughly in order of cost:

1. A manually dispatched workflow on a self-hosted or cloud GPU runner, run on demand for PRs that touch `crates/pecos-gpu-sims`, `crates/pecos-cuquantum`, or the CUDA Python simulators.
2. A documented local checklist for maintainers with hardware, naming the exact suites to run (`cargo test -p pecos-gpu-sims`, the `gate_fuzz` comparison, and the CUDA Python simulator tests).
3. At minimum, a label or PR-template checkbox recording that GPU paths in a given PR were not executed, so the gap is visible to reviewers rather than implicit.

Until one of these exists, GPU-affecting PRs should say plainly in their description that the GPU paths were not executed.

ciaranra:
## Correction to the framing above

The body of this issue said no hardware was available in the environment used to prepare the related PRs. That was wrong: the development machine has an NVIDIA GeForce RTX 4090, and the GPU suites run on it. The earlier failures came from a sandboxed tool process without access to `/dev/nvidia*` and `/dev/dri`.

[...]

PR #642, opened 2026-08-31
Evolve density matrices as U rho U-dagger and project measurements on the physical register

### LINKED ISSUE #607: Density-matrix evolution applies U to the traced index and U-dagger to the physical index
## Summary

Both density-matrix implementations appear to apply a single-qubit unitary to the traced purification index and its conjugate to the physical row index, which is the reverse of the intended `rho -> U rho U^dagger`. For a non-real unitary this produces `U^dagger rho U`.

The density state is reconstructed as

    rho[row, col] = sum_k psi[(row << n) | k] * conj(psi[(col << n) | k])

so qubits `q + n` act on the physical row index while qubits `q` act on the traced index (`crates/pecos-gpu-sims/src/gpu_density_matrix.rs:197`, `:208`). The helper instead applies the "system" operation to low qubit `q` and the conjugate operation to high qubit `q + n` (`gpu_density_matrix.rs:506`), so `t()` applies `T` to the traced index and `Tdg` to the physical row index (`gpu_density_matrix.rs:934`).

Concretely, after `H|0>` the correct `T rho T^dagger` has `rho_01 = e^{-i pi/4} / 2`; this produces `e^{+i pi/4} / 2`.

## Why it has not been caught

The CPU density implementation has the same reversal (`crates/pecos-simulators/src/density_matrix.rs:238`, `:1315`), so the GPU-versus-CPU comparison test compares two mutually inverted implementations and agrees (`gpu_density_matrix.rs:1074`). It is also invisible for real unitaries and for any observable that does not depend on the sign of an off-diagonal phase.

## Scope

Pre-existing and independent of the T-gate convention work: `git diff origin/dev` for both density files is empty on the branch for #597. Filing separately rather than folding a density-matrix correctness fix into a gate-convention change.

## Proposed resolution

Swap the system/environment placement for non-real unitaries in both implementations, and add an analytical assertion on an off-diagonal element -- for example the `rho_01 = e^{-i pi/4} / 2` case above -- so that a test failure cannot be masked by both implementations sharing a convention.

ciaranra:
Closes #607.

## Summary

Both density-matrix simulators store a purification with `rho[row,col] = sum_k psi[(row<<n)|k] * conj(psi[(col<<n)|k])` — the physical index lives on the high qubits, the traced environment on the low qubits. Two defects against that convention, both execution-confirmed:

1. **Reversed conjugation (issue #607).** Every non-real gate applied the plain unitary to the traced environment and its conjugate to the physical register, evolving `rho -> U-dagger rho U`. Witnesses: `T(H|0>)` returned `rho_01 = e^{+i pi/4}/2` (correct: `e^{-i pi/4}/2`); `S|+>` returned `+i/2` (correct: `-i/2`). Invisible to existing tests because the CPU and GPU implementations shared the reversal and their comparison test validated them against each other. Fixed gate-by-gate in both implementations: plain `U` on the physical (high) register, elementwise `conj(U)` on the environment, with a per-gate audit of every entry point (real-matrix gates need no distinction; each complex gate's conjugation identity is derived in a comment).

2. **Measurement co-projected the traced environment.** `mz` kept only purification entries whose environment bits also matched the outcome. A purification is defined only up to a right-unitary on the environment, so this destroyed real information for every mixed state — `H, CX, bit-flip(0.3), mz` returned a pure state where `P rho P / tr` is `diag(0.7, 0, 0.3, 0)` — and, combined with the `from(&StateVecSoA)` constructor (which concentrates environment weight on index 0), silently zeroed the entire state after measurement. Fixed: projection on the physical register only (the norm is then exactly `tr(P rho P)`), and a zero-projected-norm outcome now fails loud instead of silently skipping renormalization. CPU measurement also now takes one pass over a single state snapshot instead of materializing a fresh full state per basis-state probability.

## Verification

- Analytic off-diagonal pins with hand-derived expected values (`T`, `S`, `RX(pi/2)`, `RZZ(pi/2)` on `|++>`), plus mixed-state measurement regressions matching `P rho P / tr` elementwise.
- A seeded multi-gate outer-product oracle compares the full density matrix against an independently implemented state-vector simulator (`StateVecAoS`, not the density type's own backing store) at 1e-10.
- An independent physics review drove all 55 gate entry points through an external oracle (worst error 1.5e-15), verified each environment-side conjugation identity against the canonical matrices (all exact, including `conj(RZZ) = (X ox I) RZZ (X ox I)`), and confirmed the noise-channel and re-purification paths were correctly left untouched. Its two measurement findings are the second commit.
- Mutation checks: reversing one gate's placement kills the analytic and oracle tests; restoring the environment co-projection kills the mixed-state regressions and trips the new assertion (verified independently of the implementation session).
- Environment-side operations are a representation choice (any unitary on the traced register leaves `rho` invariant); the module docs now say so explicitly.
- `cargo test -p pecos-simulators` (19 suites) and the full `pecos-gpu-sims` package (201 tests, including the hardware-wired variants executing on a live Vulkan adapter at the crate's established 1e-3 GPU tolerance) pass locally; clippy `-D warnings` and fmt clean. Note CI never executes GPU tests (#608) — the GPU-hardware evidence here is from a local adapter run.

Found in passing and filed separately: #641 (`StateVecSoA::g` misses a pending-queue flush).

## Consumer impact

Diagonal probabilities and measurement outcome statistics were always correct and are unchanged; off-diagonal phases, phase-sensitive expectations, and post-measurement mixed states change to the correct values. No workspace consumer reads the purification's internal layout directly.


ciaranra:
Independent corroboration for the merge decision.

A cross-simulator gate-conformance suite (separate work, unmerged branch) flagged 36 `DensityMatrix` failures, and a verification pass built its own harness -- literal canonical matrices, not the suite's oracle -- to determine whether they were real. It reproduced this PR's exact defect from first principles:

```
|+><+| then S
expected  S rho S-dagger = [[0.5, -0.5i], [+0.5i, 0.5]]
observed                 = [[0.5, +0.5i], [-0.5i, 0.5]]
max_error_expected = 1.0
max_error_reversed = 2.2e-16
```

Same witness as this PR's `S|+>` case. It characterises the implemented behaviour as `U* rho U-transpose` where `U rho U-dagger` is required, tracing it to the purification half-assignment (low qubits act on the gauge/column index of `A`, high qubits on the physical row index, so the physical gate must be plain `U` on the high half).

It also confirmed the GPU wrapper inherits the same defect: unable to run on hardware (llvmpipe only), it instantiated the generic `GpuDensityMatrix<SV>` against a CPU-backed state-vector backend so the real wrapper logic -- half selection and gate dispatch -- executed, and got the identical reversed matrix. The shaders address qubit `q` as `1 << q`, so no backend-specific indexing rescues it.

Two independent instruments, same defect, same witness. Nothing in that verification contradicts this PR.

PR #648, opened 2026-08-31
Route the DEM two-qubit noise gate list through one shared predicate

ciaranra:
Consolidation only. No gate is added to or removed from the set, and this changes no behavior.

## Problem

The DEM builder has one concept -- "a two-qubit gate that carries two-qubit noise during DEM construction" -- spelled out four separate times, in three different syntactic forms:

- `dem_builder/builder.rs` -- a named predicate `is_two_qubit_noise_gate`, used twice.
- `dem_builder/dem_sampler.rs` -- the same thirteen gates as a match-arm or-pattern with an `!loc.before` guard, feeding `cx_groups`.
- `dem_builder/mem_builder.rs` -- the same, feeding `two_qubit_groups`.
- `dem_builder/sampler.rs` -- the same membership test in `compute_location_probs_from_noise`, without `!loc.before`, feeding `p2_rate_for_gate`.

The fourth copy is the instructive one: it asks the same membership question for a different purpose -- classifying a location's scalar rate rather than grouping "after" locations into pairs. Four sites that must agree, with nothing enforcing that they do.

## Change

`is_two_qubit_noise_gate` moves to `dem_builder/types.rs` as `pub(crate)`, alongside the module's other shared helpers, and all four sites consume it. Membership and ordering are untouched.

A regression test in `types.rs` enumerates every valid `GateType` discriminant and asserts the predicate accepts exactly the current thirteen, so editing one site's notion of the set now fails the build rather than drifting silently.

## Behavior

Unchanged, by construction:

- The predicate body is the original, moved verbatim.
- Both or-pattern arms became guarded arms in their original positions, with the same `!loc.before` condition they already had.
- The `sampler.rs` arm did not have `!loc.before` and did not gain one; it classifies a location's rate rather than grouping "after" locations, and keeps that distinct behavior.

## Scope note

This PR was originally motivated by #575, which claimed the omission of `U2q`, `RXXRYYRZZ` and `CRZ` from this list caused silently under-noised DEMs. That premise does not hold on current `dev` and I have retracted it on the issue: `apply_gate` now reports `PauliPropagationOutcome::Unsupported` for gates it cannot model (`propagator/pauli.rs:175`), the DEM entry points preflight on that, and `GateType::CRZ` was removed entirely by #638. Those gates are rejected before they could reach this classification.

What remains, and what this PR is for, is the duplication itself: one concept, four hand-maintained copies. The change stands on that alone.

## Verification

- `cargo test -p pecos-qec` -- 792 unit, 137 integration, 57 doctests, 0 failed.
- `cargo clippy -p pecos-qec --all-targets -- -D warnings` -- clean, re-run cold after touching all changed files.
- Full CI green on this branch: 54 passed, 8 skipped, 0 failed.


ciaranra:
Reviewed. The consolidation is faithful and I agree it is behavior-preserving.

Verified rather than taken on trust:

- All four original sites on `dev` name exactly the same thirteen gates, and the predicate moved into `types.rs` names that same set. No gate is added or dropped.
- The three converted match arms are semantically equivalent. Turning `GateType::CX | ... | GateType::RZZ if !loc.before` into `gate_type if is_two_qubit_noise_gate(gate_type) && !loc.before` preserves fall-through: the guard fails in exactly the cases where the old pattern failed to match, so a "before" location still falls to the later arms as before. None of the thirteen gates is claimed by an arm preceding the converted one in any of the three matches, so arm ordering does not change which branch wins.
- `sampler.rs` correctly keeps its distinct behavior: it had no `!loc.before` guard and does not gain one.

Two notes, neither blocking.

**The description overstates what the new test pins.** It says editing one site's notion of the set "now fails the build rather than drifting silently". The test pins the membership of `is_two_qubit_noise_gate` itself, which is a real and worthwhile guarantee. It does not prevent a future contributor from introducing a fourth spelling of the set at a call site, since nothing forces a call site to consult the predicate. Worth rewording so the guarantee is not read as stronger than it is.

**A latent hazard worth knowing about.** A catch-all binding with a guard is only safe while the guard is there. If the guard is later relaxed or removed during a refactor, every arm after it dies silently, with no unreachable-pattern warning to catch it. The pinning test would not detect that either, since the predicate would be unchanged. This is not a reason to avoid the pattern, only a reason to be careful in that neighbourhood later.

I did not run the test suite for this review; the reasoning above is from reading the diff against `dev`, and CI is green.

Issue #649, opened 2026-08-31
SLR QASM generation should fail when an emitted gate is not resolvable under the declared includes


ciaranra:
## Summary

`SlrConverter(...).qasm()` will happily emit QASM containing gate names that the include list it declares cannot resolve. The failure surfaces later, when the QASM is executed:

```
RuntimeError: Simulation failed: Invalid QASM operation: Undefined gate 'crz'
  - gate is neither native nor user-defined. Did you forget to include ...
```

Reproduce (after #646's fix, which makes the DEFAULT include list sufficient):

```python
prog = Main(q := QReg("q", 2), c := CReg("c", 2), qubit.CRZ(0.7, q[0], q[1]),
            qubit.Measure(q[0]) > c[0], Return(c))
qasm = SlrConverter(prog, includes=["hqslib1.inc"]).qasm()   # succeeds
sim(Qasm.from_string(qasm)).quantum(state_vector()).run(1)   # fails here
```

## Expected

Generation fails, naming the gate and the declared includes, rather than producing a QASM string that cannot run. The generator knows both the gate it is about to emit and the includes it declared, so this is checkable at the point of emission.

## Notes

- The runtime error is clear and actionable, so this is a "fail earlier" improvement, not a silent-corruption bug.
- #646 added `test_qasm_gate_resolvability.py`, which asserts the invariant for the DEFAULT include list. That test is the natural place to extend once generation-time checking exists, and it already contains the machinery for enumerating include-defined gates.
- Deliberately NOT in scope: adding `crx`/`cry`/`crz`/`rzz`/`ch` to `hqslib1.inc`. That file models a specific vendor's gate set; widening it to make an unrelated program resolve would be the wrong fix.

PR #650, opened 2026-08-31
Make every gate the SLR QASM codegen emits resolvable and pin it with a test

### LINKED ISSUE #646: SLR QASM codegen emits crx/cry/crz but defaults to an include that does not define them
## Summary

`SlrConverter`'s QASM codegen defaults its include list to `hqslib1.inc` (`python/quantum-pecos/src/pecos/slr/ast/codegen/qasm.py:211`), but the same codegen emits `crx`, `cry`, and `crz` (`:110-112`), none of which `hqslib1.inc` defines. A program using an SLR controlled rotation therefore emits QASM whose gates are undefined under the default include.

## Reproduce

Build any SLR program containing `CRZ`/`CRX`/`CRY` and render it with `SlrConverter(prog).qasm()`; the output includes `hqslib1.inc` and calls `crz(...)`, which that include does not define.

## Expected

Either the default include list covers every gate the codegen can emit, or the codegen selects the include from the gates actually used, or emitting a gate absent from the chosen includes is an error at generation time rather than a program that fails later.

## Notes

- Pre-existing; it dates from the controlled-rotation lowering in #638, which made `crx`/`cry`/`crz` reachable from this codegen.
- Found while reviewing #637 (the `qelib1.inc` conversion fixes). Practical consequence for that work: the `crx`/`cry` corrections only take effect for callers who include `qelib1.inc`.
- `qelib1.inc` does define all three, so switching or extending the default include is one candidate fix -- but the general problem is that nothing checks the emitted gate set against the declared includes.

ciaranra:
## Summary

Closes #646. SLR's QASM codegen could emit gate names that nothing resolved, so the generated program failed at execution. This completes the QASM boundary table using natives PECOS already has, makes the default include list sufficient, and pins the invariant with a test.

`.inc` files and `parse_native_gate` are the two halves of the QASM boundary: UPPERCASE names resolve as natives (exact-case), lowercase names resolve as include macros, and a name in neither is undefined. No native gate was added to `pecos-core` — every gate below already existed there and was simply unreachable from QASM.

## The defect

`GATE_TO_QASM` (`slr/ast/codegen/qasm.py`) can emit 23 gate names. Six resolved to nothing at all — `SXX`, `SXXdg`, `SYY`, `SYYdg`, `SZZdg`, `ch` — and four more (`crx`, `cry`, `crz`, `rzz`) were defined only in `qelib1.inc` while the generator defaulted to `["hqslib1.inc"]`. Conversely `ZZ` existed only in `hqslib1.inc`, so no single include covered the emitted set.

Every one of the six already existed as a PECOS core native (`gate_type.rs`: `SXX = 53`, `SXXdg = 54`, `SYY = 55`, `SYYdg = 56`, `SZZdg = 58`, `CH = 71`), and `GateType`'s own `Display` spells them exactly as SLR emitted them. The QASM parser's native table simply omitted them, and its `SZZDG` spelling disagreed with PECOS's canonical `SZZdg`.

## Changes

- **`native_gates.rs`**: completed the native table — `SXX`, `SXXdg`, `SYY`, `SYYdg`, `CH`, canonical `SZZdg`, and `SY`/`SYdg` (whose engine dispatch already existed). The exhaustive round-trip test added here also exposed pre-existing gaps for `SZdg`, `Tdg`, `SXdg`, now fixed. The classification match is wildcard-free, so a new `GateType` variant fails compilation until it is explicitly classified for QASM.
- **`engine.rs` / `byte_message/builder.rs`**: dispatch and message construction for the newly reachable two-qubit natives, so they execute rather than merely parse. `SWAP` previously parsed but errored at execution; it now routes through the existing exact three-CX handler.
- **`qelib1.inc`**: `gate ch a,b { CH a,b; }` passthrough; the stale note that controlled-H "would require additional implementations" is gone.
- **`qasm.py` / `gen_codes/gen_qasm.py`**: `GateKind.SZZ` emits the native `SZZ` rather than the `hqslib1` macro `ZZ`, which was the only reason a second include was needed. The default include list is therefore a single `["qelib1.inc"]` — generated programs no longer carry 23 duplicate gate definitions, which strict OpenQASM 2 consumers reject. The substitution is exact by construction: `hqslib1`'s body was literally `gate ZZ q0,q1 { SZZ q0,q1; }`, and the include conformance test from #647 already treats `"szz" | "ZZ"` as the same reference matrix.
- The `GATE_TO_QASM` lookup no longer falls back to `node.gate.name.lower()` for an unmapped gate; an unmapped `GateKind` now raises.

## Tests

`tests/pecos/slr/ast_tests/test_qasm_gate_resolvability.py`:

- **The invariant**: every gate SLR can emit resolves as a native or under the declared includes. It enumerates `GateKind` itself (not `GATE_TO_QASM`'s keys, so a future unmapped gate cannot slip through) and parses the include files for defined names, so neither side is a copied list. A gate that resolves to nothing fails, naming the gate and the includes searched.
- **Discriminating end-to-end tests**: emitted QASM is executed and checked against outcomes that the identity cannot satisfy — `SXX;SXX → 3`, `SYY;SYY → 3`, `SXX;SYY → 0`; `CH` on both control branches (`1` and `2`); and pairwise-distinct controlled-rotation signatures `CRX=(3,1)`, `CRY=(3,3)`, `CRZ=(1,3)`. Before this, SLR-emitted QASM had **no** end-to-end correctness test for `CRX`/`CRY`/`CRZ` at all — the only controlled-rotation coverage was in the Guppy/QIR lane, a different codegen.

Mutation-proven rather than asserted: replacing `CH → cz`, the sqrt-Pauli family → `id`, or `CRX → crz` each fails; deleting `ch` from `qelib1.inc` fails; deleting `ZZ` from `hqslib1.inc` fails precisely `[SZZ]`; a bogus emitted spelling fails naming the gate. An earlier revision of these tests passed with *every* gate under test replaced by the identity — that is what the discriminating assertions fix.

## Blast radius

151 regression snapshots were regenerated. The diff is exactly two kinds of line: 150 include-line swaps (`hqslib1.inc` → `qelib1.inc`) and 8 `ZZ` → `SZZ`. No other snapshot content changed.

## Behaviour on the explicit Quantinuum path

`SlrConverter(prog, includes=["hqslib1.inc"])` is still supported. A program using a gate that `hqslib1.inc` does not define now fails loud at execution — `Undefined gate 'crz' - gate is neither native nor user-defined. Did you forget to include ...` — rather than silently misbehaving. That is not a regression: before this change, the same failure was the *default* for every user. Deliberately not done here: adding `crx`/`cry`/`crz`/`rzz`/`ch` to `hqslib1.inc`, since that file models a specific vendor's gate set.

## Follow-ups filed

- **#649** — SLR should refuse at generation time when an emitted gate is not resolvable under the declared includes, instead of producing QASM that fails later.
- `GateType::CH` has no PHIR mapping (`qasm_to_phir_json.rs` falls through to `Unsupported gate type`), so SLR-emitted QASM containing `ch` cannot be converted to PHIR. Not a regression — it previously failed earlier, at parse — but it is now reachable.

## Verification

Cold: `cargo test -p pecos-qasm -p pecos-core` (1,750), workspace clippy `-D warnings`, `cargo fmt --check`, the Python lane (2,869), the SLR lane (437), `just docs-test` (306), `pre-commit` — all exit 0. Reviewed by an independent adversarial pass which verified every newly reachable native executes with the correct unitary, confirmed no new spelling shadows an existing one, and found the earlier end-to-end test vacuous — the finding this revision addresses.


ciaranra:
Two review passes plus an orchestrator vacuity sweep found that the end-to-end tests, while green, could not distinguish the gates they claimed to test. Three wrong-gate substitutions survived the earlier revisions; all are now killed.

**What survived, and why:**

| Substitution | Why it passed |
|---|---|
| `CRZ -> cphase` (and `CRZ -> cu1`) | every probe prepared the control with `X q[0]`, a definite `|1>`; a controlled rotation and a controlled *phase* differ only by a phase inside the control-1 block, which is unobservable with a definite control |
| `CH -> cry(pi/2)` | same definite-control weakness |
| `SXX -> SYYdg` (and `SYY -> SXXdg`) | every program started in `|00>`, where `SXX|00> = SYYdg|00>` exactly; the two differ only on the odd-parity subspace, which nothing prepared |

The first is the same blind spot as #635 itself -- the defect that issue described survived review for exactly this reason.

**Fixes.** Control-superposition probes at `theta = 2*pi`, where the discriminator is deterministic rather than statistical: `R_P(2*pi) = -I` for every Pauli, so controlled-`R*(2*pi) = Z tensor I` regardless of the target state, giving a deterministic 1, while `cphase(2*pi)` is the identity and gives 0. Plus an odd-parity probe (`X q[1]; SXX; SYY`) that separates the sqrt-Pauli pairs.

**Verified independently by substitution** -- all ten now fail: `CRZ->cphase`, `CRX->cphase`, `CRY->cphase`, `CRZ->cu1`, `CH->cry(pi/2)`, `CH->cz`, `SXX->SYYdg`, `SYY->SXXdg`, `CRX->crz`, and the whole sqrt-Pauli family to `id`. The simultaneous double swap (`SXX->SYYdg` **and** `SYY->SXXdg` together) still collides; that is not a single-edit failure mode and is left as-is.

Also in this revision: a fourth stale doc (`docs/development/slr-qeclib.md`, the one that actually prints the include line -- `docs-test` only `exec()`s the block, so it stayed green on wrong output); `Job_shotnum_usage.qasm` restored to `hqslib1.inc`, since it uses `U1q`, which is hqslib1-only and the include swap had left it unresolvable; and the `includes` docstring now names the five spellings unsupported under an hqslib1-only list.

One review finding was checked and **refuted** rather than acted on: the `F`/`Fdg` lowerings look asymmetric (`F = rx;rz` versus `Fdg = ry;rz`) but `ry(-pi/2); rz(-pi/2)` is exactly the inverse of `rx(pi/2); rz(pi/2)`, verified numerically in both compositions.

Cold verification after the change: cargo qasm+core (1,750), clippy `-D warnings`, fmt, the Python lane (2,869), the SLR lane (437), `just docs-test` (306), pre-commit -- all exit 0.

ciaranra:
A third review pass returned **SHIP, no must-fixes**, having re-derived every probe's expected value independently (including that the `CH` probe's target state is exactly H's -1 eigenvector) and run 30+ wrong-gate substitutions, all caught.

It did find one real coverage hole, now closed: an **angle sign flip** on the controlled rotations survived the entire suite. Both probe families were structurally sign-blind -- a definite control sees only a global phase, and at `theta = 2*pi`, `R(2*pi) = R(-2*pi) = -I`. Compounding it, `crx`/`cry`/`crz` have **zero** regression-snapshot coverage (every other `GateKind` has an exact-text snapshot), so this file is their only guard and nothing in the repo pinned their emitted angle sign.

Closed with sign-sensitive probes at `theta = pi`, where `R*(pi)|+*> = -i|+*>` and the flipped sign gives `+i` -- a relative phase under an `H` control, read out with `SZdg; H`. Verified independently by mutating the codegen's angle rendering to negate every emitted parameter: two tests now fail, where previously none did.

Also in this revision: the `CH` phase probe went from 8 shots to 32 (under a `CH -> cz` mutation its wrong-case pass probability was ~29%, and it *did* pass at seed 42 -- only its sibling probe caught that mutation), and `docs/development/slr-qeclib.md` had its `measure q -> c;` lines corrected and pinned, since the generator emits per-qubit `measure q[0] -> c[0];`. That was the same 'doc-test green on wrong text' failure this PR set out to fix, one line lower in the same block.

Reported and filed separately as #659 rather than widened into this PR: five stored notebook outputs in `examples/` still show the old default include, and nothing in CI executes notebooks.

Verification: 38 resolvability tests, the PECOS lane (2,869), the SLR lane (437), `just docs-test` (306), pre-commit -- all exit 0.

ciaranra:
The earlier `Test and build documentation` failure on this branch was **not caused by this PR**; it re-ran green.

Root cause is filed as #660: `docs/user-guide/hugr-simulation.md` has one block that writes `/tmp/pecos-doc-tests/circuit.hugr` (lines 157-159) and later blocks that read it (170, 173). `just docs-test` runs each block as a separate test under `pytest -n auto`, so the reader can be scheduled before the writer. Reproduced deterministically:

```
rm -rf /tmp/pecos-doc-tests && just docs-test   -> FAILS (test_hugr_simulation_block_4)
just docs-test                                   -> PASSES (the file left behind satisfies the reader)
```

`dev` has been green by luck rather than by ordering guarantee. This PR added one documentation block (the `slr-qeclib.md` assertion), which changed the block count, the xdist distribution, and therefore the outcome. Nothing here touches HUGR simulation.

Issue #645, opened 2026-08-31
propagate_s_gate_frame is a no-op for disjoint X/Z observable masks


ciaranra:
## Summary

`propagate_s_gate_frame` (crates/pecos-decoder-core/src/pauli_frame.rs:~160) documents "X-type bits that are set also flip the corresponding Z-type bit", but computes

    *frame.frame_mut() ^= (f & x_obs_mask) & z_obs_mask;

For the documented convention of disjoint X/Z observable masks (e.g. `x_mask = 0b01`, `z_mask = 0b10`), `(f & x_mask) & z_mask == 0` always — the S propagation never changes the frame. The helper needs a position mapping from each X bit to its corresponding Z bit (as `apply_boundary_gate`'s `SGate` arm already does correctly with explicit bit indices at logical_algorithm.rs:214).

## Impact

Latent: no production caller today (found during a design review of the streaming/transversal decoding subsystem, which intends to build on this family of helpers). The sibling CNOT/H helpers have tests; S has none — adding the correct implementation must come with a disjoint-mask regression test, since the current expression passes any test that uses overlapping masks.

## Proposed resolution

Reimplement with an explicit x-to-z bit mapping (mirroring the `BoundaryGate::SGate` frame arm), add a disjoint-mask test, and mutation-check it (the current no-op form must fail the new test).

PR #664, opened 2026-09-02
Add just-in-time DEM slice composition and a streaming window decoder with whole-component commit

qciaran:
## Summary

Adds PECOS-native just-in-time detector-error-model slicing, caching, placement, routing, and commit-plus-buffer composition, and replaces the `windowed` decoder with a streaming window decoder that commits whole correction components.

- reusable, absolute-index-free `DemSlice` values with checked temporal horizons
- one owner slice per physical fault source, including forward detector ports and hyperedges
- deterministic physical-fixture caching with instance-only round, label, output, and coordinate mappings
- structured `DetectorErrorModel` extraction preserving independent sources, Y decompositions, arbitrary source-frame components, correlated two-qubit sources, hyperedges, and tracked-Pauli outputs
- ownership derived from existing `DagFaultInfluenceMap` locations and `dem_slice_round` DAG annotations
- stable detector streams derived from existing `[x, y, round]` detector coordinates, with checked per-round detector-order routings
- checked soft/hard temporal boundaries, source halos, minimum buffer derivation, and probability combination after window projection
- checked instance-time GF(2) routing for standard-output and tracked-Pauli columns
- production bounded-cache providers for one- and multi-patch surface memory, including the fused one-round family
- production boundary providers for repeated transversal H, repeated transversal CX, and mixed two-patch H/CX schedules
- full structured construction retained as the independent equivalence oracle and conservative fallback
- a streaming window decoder with whole-component commit, replacing the sandwich and overlap modes (see below)

This PR remains a draft for maintainer review of the new native API and its integration surface. The fold-S frontend correction, lattice-surgery families, adaptive slice families, prior refresh, and the known anti-snake logical-observable-matching decoder issue are explicitly separate follow-ups.

## Design

Under a bounded detector-correlation horizon, each physical fault contribution is owned by one constant-depth slice. Targets use `(local detector stream, relative round)` addresses, so a contribution can retain forward-in-time edges and hyperedges without referring to final `D<n>` indices. A composer places O(1) neighboring slices into a decoding window and resolves those ports. The split follows the recipe construction of Topological code Autotune (arXiv:1202.6111), the window boundary follows the overlapping recovery of Dennis et al. (quant-ph/0110143) and the buffering condition of Bombin et al. (arXiv:2303.04846), and the port interface is the gadget-level composition formalized by Kliuchnikov, Lietz, and Pastawski (arXiv:2609.13087).

Physical cache identity is separate from scheduling state. Geometry, boundary bases, physical orientation/state, noise parameters, and circuit family select an entry. Algorithm depth, absolute round, patch label, qubit offset, spatial placement, and zero-time logical relabeling live on instances. Output routing is an explicit GF(2) transformation: one local output can fan out to several algorithm outputs, repeated targets cancel, and an empty target set deliberately projects a column away.

`DemSlice::from_detector_error_model` requires mappings for every referenced detector and output. Location-selected extraction includes a source only when all of its physical locations share the owner; partial or unattributed ownership fails. Component decompositions are checked to XOR to the source's complete effect.

`DemWindowSpec` models a hard backward boundary and an explicit soft or hard forward boundary. A soft boundary may project unmatched buffer-side support to the boundary, but may not truncate a correlation touching the commit region. `required_buffer_rounds` derives the exact safe look-ahead. Contributions that become identical after projection retain independent-source XOR probability combination.

The slice-layer vocabulary is now `compose` and cached slice (formerly stitch and template): `DemComposer`, `ComposedDem`, `CachedDemSlice`, `DemSliceRoundSchedule.compose`, `from_cached_slices`, `cached_slice`, and `composed_round_window`.

## Surface-code integration

Memory uses bounded physical fixtures for initialization, stationary bulk SEC, pre-terminal SEC, and destructive measurement. A dedicated one-round fixture represents the fused initialization/SEC and terminal roles. Simultaneous independent patches use a combined canonical fixture whose detector-stream partitions are translated independently; requested depth and placement do not enter the physical key.

Transversal-H families distinguish pre-gate bulk, pre-gate boundary, H plus first post-gate SEC, post-gate bulk, pre-terminal, and terminal roles. Repeated H composition normalizes physical X/Z swap parity and the future logical H action to four finite states.

Transversal-CX families retain correlated two-patch sources and independently place both canonical stream partitions. Repeated CX uses checked GF(2) fan-out for later logical action. Mixed H/CX schedules normalize the remaining two-qubit real-Clifford action to a finite canonical word, so cache cardinality is independent of algorithm length.

The existing standalone `add_transversal_sz` circuit is deliberately not cached. It emits a between-round physical phase layer even though its API documentation describes a mid-cycle fold-transversal S-SE protocol; its detector span grows with the preceding memory depth. A sound S provider first requires correcting and validating that physical surface-code circuit. Likewise, PECOS's existing bounded `PyLogicalCircuitDecoder` fallback remains in place because the pinned logical-observable-matching regression demonstrates the time-like-snake failure described in arXiv:2505.13599.

## Streaming window decoder

The previous `windowed` default was not a window decoder. With any buffer, `mode = auto` resolved to the sandwich decoder, whose phase-1 weight threshold committed nothing on circuit noise, so its full-graph phase 2 (correlated PyMatching by default) produced a monolithic decode. The overlap mode, the only genuinely windowed path, dropped chains straddling a commit boundary.

The replacement commits a connected component of the selected correction only when all of its defects are old and defers it otherwise (Dennis et al., quant-ph/0110143, Sec. VI). PECOS adds:

- a one-step look-behind holding only deferred defects, and a deferred-once forced commit, so every defect is resolved within two windows
- full provenance from each local edge to its global DEM column, so a forced commit through a projected column carries its true far detectors
- a production incidence check on every decode, and window solvability and completeness checks at construction
- an `EdgeDecoder` trait with PECOS UF and PECOS PyMatching (uncorrelated and correlated) inner decoders
- `windowed` now requires `inner`, `buffer`, and `step` explicitly
- construction rejects columns that share detectors but disagree on observables, because a matching edge carries one observable label; issue #780 tracks the terminal graphlike decomposition that produces such columns

Measured on one shot set (rotated surface memory, d = 5, 18 rounds, uniform circuit noise p = 5e-3, 20000 shots), failures:

| Decoder | Failures |
|---|---:|
| PyMatching, monolithic | 1048 |
| PyMatching correlated, monolithic | 729 |
| previous `windowed` default, every step/buffer tried | 729 |
| previous overlap mode | 4352 to 5304 |
| new decoder, PyMatching inner, step 5, buffer 5 | 1049 |
| new decoder, correlated PyMatching inner, step 5, buffer 5 | 730 |

The last two rows come from the reference prototype; the Rust decoder reproduced the prototype's prediction on every one of 120,000 shots compared. At short buffers whole-component commit removes 25 to 45 percent of the excess of eager column commit over monolithic. Correlation evidence does not cross windows, which leaves a small systematic excess of about 1 percent relative with correlated matching; a fixture pins it.

Related fixes: `PyMatchingDecoder`'s edge decode now applies correlations when the decoder was built with them (the bridge called the uncorrelated routine), and `DetailedDecoder::decode_to_edges` for `PyMatchingDecoder` returns correction edges with weights and observables instead of matched detection-event pairs.

## Breaking changes

- `windowed` spec: removed `mode`, `seam`, `ext`/`core_extend`, `wmax`/`commit_weight_max`, and `sandwich_phase2`; `inner`, `buffer`, and `step` are required. Callers who want the previous result should request the monolithic decoder directly.
- Removed `OverlappingWindowedDecoder`, `SandwichWindowedDecoder`, and `WindowedDecoder`; beam search is unchanged.
- The standalone Python logical-subgraph window wrapper no longer defaults to step 8 and buffer 4; both are required. Window-plan APIs take a non-zero step.
- `StreamingDecoder::flush` is replaced by `finish`.
- `DetailedDecoder::decode_to_edges` for `PyMatchingDecoder` returns correction edges.
- Slice-layer renames listed under Design.

## Verification

- includes the owner's `origin/dev` merge at `4550e84c`
- `cargo fmt --all -- --check`
- cold `cargo clippy -p pecos-qec -p pecos-decoder-core -p pecos-decoders -p pecos-uf-decoder -p pecos-pymatching -p pecos-rslib --all-targets -- -D warnings`: 0 warnings
- `cargo test` on those five library crates: 1523 passed, 0 failed
- `just build-debug`; `pytest python/pecos-rslib/tests/test_decoder_specs.py python/quantum-pecos/tests/qec/surface`: 1971 passed, 15 expected failures
- repository pre-commit hooks on every changed file, and `just docs-build`
- exact byte-for-byte reconstruction against independently compiled full structured DEMs; exact surface-memory, repeated-H, repeated-CX, and mixed H/CX equivalence; warm-cache tests that reject any full-circuit compile
- window decoder, deterministic fixtures: a straddling strip where eager commit fails, an undercut fixture, a forced commit through a projected column with independent recomputation of the carry, a far-detector join, the documented correlation limitation, a future-veto fixture, a future-edge representative fixture, and the merged tail
- window decoder, mutation runs: removing forcing, skipping the residual update, removing the incidence check, truncating the merged tail, using the uncorrelated edge routine, dropping the future veto, deriving the future flag from the representative, and preferring the non-projected member each fail their test
- window decoder, paired acceptance against the monolithic decoder with the same inner decoder (PECOS-native shots, d = 5, 18 rounds, 20000 shots, step 5, buffer 5): every setting within `a - b <= 3 sqrt(a + b + 1)` with zero decode errors

| Inner decoder | p | a (window fails, monolithic succeeds) | b (reverse) |
|---|---:|---:|---:|
| PECOS UF | 5e-3 | 178 | 315 |
| PECOS UF | 1e-3 | 8 | 15 |
| PyMatching | 5e-3 | 0 | 2 |
| PyMatching | 1e-3 | 0 | 0 |
| PyMatching correlated | 5e-3 | 15 | 7 |
| PyMatching correlated | 1e-3 | 0 | 0 |

- an independent adversarial review found no blocking or major defect; the Rust decoder reproduced the reference prototype's prediction on every one of 120,000 shots compared

## Follow-ups

- [ ] correct and verify the documented mid-cycle fold-transversal S-SE/Sdg-SE physical circuit, then add its bounded families
- [ ] add shallow logical-gate boundary combinations where adjacent memory segments contain only one SEC round
- [ ] add lattice-surgery and adaptive syndrome-extraction families
- [ ] enable windowed logical-observable matching only after implementing the published anti-snake modification (the windowed logical-subgraph wrapper now suppresses from d = 3 to d = 5 but not from d = 5 to d = 7; a test pins it)
- [ ] add a topology-stable prior-only update fast path and local cache-weight refresh for drift/leakage estimates
- [ ] add decoder-size-aware logical-operation scheduling
- [ ] streaming contract for the window decoder (watermark, finish semantics) and beam search onto `EdgeDecoder`
- [ ] #780: terminal graphlike decomposition observable placement, and an observable-agreement check in the monolithic matching-graph builder


ciaranra:
## Review: REVISE

Reviewed at `8774dee59` against `origin/dev` `9634becaf` in a fresh worktree with a cold target directory. Verification: `cargo test -p pecos-qec --no-fail-fast` (812 unit tests pass, 6 ignored, all integration and 57 doc tests pass), `cargo clippy -p pecos-qec -p pecos-rslib --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean, `just build-debug` and the two changed Python test files (115 passed). A mutation probe that swaps the `bulk` and `pre_terminal` template roles makes the stitcher fail loudly, so the byte-for-byte oracle tests are real.

The Rust slice/stitch core is careful: checked round arithmetic, loud failure on missing mappings, no `unwrap` in production paths, and the repeated-CX GF(2) routing rule was derived independently and matches. The findings below are what keep this from merging as is.

### Blocker

**B1. The mixed H/CX provider silently emits a wrong DEM.** `_canonical_two_patch_suffix` normalizes the future logical action to the shortest word reproducing the sign-free two-qubit Pauli transform. An even number of trailing CX gates therefore normalizes to a word with no CX at all. The surface frontend drops an observable as unreliable when a patch has a CX partner finally measured in a different basis, so the no-CX fixture declares `L0`/`L1` while the real circuit declares none. `from_templates` takes its observable declarations from `boundary_templates[0].output_model` and identity-routes them, so the stitch succeeds.

Reproduced on this branch: two d=3 patches, `mem(3), H(A), mem(3), H(B), mem(3), CX(A,B), mem(3), CX(A,B), mem(3)`, final bases `{A: "X", B: "Z"}`. The cached path is taken, `build_dem()` differs from `_build_structured_dem()`, the oracle declares no observables and the cached output declares `L0` and `L1`. The pure-CX provider guards exactly this case (`final_control_basis != final_target_basis` returns `None`); the mixed provider has no such guard, and `test_mixed_h_cx_provider_matches_full_compile` only covers words where every boundary retains a CX.

Fix direction: include per-patch CX presence in the canonical state (or apply the same final-basis guard), and structurally have `from_templates` reject a template whose declared outputs disagree with the assembled circuit's observable set instead of trusting `output_model`.

### Major

**M1. Coordinate-keyed detector streams break a circuit that builds on `dev`.** `detector_round_layout` keys streams on `(x, y)`. `add_patch`'s default `coord_offset` spaces patches by the newly added patch's `dz`, so heterogeneous sizes overlap. Reproduced: `SurfacePatch.create(5)` and `SurfacePatch.create(2)` with default placement and distinct qubit offsets, `add_memory(["B", "A"], 3, "Z")`. On `dev`, `build_algorithm_descriptor` returns one 80-detector segment; on this branch it raises `maps multiple offset-zero source detectors to emitted local detector 2`. The fallback path calls `round_schedule` unconditionally, so this is a regression of a public API, and the error names neither the patch nor the coordinate. The PR's own canonical fixtures space patches correctly; `add_patch` does not. Deriving stream identity from stabilizer identity (which the frontend already knows) removes the coordinate-uniqueness contract entirely; validating placement at `add_patch` is the smaller fix.

**M2. Unbounded module-global caches keyed on raw noise floats.** All five providers are `functools.cache` with no bound, each retaining a full fixture `DetectorErrorModel` plus templates for the process lifetime. A noise sweep, the headline use case for this machinery, retains one fixture per noise point with no way to release it. The Rust `DemSliceCache` exists for this purpose and has zero production callers.

### Minor

- `required_buffer_rounds` and `contribution_is_relevant` iterate `combined_effect()`. A detector present in both components of a decomposed contribution cancels there, and components are instantiated with `validate_boundary = false`, so a component target can be projected past the window silently. Probe: `required_buffer_rounds(0, 1)` returns 0 for a slice whose component reaches two rounds forward. The combined mechanism is unaffected; the decomposition handed to a graphlike decoder is. Iterate the union over components, or state the contract precisely in the doc.
- Stitched models drop `location_indices`, so a stitched model cannot be re-sliced (`UnattributedSourceContribution`) and `collect_source_graphlike_mechanisms` skips every contribution. Nothing pins that this index is irrelevant for the surface families.
- `segment["dem"]` now includes look-behind and look-ahead detectors while `num_detectors` counts only the commit partition. No current consumer indexes the segment DEM with a `num_detectors`-sized syndrome, but the dict contract changed without a checked representation. `detector_rounds` also silently drops detectors without coordinates.
- `buffer=0`, the previous default, now raises. Deliberate and tested, but source-breaking for external callers; worth a release note.
- Public API with zero non-test callers across the workspace and bindings: `DemSliceCache`, the `DemSliceModelMap` builders, direct `DemStitcher::new`, `StitchedDetectorAddress`, `DemStitchDiagnostics`, `DemSlice::from_detector_error_model`, `DemSliceTemplateCompiler::from_annotated_circuit`, both `from_location_rounds`, and `with_dem_output`/`with_tracked_pauli` (bindings use the `_targets` forms). All are re-exported from the crate root. `_boundary_template_instances` in `logical_circuit.py` is defined and never called.
- `DEM_SLICE_ROUND_ATTRIBUTE` is duplicated as a Python literal with a "kept in sync" comment; export it from `pecos_rslib.qec`.
- The CX and mixed providers never pass `tracked_pauli_routings`; a tracked-Pauli output would get identity routing where the CX fan-out applies. Latent today, so raise if `template.tracked_paulis` is non-empty.
- One unchecked `i64` subtraction in `required_buffer_rounds` where every other step is checked.
- The Rust equivalence tests use `compare_dems_exact`, which compares the aggregated mechanism set. Only the Python tests are byte-for-byte; the design doc's "byte for byte" claim should say so.

### Reuse against existing DEM functionality

Reused correctly: `DagFaultInfluenceMap` locations for ownership, the `FaultContribution` component accessors, `DetectorErrorModel::add_*_contribution` with render-time XOR probability combination, `compare_dems_exact`, and the existing `[x, y, round]` detector coordinates. Deleting the string-parsing segment filter in `build_algorithm_descriptor` and folding `build_sampler_and_decoder`'s duplicated builder setup into `_build_structured_dem` are real simplifications. `dev` has no prior DEM template cache (the caches in `builder.rs`, `types.rs`, and `fault_sampler.rs` are internal memoization), so the JIT layer itself is new ground.

Duplication that adds complexity and should be reduced before or shortly after merge:

1. `SliceFaultMechanism` and `DemSliceContribution` are relative-address twins of `FaultMechanism` and `FaultContribution`: the same three namespaces, `from_unsorted`, `xor`, `is_empty`, and the same direct / Y-decomposed / source-decomposed shape. Only the detector element type differs (and `parity_sorted` cancels where `FaultMechanism::from_unsorted` only sorts). A container generic over the detector type, or a thin wrapper that converts at the boundary, would remove the second copy.
2. Window extraction now exists twice. `extract_window_dem` and `parse_dem_params` in `pecos-uf-decoder/src/windowed.rs` filter DEM text by time coordinate with implicit boundaries; `DemStitcher::stitch` does the same job structurally with checked boundaries. The windowed decoders are not pointed at the new path, so the decoder side still re-parses strings.
3. H/CX Pauli propagation is written a third time in `_apply_two_patch_gate_to_pauli`, alongside `apply_boundary_gate` in `logical_algorithm.rs` and `propagate_cnot_frames` in `pauli_frame.rs`.
4. The five providers re-inline the same boundary-scheduling loop and coordinate-offset computation. Each new logical gate family costs another 150-300 lines in this shape, which is the main long-term complexity cost of the PR.
5. `DemSliceDetector` beside `DetectorDef`, and `DemSliceCache` beside `functools.cache`: two representations of the same idea, one of them unused.

qciaran:
Addressed the review in ad05948d3, then merged the latest origin/dev in 7a199827a.

Correctness and contracts:
- B1: mixed H/CX schedules with history-sensitive mixed-basis CX outputs now take the exact full-model fallback; the reported even-CX reproduction is pinned.
- from_templates now requires an independently supplied standard-output/tracked-Pauli schema, checks it against output_model, and rejects any unprojected template output outside that schema.
- M1: default heterogeneous patch placement is cumulative by actual patch width; duplicate detector coordinates in one round now fail with both D IDs and the coordinate.
- M2: all six physical provider caches are bounded LRUs; the unused duplicate Rust cache wrapper was removed.
- Buffer sizing, relevance, and hard/soft boundary validation now inspect every decomposition component before XOR cancellation, with a regression for the reported +2-round canceled target.
- Template source location IDs are retained and remapped per instance in stitched DEMs, with uniqueness/provenance tests.
- Segment dictionaries now expose num_commit_detectors and num_window_detectors while preserving num_detectors as the commit count; missing detector round coordinates fail explicitly.
- The buffer=0 behavior has an Unreleased breaking-change note.
- DEM_SLICE_ROUND_ATTRIBUTE is exported from pecos_rslib.qec instead of duplicated in Python.
- Tracked-Pauli output mismatches fail through the same structural schema check.
- The remaining required-buffer subtraction is checked.

Cleanup:
- DemSliceContribution now has one shared probability/provenance envelope.
- H and mixed H/CX providers share boundary placement.
- The dead boundary helper was replaced by the shared implementation.
- Slice APIs are no longer duplicated at the pecos_qec crate root.
- The documentation now scopes byte-for-byte claims to Python serialization and states the coordinate identity and segment halo contracts.

Validation on the merged tree:
- cargo test -p pecos-qec --lib: 814 passed, 6 ignored before the dev merge; all 22 slice tests pass after the merge.
- cargo test -p pecos-qec --test jit_dem_slice_tests: passed.
- cargo clippy -p pecos-qec --all-targets -- -D warnings: clean.
- Focused Python surface suites: 118 passed after the merge.
- Full surface directory: 774 passed, 15 xfailed, with one unrelated BP/OSD golden-fixture mismatch: the decoder chooses qubit 6 instead of qubit 3 for an equal-weight logically equivalent correction; logical flags and decoding weight are identical.

I left the decoder-side text-window implementation and cross-layer H/CX propagation consolidation as separate architectural follow-ups: replacing those safely requires changing the windowed decoder input boundary, not just deduplicating this JIT provider. The new production descriptor path itself uses structured stitching and does not reintroduce text filtering.

ciaranra:
## Re-review of ad05948d3 + 7a199827a

Re-verified in the same worktree, updated to `7a199827a`, cold target directory: `cargo test -p pecos-qec --no-fail-fast` green (all unit, integration, and doc tests), clippy for `pecos-qec` and `pecos-rslib` with `-D warnings` clean, `cargo fmt --check` clean, `just build-debug`, then the whole `tests/qec/surface` directory under the CI marker filter: 775 passed, 15 xfailed. The BP/OSD golden-fixture mismatch mentioned above does not reproduce here; `test_decode_numpy_migration_equivalence.py` passes 63/63 on the merged tree, and the slow/optional lane selects no surface tests. If it is still failing on your machine it is worth checking that the extension in that environment was rebuilt after the `dev` merge.

### Confirmed fixed

- **B1.** The even-CX mixed-basis reproduction now takes the exact fallback: cached path returns `None`, `build_dem()` equals the full compile.
- **M1.** Default heterogeneous placement (`create(5)` then `create(2)`) now builds a descriptor; a manual overlap fails with `source DEM detectors D15 and D19 share spatial coordinates (7, 1) in round 1`.
- **M2.** All six providers are `lru_cache(maxsize=16)`; the Rust cache is gone.
- Buffer sizing, relevance, and hard/soft validation now iterate `detector_targets()` (the union over components) with the regression test for the cancelled `+2` target; source locations are retained and remapped per instance; `num_commit_detectors` / `num_window_detectors` are exposed and missing round coordinates raise; the attribute is exported from `pecos_rslib.qec`; the subtraction is checked; the changelog and design doc updates match the code.

### Still open

**1. The new schema check does not backstop B1; only the Python guard does.** With the `"cx" in gate_names and final_control_basis != final_target_basis` guard removed, `from_templates` accepts the mixed-provider assembly and `build_dem()` again returns a model with `L0`/`L1` where the oracle has none. The reason is that `expected_dem_outputs` is not an independent declaration: each provider hard-codes it (`[0]`, `[0, 1]`, `list(range(len(patch_states)))`, and in the CX provider an inline re-statement of the frontend's reliability rule). `output_model` and the literal agree with each other, so the equality check passes. The check as written catches a template whose local outputs are not routed, which is useful, but not the failure that was reported. To make it structural, derive the expected output set from the assembled circuit itself (the builder already computes observable reliability when it emits the final measurement layer; factor that into one method every provider calls), so a fixture that disagrees with the real circuit is rejected regardless of which provider assembled it. That also removes the second copy of the reliability rule now living in the CX provider.

**2. Stitched `location_indices` live in a different identity space, and nothing marks that.** They are dense fresh IDs per `(instance, local location)`, not indices into any influence map, as the doc now says. Before this round, re-slicing a stitched model against an influence map failed loudly with `UnattributedSourceContribution`. Now `window.round_schedule(influence_map, circuit)` on a stitched window is only rejected because the fresh IDs happened to straddle two owner rounds (`owns only part of correlated source locations [42, 43]`); for a different window they could all land in one round and be silently misattributed. Either reject re-slicing of a stitched model explicitly or keep the provenance in a field that is not read as an influence-map index.

**3. Remaining cleanup, none blocking.** The CX provider still carries its own placement loop (`placed.extend(...)`) beside the shared `_boundary_template_instances`. Zero-caller public API in `slice.rs` outside its own tests: the `DemSliceModelMap` builders, `DemSlice::from_detector_error_model`, `DemSliceInstance::with_dem_output` / `with_tracked_pauli`, `StitchedDetectorAddress`, `DemStitchDiagnostics`, and direct `DemStitcher::new`. Setting `location_indices` through `last_contribution_mut` after `add_*_contribution` works but couples the stitcher to "the last pushed contribution"; an `add_*` variant that takes the locations would be more direct.

Deferring the decoder-side text-window consolidation and the H/CX propagation consolidation to follow-ups is reasonable given the reason stated; please file them so they do not get lost.

Verdict: mergeable once item 1 is addressed. Items 2 and 3 can ride along or follow.

qciaran:
Addressed the re-review in `03c06f21a`, then merged the current `origin/dev` (`3aaa94a4a`) in `96f73325c`.

1. The output-schema check is now genuinely independent of the cached fixture.
   - `LogicalCircuitBuilder._assembled_dem_output_ids()` reads the observable IDs emitted by the actual assembled circuit frontend, including its complete logical-gate history and final-measurement reliability decisions.
   - Every production template provider (memory, multi-memory, H, CX, and mixed H/CX) passes that one result to `from_templates`.
   - The duplicated CX reliability expression was removed.
   - A mutation-style regression forces a fixture/circuit schema disagreement and pins loud rejection. The reported history-sensitive even-CX circuit also explicitly pins that the assembled circuit declares no outputs.

2. Stitched source identities can no longer be mistaken for influence-map indices.
   - `DetectorErrorModel` records whether its location IDs are physical influence-map indices or stitched provenance.
   - Stitching marks the latter identity space explicitly.
   - All ownership-based/re-slicing entry points reject a stitched model immediately with a dedicated diagnostic, before consulting any supplied influence map or round-owner table.
   - The regression verifies the explicit rejection.
   - Stitching now uses direct add-with-stitched-locations variants, removing the coupling to `last_contribution_mut`.

3. Additional cleanup:
   - Repeated CX now uses the shared boundary-placement implementation; the routing-specific boundary index is retained by the common helper.
   - Filed #748 for decoder-side structured window extraction and #749 for cross-layer H/CX Pauli propagation consolidation.

Validation on the merged head:
- `cargo test -p pecos-qec --no-fail-fast`: all unit, integration, and 57 doc tests pass (823 unit passed, 6 ignored).
- `cargo clippy -p pecos-qec -p pecos-rslib --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- Focused Python surface suites: 119 passed.
- Full fast surface directory: 775 passed, 15 xfailed, with the same isolated `d3-x-bp_osd` golden correction mismatch seen previously. The rebuilt extension selects Z correction qubit 6 rather than qubit 3 at the same weight; logical flags are identical.
- The repository-wide `just build-debug` completed dependency sync and all six Selene plugin builds, but its final maturin stage repeatedly hit a local rustc SIGSEGV during LLVM target initialization before compilation. Direct cargo test/clippy, the exact rustc probe, and uv's native-extension builds all succeed; no code diagnostic was emitted.

qciaran:
Added the remaining low-risk ride-along cleanup in 452a03a1.

- Narrowed the Rust API to the production schedule-level path: the raw model map, template compiler, and direct stitcher are now internal.
- Removed the zero-caller single-output DemSliceInstance convenience methods and the genuinely unused compiler constructor. The public integration test now assembles through DemSliceRoundSchedule, matching production.
- Consolidated the duplicated two-patch detector-coordinate translation used by the CX and mixed H/CX providers.
- Updated the design note to distinguish the public schedule API from internal compilation/stitching machinery and to link the remaining follow-ups.

The architectural items are now tracked with acceptance criteria:
- #748 structured decoder window extraction
- #749 shared H/CX Pauli propagation
- #750 unified slice/fault-model representations
- #751 data-driven logical-boundary providers

Validation on 452a03a1:
- cargo test -p pecos-qec --no-fail-fast: all unit, integration, and 57 doc tests pass (823 unit passed, 6 ignored)
- cargo clippy -p pecos-qec -p pecos-rslib --all-targets -- -D warnings: clean
- focused slice tests: 22 passed
- focused Python surface suites: 119 passed
- repository-wide pre-commit: clean

ciaranra:
## Re-review of 03c06f21a, 96f73325c, 452a03a17

Re-verified in the same worktree at `452a03a17`: `cargo test -p pecos-qec --no-fail-fast` (823 unit passed, 6 ignored, all integration and doc tests), clippy for `pecos-qec` and `pecos-rslib` with `-D warnings` clean, `cargo fmt --check` clean, `just build-debug` completed, full `tests/qec/surface` under the CI marker filter: 776 passed, 15 xfailed. `test_decode_capture_is_bit_for_bit_identical[d3-x-bp_osd]` passes here in isolation as well, so the mismatch on your machine is environmental (consistent with the rustc SIGSEGV you saw in the maturin stage); a clean rebuild of the extension in that environment should clear it.

Both open items from the previous round are closed, and I confirmed each by execution rather than by reading:

1. **Schema check is now real.** With the mixed provider's final-basis guard removed, `from_templates` rejects the even-CX fixture on its own: `output_model declares standard outputs {0, 1}, but the assembled circuit expects {}`. Every provider passes `_assembled_dem_output_ids()`, and the inline reliability rule in the CX provider is gone.
2. **Stitched provenance cannot be mistaken for influence-map indices.** `round_schedule` on a stitched window (start rounds 0, 1, 2) and on a provider-assembled model all fail immediately with the dedicated `stitched provenance` diagnostic, before any owner table is consulted. The `add_*_with_stitched_locations` variants remove the last-contribution coupling.

The API trim, the shared placement and two-patch offset helpers, and the four follow-up issues (#748 to #751) match what was described.

One non-blocking note for #751: on the warm cached path `_assembled_dem_output_ids()` still emits the entire physical circuit through `to_tick_circuit()` to read the observable metadata. At d=5 with 60 rounds a warm `build_dem()` is about 72 ms, of which the schema derivation is about 12 ms, against roughly 197 ms for the full compile. Correct, and a clear win, but the JIT path is still doing O(physical circuit) work per assembly. Deriving final-measurement reliability from the logical operation list, with a test pinning it to the emitted metadata, would make the warm path independent of physical depth.

Verdict: mergeable.

qciaran:
Implemented and pushed all four review follow-ups:\n\n- d3626667 — shared Rust H/S/CX frame propagation, including the Python surface path (#749)\n- 8ad5d4d3 — generic FaultMechanism plus shared full/slice contribution shape (#750)\n- 9bb87847 — typed, data-driven H/CX/mixed boundary-provider assembly (#751)\n- e3602c2e — structured decoder DEM windows and direct stitched-model handoff (#748)\n\nValidation on current origin/dev (3aaa94a4a, already contained): full pecos-qec, pecos-uf-decoder, and pecos-decoder-core unit/integration/doc suites pass; clippy with -D warnings and cargo fmt are clean; logical-circuit provider/windowing Python file passes 65/65. Full surface suite: 782 passed, 15 xfailed, with the same isolated local d3-x-bp_osd equal-weight correction tie-break mismatch previously documented (logical flags and decoding weight agree). CI has been triggered on e3602c2e.

ciaranra:
## Review of the follow-up commits (d36266676, 8ad5d4d3a, 9bb878476, e3602c2eb)

Verified at `e3602c2eb` in the same worktree: `cargo test --no-fail-fast` for `pecos-qec`, `pecos-decoder-core`, and `pecos-uf-decoder` green; clippy on those plus `pecos-rslib` with `-D warnings` clean; `cargo fmt --check` clean; `just build-debug`; Python surface and rslib lanes under the CI marker filter: 3084 passed, 15 xfailed. All round-3 probes still hold (schema check rejects the even-CX fixture without the guard; stitched models are rejected on re-slicing). A fresh adversarial arm executed an old-versus-new comparison of the windowed decoders (below).

### The load-bearing part is sound

**#748 windowed-decoder equivalence, verified by execution.** The old `extract_window_dem` from `452a03a17` was copied into a temporary integration test and compared against `StructuredDem::window_by_time` across 35 window bounds on the real d=3 DEM, a hand-built DEM with `^` decomposition, hyperedges, a third observable and duplicate lines, and a coordinate-free DEM. `local_to_global`, every `DemMatchingGraph` edge tuple, `skipped_hyperedges`, and probabilities were identical in every case. Look-behind and look-ahead bounds are passed through unchanged, `ensure_observables_fit_u64` is retained, and plain text or Stim DEMs still work. One real improvement surfaced: the old filter dropped bare `detector D<k>` lines and kept only the last target of a multi-target declaration; the structured path declares every local detector.

**#749 frame algebra** is correct at every call site (`conjugate_s` reproduces the old `SGate` arm, `conjugate_cnot` keeps control-X to target-X and target-Z to control-Z, and the rewritten `propagate_cnot_frames` is algebraically identical to the old mask XORs for any masks). **#750** leaves the `u32` `FaultMechanism` semantics and rendered text unchanged; the full `pecos-qec` suite including the `to_string` goldens passes. **#751** keeps all seven byte-for-byte provider tests and the repeated-CX routing rule.

### Must fix before merge

**1. A single transversal CX with a Y final basis now raises instead of assembling.** `_boundary_output_routings` checks `provider.final_basis not in {"X", "Z"}` before it looks at `later_gate_count`, while the CX provider only screens the basis when `len(gates) > 1`. Before #751 the single-CX case took the identity branch and never read the basis. Reproduced on this branch: two d=3 patches, `add_memory([...], 2, "Y")`, one `add_transversal_cx`, `add_memory([...], 2, "Y")`. The full compile succeeds (2365 contributions); `build_dem()` raises `repeated-CX output routing requires a final X or Z basis`. `Y` is a documented `add_memory` basis and nothing catches the error, so a fallback became a user-facing failure. Move the check inside the odd-parity branch, or hoist it into the provider's `return None` guard, and add `Y` to the single-CX equivalence parametrization.

**2. `SliceFaultMechanism::from_unsorted` silently inverted its XOR semantics.** Before #750 the slice type's only unsorted constructor cancelled repeated targets by parity. As `FaultMechanism<RelativeDetectorTarget>` it inherits the `u32` model's sort-only `from_unsorted`, with the same name and signature, and only a doc comment says to use `from_unsorted_parity` instead. Every in-tree call site was re-pointed, so nothing is broken today; the defect is that the wrong spelling compiles and yields `[D0@+1, D0@+1]` where it used to yield the empty mechanism, and `symmetric_difference` assumes duplicate-free input. Make it structural: a newtype for the relative instantiation whose only constructor is the parity one, or hide the sort-only constructors for non-`u32` detectors.

### Should fix in the same pass

**3. #748's advertised handoff has no consumer, and the doc claim is not realized.** `DetectorErrorModel::to_structured_decoder_dem` and the five `from_structured_dem` constructors are called only by their own `from_dem` text adapters and tests; `pecos-decoders/src/spec/build.rs` and `logical_subgraph_windowed.rs` still enter through text. The design note says a stitched model "does not need a render/parse round trip", but every real path is text to `StructuredDem` to text to inner parser, and the sandwich and beam-search families now re-render the whole model for phase 2 instead of forwarding the caller's string. The re-render was verified byte-safe, so this is a scope and claim defect: either wire the handoff (bindings and `spec/build.rs`) or drop the claim and the unused constructors.

**4. `to_structured_decoder_dem` is a third, divergent conversion of the same model.** `to_mechanisms()` is documented as the structured equivalent of `to_string()` and groups equal effects with combined probabilities; the new conversion neither groups nor takes a render policy and always emits the decomposed form. Executed: two independent `0.01` contributions on `D0 D1` render as one `error(0.0198)` line via `to_string` and as two `error(0.01)` lines via the new path. Matching decoders merge parallel edges so they are unaffected, but a check-matrix consumer would see a different factorization. Since there are no callers yet, route it through `to_mechanisms`' grouping now.

**5. #751 carries validation that cannot fail and data nothing reads.** All seven `raise ValueError` guards in `_assemble_boundary_dem_provider` are unreachable from the three providers (each restates a condition the caller already established or `_boundary_template_placements` already raises), and none is tested. `_BoundaryTemplateSelection.canonical_fixture_key` is written by every provider and never read; `fallback_guards` is a tuple of strings checked only for non-emptiness. Both document rather than constrain. Drop them, or make `fallback_guards` real predicates the assembler evaluates.

**6. `StructuredDem::from_dem_str` is stricter than every other PECOS DEM parser** and the windowed decoders inherit that through `DecodeModel::SingleDem`. Measured: a trailing `#` comment token and an unknown instruction line parse under `DemMatchingGraph::from_dem_str` and are rejected here; probabilities outside `[0, 1]`, non-finite coordinates, and `TP<n>` targets are also newly rejected. Failing loud may be right, but it is undocumented and leaves four parsers in `pecos-decoder-core` disagreeing on the grammar. State the accepted grammar in the `from_dem` error docs, or align the other parsers.

**7. `conjugate_s` and `conjugate_h` are public generic helpers that are wrong for mask words.** PR #651 deleted `propagate_s_gate_frame` and `propagate_h_gate_frame` for exactly this reason (#645): X and Z components of an observable live at different bit positions, so `z ^ x` mixes lanes. `conjugate_cnot` survives the mask instantiation only because it XORs same-type lanes. Constrain the S and H helpers to `bool`, or document the lane-alignment precondition and cite #645.

### Nits

`parity_sorted` is now duplicated byte-for-byte in `slice.rs` and `types.rs`, in the commit whose purpose is unification. `_apply_two_patch_gate_to_pauli` is a one-line pass-through with one caller. The `!window_dem.is_empty()` guards in `windowed.rs` are dead now that every local detector is declared. `transform_two_patch_pauli`'s doc says higher bits are preserved while the binding rejects them.

Verdict: REVISE. Items 1 and 2 before merge; 3 to 5 are cheap now and expensive later.

qciaran:
Thanks — addressed the re-review in d458da16, then merged the current origin/dev at 47762cb11 in cc6bc3055.

The substantive changes are:

1. Single-transversal-CX schedules with a Y final basis now fall back before provider assembly. The Y case is covered by the provider/direct equivalence test; the repeated-CX matched-basis restriction remains explicit.
2. Sort-only FaultMechanism constructors are now structurally limited to FaultMechanism<u32>. SliceFaultMechanism exposes only parity-normalizing construction. A compile-fail doctest prevents the bad constructor from returning, and a runtime test covers duplicate cancellation.
3. The structured DEM is now a real decoder handoff: DecodeModel accepts StructuredDem directly; windowed and beam decoders consume it without a text round trip; sandwich/beam phase two receives structured data; Python exposes DetectorErrorModel.build_decoder; and the logical-subgraph windowing path has a direct structured constructor.
4. DetectorErrorModel.to_structured_decoder_dem now delegates to the same to_mechanisms grouping and XOR-probability factorization as the rest of the DEM path. The regression test checks that two equal 0.125 effects combine to 0.21875.
5. The unused provider eligibility/selection layers, unreachable validation branches, canonical fixture key, and inert string guards were removed.
6. The exact strict flattened StructuredDem grammar and all rejection cases are documented on the public UF constructors.
7. Public H and S conjugation helpers are now bool-only. Generic CNOT mask-word support remains available and tested.

The smaller follow-ups also landed: one shared parity helper, removal of the Python pass-through and dead window guards, aligned high-bit transform behavior/docs, and a warm-path output schema derived from logical operations rather than rebuilding the physical circuit. The latter is pinned against emitted observable metadata across memory and Y-CX schedules.

Post-merge validation is green:
- full cargo test suite for pecos-decoder-core, pecos-uf-decoder, pecos-qec, and pecos-decoders with all features
- 68 focused Python logical-circuit/windowing tests
- strict Clippy for the decoder/QEC/Python Rust crates
- cargo fmt check and git diff check

ciaranra:
## Re-review of d458da16e + cc6bc3055, and merge guidance

Verified at `cc6bc3055` in the same worktree: `cargo test --no-fail-fast` for `pecos-qec`, `pecos-decoder-core`, `pecos-uf-decoder`, and `pecos-decoders` green; clippy on those plus `pecos-rslib` with `-D warnings` clean; `cargo fmt --check` clean; `just build-debug`; surface and rslib Python lanes under the CI marker filter: 3087 passed, 15 xfailed. CI is all green on this head.

Every item from the previous round landed properly, confirmed by execution where it mattered: the single-CX Y-basis schedule now assembles and equals the full compile; the round-3 probes still hold (schema check rejects the even-CX fixture without the guard, stitched models are rejected on re-slicing); the #748 handoff is now real (`DecodeModel::StructuredDem`, `DetectorErrorModel.build_decoder`, structured constructors on the logical-subgraph windowed decoder, `SparseDem::from_structured_dem`); `to_structured_decoder_dem` groups through `to_mechanisms`; the slice type exposes only parity construction; the descriptor decoration is gone; S/H helpers are bool-only; the warm-path schema derivation dropped from about 12 ms to 0 ms.

### Why it cannot merge as is

The PR is now conflicting against `dev` (7 commits behind). The textual conflict is only the import block of `logical_circuit.py`, but the semantic one is larger: #753 and #762 (the surface gadget library) rewrote that file by roughly 800 lines and replaced the entangled-partner observable-reliability rule with `_logical_readout_is_deterministic(operations, segment_idx, patch_label, basis)`, a backward walk over the logical operation list (H swaps the type, repeated CX images cancel, S has no supported logical X image, and a patch consumed before a crossed gate is unreliable). The new `_assembled_dem_output_ids()` on this branch re-implements the old partner-basis rule. After a merge, the schema guard and the emitter will disagree on exactly the schedules where the two rules differ, and `from_templates` will raise out of `build_dem` for them.

The mirror is also already unfaithful to this branch's own emitter. A fuzz of 1370 valid random memory/H/CX/SZ schedules (one to three d=3 patches, X/Y/Z bases) against `to_tick_circuit()`'s observable metadata gives 97 mismatches. The simplest:

```
mem(all); cx(p0, p1); mem([p0])   ->  derived [1], emitted [0, 1]
mem(all); cx(p2, p0); mem([p2]); h(p2); mem([p0])   ->  derived [0], emitted [0, 1]
```

The cause is that the mirror evaluates reliability with end-of-circuit entanglement state, while the emitter decides at each patch's last memory operation. Every mismatching shape is provider-ineligible today (a patch missing from the final memory, or an SZ), so no user-facing failure yet, but the guard is only as good as the mirror, and the mirror is a second copy of a rule the frontend owns.

### Recommended merge move

1. Merge `origin/dev`, then delete the reliability logic inside `_assembled_dem_output_ids()` and have it call `_logical_readout_is_deterministic` with the same observable-id-advancing loop `_emit_final_detectors_and_observables` uses. One rule, one owner; the warm path stays free of physical-circuit emission because dev's rule is already a function of the logical operation list.
2. Pin it with a fuzz, not a handful of cases: random schedules over memory/H/CX/SZ with mixed bases and partial-patch memories, asserting `_assembled_dem_output_ids() == [o["id"] for o in json.loads(tc.get_meta("observables"))]` for every schedule the emitter accepts. The 97 cases above are the regression corpus.
3. Re-check that `dem_slice_round` stamping in `_CircuitGenerator._end_tick` and transversal-gate round ownership survive the gadget-based generator. `dev` still has `_CircuitGenerator`, `round_time`, and `_end_tick`, so the hooks have a home, but the step emission now comes from `gadgets.*` and the byte-for-byte provider tests are the thing that will tell you whether ownership drifted.
4. Re-run the full surface directory after the merge; the gadget commits carry their own byte-parity tests that the JIT stamping must not disturb.

Once that lands with green CI, this is mergeable from my side.

qciaran:
Addressed the latest re-review and merged current `origin/dev` through `d8d57b1f5`. The updated PR head is `4550e84c9` (the implementation/semantic merge is in `072f63fe5`; the final merge only carries the two newer CI-cache/workflow commits).

1. `_assembled_dem_output_ids()` now delegates each candidate readout to `_logical_readout_is_deterministic(...)` at that patch's last memory segment and advances observable IDs with the same loop shape as the emitter. The old duplicated final-basis/CX-partner reliability rule is gone.
2. Added a deterministic 1,370-schedule fuzz across memory, H, CX, physical S/S-dagger, X/Y/Z preparation, and partial-patch memories. Every accepted schedule asserts exact equality between the assembled schema IDs and `to_tick_circuit()` observable metadata; the simple partial-memory regressions are covered explicitly as well.
3. Verified the gadget frontend ownership path directly. A new test checks that transversal H and cross-patch CX gates retain the logical boundary's `dem_slice_round`, while terminal readout H gates retain the terminal round.
4. The dev gadget emitter can reorder dense X/Z detector declarations after H without changing the underlying spatial syndrome stream. The stitch schedule now accepts a checked per-round detector-order permutation, keeping stable stream identity/relative targeting separate from dense emitted ID order. H, repeated-H, mixed H/CX, and provider byte-equivalence tests pass with this integration.
5. Updated the design note and Python binding/stubs to document and validate the new ordering contract.

Local validation on the merged implementation:
- focused Python surface integration: 131 passed
- affected Rust crates with all features, including doctests: passed
- strict Clippy across QEC/decoder/rslib crates: clean
- cargo fmt, Ruff, selected pre-commit hooks: clean
- complete surface directory: 1,876 passed, 15 xfailed, with one local failure in the unrelated `d3-x-bp_osd` bit-for-bit fixture. The decoder chose the other equal-weight physical Z correction; decoding weight, logical flips, and logical-error status match. I left that unrelated tie-sensitive fixture untouched.

Fresh CI is now running on the fully current head.

PR #674, opened 2026-09-02
Make zlup angles a propagating a64 type lowered internally to turns

### LINKED ISSUE #636: zlup HUGR backend rejects gate-target forms and comptime angles that the SLR backend accepts

ciaranra:
## Summary

Closes #636. `a64` becomes a real propagating type in zlup, with a unit written exactly once — where a numeric literal becomes an angle — and every angle resolved internally to turns. Each backend converts once at emission: SLR turns, HUGR half-turns, PHIR/QASM radians.

Issue #636 was scoped as three HUGR lowering gaps. Fixing the third one surfaced the actual defect: angle meaning depended on how a value was spelled and where it was written, so the same angle could emit two different numbers.

## The root defect

`ExportedSymbol::Const` carried no type:

```rust
-    Const { name: String },
+    Const { name: String, ty: Option<TypeExpr> },
```

It was the only export kind that discarded its type — functions kept full signatures. Every imported constant therefore typed as `Unknown`, which meant **no imported angle constant was usable in a rotation gate in any spelling**: `rz(std.a64.t_angle)` failed the type check, and `rz(std.a64.t_angle turns)` failed the literal check. That is the whole standard-library angle surface, `exp/zlup/std/a64.zlp`.

Separately, `extract_angle` applied the turns-to-radians conversion only when the use site was *syntactically* an `AngleLit`, while the comptime evaluator normalised every angle literal to turns on the way in. A named constant holding `0.5 turns` was emitted as if it were 0.5 radians.

## The language rules

A unit is required exactly once, where a literal becomes an angle; re-applying one to something already `a64` is an error. That is what makes `a64` a type rather than a decoration, and it is why a doc example loses a redundant unit:

```
-///   rz(std.a64.quarter_turn turns) q[0];
+///   rz(std.a64.quarter_turn) q[0];
```

Dimensional arithmetic is enforced: `a64 ± a64` is `a64`; `a64 ± scalar` is an error; `a64 * scalar` is `a64`; `a64 * a64` is an error; `a64 / a64` is a dimensionless `f64`; `scalar / a64` and `a64 % _` are errors; compound assignment follows the expanded form; comparisons require both sides `a64` or neither.

`exp/zlup/examples/qft_3qubit.zlp` changes spelling, not circuit — 0.25 turns is pi/2 rad:

```
-pi_2: a64 = 1.5707963267948966;
+pi_2: a64 = 0.25 turns;
```

## Behaviour changes worth stating out loud

- **A symbolic Guppy angle now lowers to no backend.** `def main(theta: float)` with `rz(q[0], theta)` compiles to valid zlup, but a runtime-unresolvable angle is a loud error in all four backends. Previously SLR emitted a bare `VarExpr` naming no declaration, so the Python consumer received an undefined name. Failing closed replaces silently-broken output, but users will notice it.
- **A unitless `a64` declaration is now rejected**, naming the binding.
- **A diverging `if` branch no longer participates in result-type unification**, so `y: i32 = if (c) { return 1; } else { 2 };` type-checks. Two diverging branches still yield `unit`. Ordinary branches now use `check_assignable` rather than exact equality, so the site is no longer stricter than assignability everywhere else.

## The Guppy boundary

Guppy's `angle` is half-turns; the transform converts `f half-turns` to `f/2 turns` at the boundary, in one place. Angle origins that have a real binding site get a converted alias; loop variables and measurement results have none, so they lower directly rather than referencing an alias that was never declared. A mutable angle origin refreshes its alias at every assignment — including inside a loop body or an `if` branch — so a reassignment can never leave a stale value behind.

## Tests

Cross-backend agreement is pinned by **value**, not by "each compiles": `pi rad`, `tau rad`, `1/4 turns`, `0.25 turns`, a module constant, `std.a64.t_angle`, `module_angle / 2`, and `-module_angle` each assert their SLR turns, HUGR half-turns, and PHIR/QASM radians. Two spellings that denote the same angle must produce the same emitted number in every backend.

All eight `.zlp` files under `exp/zlup/examples/` plus `std/a64.zlp` compile on all four backends.

## Follow-ups filed

- QASM codegen silently drops `for` statements — `for i in 0..3 { h q[0]; }` compiles to a file containing only the header and `qreg`. This is what hid an angle-alias scoping bug through three review rounds: no value-level test could observe a loop.
- HUGR collects a loop body once rather than unrolling it.
- Guppy's function scope versus zlup's block scope: a first assignment inside an `if` or `for` body, used after the block, lowers to an out-of-scope reference. Pre-existing and not angle-specific — a plain `int` fails identically — and it fails loud through the generated-zlup validation rather than emitting a wrong program.

## Verification

Cold, CI recipe by name: workspace `cargo test --features=runtime,hugr,neo` with CI's full exclude list (11,182 passed, 0 failed), `zlup` (953), the HUGR codegen suite (61), `guppy-zlup` (165), zluppy (76), workspace clippy `-D warnings`, `cargo fmt --check`, pre-commit, and the four-backend matrix.

Reviewed by independent adversarial passes across eleven rounds. The final round's two must-fixes — an alias emitted for origins with no binding site, and a mutable angle captured once and going stale — were both mutation-proven and then re-probed by the orchestrator on cases the review packet never named (reassignment inside a loop body, reassignment inside an `if` branch, a loop variable inside a nested `if`).


ciaranra:
Rebased onto current `dev`, which has moved 20 commits since this PR's checks last ran.

Re-verified rather than trusting the existing green: an earlier PR in the angle/convention arc (#681) looked ready at 77 green checks and had twelve stale pinned constants, because its ticks predated a convention change. This PR is angle work, so the same check applies -- though its blast radius is smaller, since it touches only `exp/zlup` and `exp/guppy-zlup` rather than the core crates that changed.

Merged cleanly, no conflicts. On the new base:

- `cargo test -p zlup -p guppy-zlup` green in **both** debug and release
- cold `cargo clippy -p zlup -p guppy-zlup --all-targets --all-features -- -D warnings` exit 0
- `cargo fmt --all -- --check` clean

PR #678, opened 2026-09-02
Move the native min-sum BP into a new pecos-bp crate

ciaranra:
## Summary

Creates a new workspace crate, `pecos-bp`, to own PECOS's native belief-propagation primitives, and moves the existing native min-sum BP there. This is a relocation with no change in numerical behavior; it is the first step toward a native Relay-BP, DEM views (basis mask, init-detector restriction, GARI transform), a staged decoder with a deferred outcome, and a GPU BP kernel.

## Changes

- `crates/pecos-bp`: `src/lib.rs` (crate docs and re-exports), `src/min_sum.rs` (former `pecos-decoder-core/src/bp.rs`), `tests/bp_promotion.rs` (moved; its bitwise fixtures pass unchanged and are the proof the relocation changed nothing), `README.md`, `Cargo.toml`.
- `pecos-decoder-core` no longer has `pub mod bp`. There is no compatibility re-export, since `pecos-bp` depends on `pecos-decoder-core` and a re-export would form a cycle.
- Consumers repointed to `pecos_bp`: `pecos-uf-decoder` (BP+UF) and `exp/pecos-trellis`. `pecos-decoders` re-exports `BpGraph`, `BpScratch`, and `min_sum_bp_into`.
- The crate does not carry `pecos-decoder-core`'s crate-level `clippy::cast_possible_truncation` allow. The seven `as u32` index casts in graph construction are replaced by one `index()` helper that narrows with `u32::try_from` and fails loudly if a Tanner graph exceeds the `u32` index space, instead of wrapping silently.
- Workspace dependency entry and `Cargo.lock` updated.

## Verification

- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings`: clean
- `cargo clippy --locked --workspace --all-targets -- -D warnings`: clean
- `cargo test -p pecos-bp`, `cargo test -p pecos-uf-decoder`, `cargo test -p pecos-decoder-core --tests`: all pass
- `cargo doc -p pecos-bp --no-deps`: clean
- `pre-commit run --all-files` with the new files staged: pass
- No remaining references to `pecos_decoder_core::bp` under `crates/`, `exp/`, `python/`, `docs/`, `examples/`

## Note

`cargo clippy -p pecos-decoders --all-targets` with default features is already red on `dev`: three test helpers in `spec/build.rs` are only used under optional features. Workspace-wide clippy unifies features so it passes, and CI's per-crate lane does not include that crate. Not addressed here.


ciaranra:
Rebased onto current `dev` (now carrying #692, #706, #707) and re-verified locally: `cargo test -p pecos-bp -p pecos-decoders -p pecos-decoder-core -p pecos-uf-decoder --release` gives 222 passed across 12 suites, 0 failures; cold `cargo clippy -p pecos-bp -p pecos-decoders --all-targets --all-features -- -D warnings` exit 0; `cargo fmt --all -- --check` clean.

One thing worth stating explicitly, because it is not implied by the title. This is **not** a pure move: it also changes behaviour.

With rename detection, `crates/pecos-decoder-core/src/bp.rs` -> `crates/pecos-bp/src/min_sum.rs` is 95% similar, and the whole delta is (a) import rewrites from `crate::` to `pecos_decoder_core::`, and (b) every `as u32` replaced by a checked narrowing:

```rust
fn index(value: usize) -> u32 {
    u32::try_from(value).expect("Tanner graph index exceeds the u32 index space")
}
```

Silent wraparound becomes a loud panic. That looks like a strict improvement -- a wrapped index would silently corrupt the Tanner graph rather than fail -- and `index()` is only called while building the CSR arrays, not inside the message-passing loop, so there is no hot-path cost. Flagging it so it is on the record rather than found later by someone who read "move" and expected identical semantics.

The rest checks out as a clean extraction: `tests/bp_promotion.rs` moved at 99% similarity (a single import line), so the tests travelled with the code and coverage did not silently drop; no references to the old `pecos_decoder_core::bp` path remain anywhere in `crates/` or `exp/`; and `pecos-bp` is wired into the workspace.

PR #679, opened 2026-09-02
Rename the two-qubit replacement-branch label prefix from *XX to ~XX or :replace:XX

ciaranra:
## Summary

`p2_weights` dictionaries for DEM building accept two-qubit Pauli labels such as `XX` as post-gate Pauli branches. A replacement branch, which omits the ideal two-qubit gate before applying the Pauli, was spelled with a `*` prefix (`*XX`, `*II`). The `*` symbol is being reserved for a different meaning in upcoming noise APIs, so replacement branches are now spelled `~XX` or `:replace:XX` (and `~II` / `:replace:II` for omission only).

## Behavior change (breaking for configurations)

- `~XX` and `:replace:XX` are accepted, equivalent, case-insensitive, and whitespace-trimmed.
- `*XX` is rejected with a `ValueError` that names both new spellings.
- Supplying both spellings for the same Pauli in one dictionary is rejected as a duplicate.
- Plain `II` remains rejected, and its message now points at `~II` / `:replace:II`. That message was unreachable before, because plain `II` failed the label check first; it is reachable now.
- DEM builder behavior (`PauliWeights::replacement_entries`, approximation modes) is unchanged. Docs, diagnostics, and test messages in the DEM builder, the bindings, and the Python `NoiseParameters` docstrings use the new vocabulary.

Existing user configurations that use `*XX` will fail at construction with the migration message.

## Tests

Rust, in the bindings parser: both spellings parse, case-insensitively; `~II` parses as omission-only; `*XX` is rejected with a hint containing `~XX` and `:replace:XX`; duplicate aliases are rejected; plain `II` names the new spellings. Python, in the surface-code noise-parameter tests: a `~XX` replacement branch builds a DEM and samples; `*XX` raises `ValueError` mentioning `~XX`.

## Verification

- `cargo test -p pecos-qec`
- `cargo test -p pecos-rslib p2_weight_parser_tests`
- `cargo clippy -p pecos-qec -p pecos-rslib --all-targets -- -D warnings`
- `cargo fmt -p pecos-qec -p pecos-rslib -- --check`
- `just python-ci-build-test`, then pytest on the QEC surface and DEM test files: 338 passed, 18 xfailed
- `just lint check`

## Context

Second of the separable changes ahead of the state-transition noise channel work.


ciaranra:
Reviewed. No objection to merging, once the branch is refreshed against `dev` (it is currently BEHIND).

## The rename is complete

The main risk in a rename with a hard rejection is a surviving producer: something still emitting `*XX` while the parser now refuses it. I searched the branch for the old form across `.rs`, `.py`, `.md`, `.ipynb` and `.toml`.

Every hit outside the parser is multiplication inside a mathematical comment, `exp(-i*theta/2*XX)` and similar, in `clifford.rs`, `gate_type.rs`, `unitary_rep.rs`, `unitary_matrix.rs`, `sparse_stab*.rs` and `state_vec_sparse_soa.rs`. The only genuine occurrence is `test_noise_parameters.py:83`, which is the rejection test asserting `*XX` raises. So nothing in the repository still produces the old spelling.

## The normalization is the right shape

`trim()` then `to_ascii_uppercase()` happen **before** prefix matching, which is what makes `:replace:` and `~xx` work case-insensitively without a second code path. Prefixes are then tried in order, `:REPLACE:` before `~`, with `*` falling into the migration error.

Duplicate detection keys on the normalized `(replacement, label)` pair rather than the input string, so `~XX` and `:replace:XX` collide as intended. That is the part a simpler implementation would get wrong, since the incoming `BTreeMap` already deduplicates identical strings and hides nothing else.

## Verified by running, not only reading

- All five parser tests pass. Note for anyone reproducing: the `pecos-rslib` test binary needs `libpython3.12.so` on `LD_LIBRARY_PATH`, otherwise it exits 127 with a loader error that looks like a test failure and is not one.
- Mutation check: I made the parser silently accept `*` again, as it did before this change. `rejects_removed_star_replacement_syntax_with_migration_hint` failed and the other four still passed. So the test genuinely pins the breaking behaviour rather than merely exercising it, and it fails for the right reason.

## Two small notes, neither blocking

The duplicate check runs before the label-validity check, so `{"~QQ": 1.0, ":replace:QQ": 1.0}` reports a duplicate rather than an invalid Pauli. Harmless, and arguably the duplicate is the more actionable of the two messages, but worth knowing if someone later wonders why an invalid label produced a duplicate error.

The migration message echoes the uppercased label, so `*xx` is told to use `"~XX"`. That is canonical rather than literal, which seems right; mentioning it only so it is a decision rather than an accident.

## Behaviour change

Worth restating that this rejects existing user configurations at construction. The message names both new spellings, which is the right call for a breaking rename, and the previously unreachable plain-`II` message now being reachable is a genuine improvement rather than a side effect.

PR #681, opened 2026-09-03
Pin exact residual phases where the named-gate convention makes them known

### LINKED ISSUE #673: clifford_simplify matrix verification asserts only projective equivalence where exact phases are now known
## Summary

`crates/pecos-quantum/tests/clifford_simplify_matrix_verification.rs` verifies the rotation-to-named-gate simplification table using `unitaries_equiv`, which compares **up to global phase**. That was the strongest available check when named gates *were* the rotation representatives. It no longer is.

Since the named-gate convention work (#612, #616, #620, and the two-qubit roots), every named discrete gate has an exact phase-fixed matrix, and its relationship to the corresponding rotation is a specific, known phase — for example `SZ = e^{i*pi/4} RZ(pi/2)`, `T = e^{i*pi/8} RZ(pi/4)`, `SZZ = e^{i*pi/4} RZZ(pi/2)`. A projective check cannot detect a wrong phase in either the table or the canonical matrices, which is precisely the defect class this file exists to guard.

## Proposed change

Strengthen the file to pin the **exact residual phase** per case rather than asserting projective equivalence, following the pattern already used well in `pecos-selene-general-noise`'s `residual_phase_per_arm` test: compute `phi` such that `U_named == e^{i phi} U_rotation` and assert it against a recorded expected value, while still asserting entrywise agreement after removing that phase so a case wrong by more than a phase still fails.

That converts roughly 44 projective assertions into exact ones and turns the file into a record of the convention's phase relationships, which is useful documentation in its own right.

## Scope note

The simplification table itself is correct and deliberately projective — `try_simplify_rotation` is consumed by Clifford-only and stabilizer paths, and #632 documents that contract. This issue is about the *test* being weaker than the code now permits, not about changing the table's semantics.

Assertion messages in the file were also stale, reading "should equal" for relationships that are only projective; the three two-qubit cases were corrected in #644 and the remaining single-qubit ones should be corrected alongside this work.

ciaranra:
Closes #673.

## Summary

`unitaries_equiv` compares up to global phase. That was the strongest available check when named gates *were* the rotation representatives; since the convention work (#612, #616, #620, #644) every named discrete gate has an exact phase-fixed matrix and a known phase relationship to its rotation, so a projective check can no longer detect a wrong phase in either the simplification table or the canonical matrices -- exactly the defect class that work has been finding.

This converts the assertions where an exact phase is now known, and leaves the ones where projective comparison is genuinely correct.

## Audit

Every call site was classified before any conversion:

**Converted (exact phase is known, projective was merely weaker):** all `rz_*`/`rx_*`/`ry_*`/`rzz_*`/`rxx_*`/`ryy_*` cases, `crz_zero_equiv_identity`, the seven positive `rxy1q_*` cases and four `u_*` cases in `clifford_simplify_matrix_verification.rs` (43 sites); the Pauli multiplication/tensor cases and ten positive `UnitaryRep` cross-type cases in `gate_algebra_matrix_tests.rs` (16 sites) -- `PauliString` tracks its phase exactly and `UnitaryRep` composition has an exact contract.

**Retained (projective is correct):** `crz_pi_not_equiv_cz` and `crz_pi_twice_equiv_cz_squared`, which establish inequivalence *even up to global phase* -- an exact check there would be weaker; `matrix_composition_order_xh_vs_hx`, same reason; all 14 calls in `pass.rs`, whose rewrites have a documented up-to-phase contract; and the `unitary_matrix.rs` rustdoc and helper tests, which document the helper's projective behaviour. `unitaries_equiv` itself is unchanged.

`op_matrix_tests.rs` no longer references the helper -- #612 and #644 had already converted its sites to exact entrywise checks.

## Derivations

Each expected residual was derived from `R_P(theta) = cos(theta/2) I - i sin(theta/2) P` and confirmed by execution, not fitted: `P = exp(i*pi/2) R_P(pi)`; `S_P = exp(i*pi/4) R_P(pi/2)`; `S_P-dagger = exp(i*3pi/4) R_P(3pi/2)` on the unsigned-representative baseline; `Z = i U(0,0,pi)`; `X = i U(pi,0,pi)`; `SZ = exp(i*pi/4) U(0,0,pi/2)`; `SY*Z = exp(i*3pi/4) U(pi/2,0,pi)`. Every derived value matched observation, and every case also asserts entrywise agreement after removing the phase, so a case wrong by more than a phase still fails rather than being reported as a phase.

## Mutation evidence

Multiplying the canonical `SZ` matrix by `exp(i*pi/4)`:

- the new exact `rz_quarter_equiv_sz` **fails**, reporting expected residual `pi/4` against observed `pi/2`;
- the retained projective `pass::tests::matrix_rz_quarter_equiv_sz` **passes** under the same mutation.

That contrast is the point of the change: a wrong phase in a canonical matrix was previously undetectable by these tests and now is not.

## Verification

Debug and release suites for `pecos-core` and `pecos-quantum`; clippy `--all-targets -D warnings`; `cargo fmt --check`; `git diff --check`. Targeted: 45/45 simplification tests, 27/27 algebra tests. Test-only change -- no production files modified.


ciaranra:
Review found three classes of wrong expectation, and one of them uncovered a production defect. Holding this PR until #686 lands, since the two are coupled.

**1. The twelve three-quarter/negative-quarter phases are fitted to a production bug.** `rotation_to_matrix` halves the *unsigned* modular angle, so it returns the negated matrix for every negative-angle rotation — filed as #686. These expectations encode that behaviour, which means they would **reject the fix**. Correct residual under the documented signed convention is `-pi/4`, not `3*pi/4`.

**2. The four `U` residuals are not residuals of `U`.** The tests construct the unphased Euler product `E = RZ(phi) RY(theta) RZ(lambda)`, but PECOS's phase-fixed `U = exp(i(phi+lambda)/2) E`. Against the actual `U3`, the correct residuals are `0, 0, 0, pi/4` — not `pi/2, pi/2, pi/4, 3*pi/4`.

**3. The RXY1Q section never constructs RXY1Q.** It substitutes RX/RY operands. That masks a sign error in the negated-axis case: `RXY1Q(pi, pi) = i X`, so the residual is `-pi/2`, not `+pi/2`. It passes only because `RX(-HALF_TURN)` aliases back to `RX(pi)`.

What the review confirmed as sound: the classification audit (all retained sites correctly retained — both CRZ tests and `XH`/`HX` use `!unitaries_equiv` to establish inequivalence *even up to phase*, and `pass.rs`'s projective contract is genuinely documented), the shared helper's entrywise check before the phase assertion, and the quarter-turn, two-qubit-root, and Pauli-algebra derivations.

The lesson for this PR: the mutation evidence proved the new assertions were *sensitive*, not *correct*. Values fitted to observed output behave identically under mutation while encoding a present error. The corrected expectations must be derived independently and the actual `U3`/`RXY1Q` operands constructed.

ciaranra:
Rebased onto `dev` (which now carries #668) and corrected the pinned residual phases.

Twelve of the pinned constants in `clifford_simplify_matrix_verification.rs` were computed before #668 landed and were stale by exactly a factor of -1: they expected `3*pi/4` where the corrected dense path produces `-pi/4`. They were updated to `-FRAC_PI_4`.

The observed value is the correct one. For any involutory `P`:

```
SPdg = ((1-i)I + (1+i)P)/2 = ((1-i)/2)(I + iP) = e^{-i*pi/4} * (1/sqrt(2))(I + iP) = e^{-i*pi/4} * RP(-pi/2)
```

using `(1+i)/(1-i) = i`. That covers X, Y, Z, XX, YY, and ZZ identically, which is why all twelve shifted by the same amount. `THREE_QUARTERS_TURN` reads as `-pi/2` under the signed principal representative, so the `3pi/2` and `-pi/2` variants coincide -- exactly the behaviour #668 restored.

The thirteenth site, `u_quarter_zero_pi_equiv_h_like`, genuinely is `3*pi/4` and was left unchanged; it passed both before and after.

Mutation evidence is inherent: those same twelve tests failed against the old constants on this merge, and pass now. Verified with `cargo test -p pecos-quantum -p pecos-simulators --release`, cold `cargo clippy -p pecos-quantum --all-targets --all-features -- -D warnings`, and `cargo fmt --all -- --check`.

ciaranra:
Rebased onto `dev` again (now carrying #692 and #706) and re-verified: `cargo test -p pecos-quantum -p pecos-simulators --release` green, cold `cargo clippy -p pecos-quantum --all-targets --all-features -- -D warnings` exit 0, `cargo fmt --all -- --check` clean.

An independent adversarial review of the constant correction returned **SHIP with no must-fixes**. It rebuilt `RP(theta) = cos(theta/2)I - i sin(theta/2)P` by hand rather than through `rotation_to_matrix`/`half_angle_sin_cos`, so the oracle is independent of the code under test, and confirmed each of the six distinct axes (Z, X, Y, ZZ, XX, YY) separately rather than trusting the generalisation. Residual is `e^{-i*pi/4}` in every case, exact.

The control that matters: had the angle been read *unsigned* as +3pi/2, the residual would be exactly `e^{+i3pi/4}` -- the old constant. So the old values were correct under the old convention and the new ones are correct under #668's signed-halving convention. `THREE_QUARTERS_TURN` and `-QUARTER_TURN` are the same `Angle64` bit pattern, so the representation cannot distinguish RZ(3pi/2) from RZ(-pi/2); signed halving resolves it as -pi/2. This is a convention update, not a test adjusted to pass.

The 13th site was confirmed genuinely different for a structural reason: its angles (0, pi/2, pi) all have fraction <= 2^63 so the signed-halving branch never fires, and its 3pi/4 is the sum of two independent phases rather than a halving artifact.

Mutation-checked: reverting all 12 produces exactly 12 failures, one per site, and forcing the 13th to `-FRAC_PI_4` fails it -- so no assertion is vacuous.

PR #697, opened 2026-09-04
Add a native Relay-BP kernel to pecos-bp with dense-reference and external-parity oracles

ciaranra:
## Summary

Adds a native Relay-BP decoder kernel to `pecos-bp` (stacked on #678). Relay-BP is memory belief propagation run in warm-started legs with per-variable signed memory strengths (Müller et al., arXiv:2506.01779); it is the BP variant the staged qLDPC decoders of arXiv:2607.28795 are built from. PECOS previously reached this algorithm only through the external `relay-bp` crate, which returns one best result per decode, seeds one RNG across all decodes, and checks convergence on every row. The native kernel exposes per-leg outcomes, decodes each shot as a pure function of `(graph, config, syndrome, shot_seed)`, and supports both flooding and check-serial schedules.

The external crate stays as a test oracle only (a dev-dependency of `pecos-bp`). Retiring `pecos-relay-bp` and repointing the `relay_bp` / `min_sum_bp` decoder specs is a follow-up.

## Changes

- `crates/pecos-bp/src/relay.rs`: `RelayBp`, `RelayConfig`, `RelayOutcome`, `LegOutcome`, `Schedule`. Leg 0 with scalar `gamma0`, then up to `num_legs` relayed legs with per-variable gamma drawn from `gamma_range` (or `explicit_gammas`), messages reset per leg, posterior warm-started, early stop on a satisfied syndrome, minimum signed prior-LLR cost among converged legs. Seeds are avalanched from `(shot_seed, leg)` before `PecosRng`. Degree-1 checks send a message of certainty magnitude `LLR_SATURATION` (the same 30 that priors are clamped to) instead of an infinity; a non-finite posterior is an `InternalError`; non-binary syndromes and inconsistent configs are rejected before any state mutation.
- `crates/pecos-bp/src/min_sum.rs`: `LLR_SATURATION` named and re-exported; numerics unchanged (the bitwise promotion fixtures pass).
- `crates/pecos-relay-bp`: `explicit_gammas` pass-through for the oracle, with the external crate's set indexing (from 1) compensated so row 0 is the first relay leg; non-finite gammas and invalid intervals rejected before the external `Uniform` can panic.
- Root `Cargo.toml`: `opt-level = 2` in the dev profile for `pecos-bp`, `pecos-relay-bp`, and `relay-bp` (existing workspace pattern); the parity suite went from 454 s serial to 16 s.

## Tests

- `tests/relay_reference.rs`: a deliberately naive dense reference implementation is the normative algorithm. The sparse kernel must match it bit for bit on five toy graphs: both schedules, with and without memory, four relayed legs with cycling explicit gammas, early stop mid-leg, minimum-cost selection with priors above 0.5, a fired degree-1 check with gamma 0 / 0.1 / -0.24, decode-order independence, `gamma_range` honoured, and config validation.
- `tests/relay_parity.rs` against the external crate on the [[144,12,12]] gross-code circuit-level DEM converted exactly from its own test data (`tests/data`, provenance in the README there): leg-0 per-shot comparison across budgets 2/4/8/16 (mixed regime at 8: 9/100 converge on both sides, flags agree 100/100, corrections identical), one and two shared explicit gamma rows over relayed legs (62/100 both sides, 62/62 identical corrections), independent random gammas in aggregate (71 vs 72), and the full operating point as a smoke test (100/100, 0 logical errors, both sides).
- Unit tests in `relay.rs` for saturated degree-1 messages, non-finite posterior rejection, span validation, syndrome validation, seed-stream independence across legs and shots, and signed-cost selection.

## Verification

- `cargo test -p pecos-bp -p pecos-relay-bp`: all pass (8 unit, 4 promotion, 12 reference, 5 parity; 22 + 37 + 2 in the wrapper).
- `cargo clippy --locked --workspace --all-targets -- -D warnings` with and without `--all-features`: clean.
- `pre-commit run --all-files`: pass.
- Two independent adversarial reviews of the kernel and oracles were run; every finding was either fixed (infinite degree-1 messages, seed lattice correlation, oracle sampler bug, smoke-only parity, wrapper row order, non-finite gamma forwarding, bias overflow on iteration 1, unsigned cost) or refuted with the reasoning recorded in the design note.


ciaranra:
Closed the coverage gaps from an independent adversarial review. **The Relay-BP algorithm itself is unchanged** -- the review checked its algebra against upstream `relay-bp` 0.2.2 and could not break it, and confirmed the test fixtures are genuinely independent (upstream's own shipped data, verified by exact float compare). Every change here is about tests that could not fail.

**`Schedule::CheckSerial` now has an independent oracle.** New `crates/pecos-bp/tests/relay_exact.rs` exhaustively enumerates all error patterns on a small Tanner tree and minimises raw negative log probability -- it shares no BP expressions with the kernel, so unlike `relay_reference.rs` it can catch a shared misunderstanding. This was the top finding: deleting the self-exclusion term from `update_variable_serial` (the textbook BP double-counting bug) previously survived every test except a transliteration of the kernel's own algebra.

Verified independently rather than taken on report: applying that mutation now fails `check_serial_matches_exhaustive_minimum_weight_decoding`, and the tree was restored and re-confirmed green afterwards.

**Vacuous coverage removed.** Pre-iteration budgets 2 and 4 had both decoders converging 0/100, so their assertions reduced to `100 >= 98` and `0 >= 0`. Dropped; the remaining budgets 8 and 16 must now each be non-degenerate rather than the guard being satisfied globally. `full_operating_point_smoke` now requires the observed 100/100 instead of `>= 80` with slack.

**The leg-zero parity config is corrected** to `num_sets(0)`, with an added guard that the external decoder ran no relay leg. `PreIter` only short-circuits when leg 0 succeeds, so `num_sets(1)` did not mean what its docstring claimed.

**The unverifiable seed mixing is gone.** The golden-ratio multiply plus murmur3 finalizer in `leg_seed` could be replaced by a plain XOR without any test noticing, and `PecosRng::seed_from_u64` already SplitMix64-expands its input. Deleted rather than papered over with a stronger test for machinery that earns nothing.

**The five intentional divergences from upstream are now pinned by explicit tests** -- degree-1 message magnitude, hard decision at exactly zero, `-0.0` sign handling, which leg's decision a failed decode returns, and the clamped correction weight -- so they are asserted behaviour rather than incidental.

`RelayOutcome::weight` is now `Option<f64>`, absent on failure, so a weight describing a correction that does not satisfy the syndrome is structurally unrepresentable rather than documented. `num_legs == usize::MAX` is rejected, matching how every other loop bound in the file is validated.

One correction to the review worth recording: its `p=9.4e-14` example sits just above the saturation threshold (about 9.3576e-14) and yields LLR 29.995, so it is in fact distinguishable. The test uses `p=9.3e-14`, which exercises the intended behaviour.

Verified: `cargo test -p pecos-bp` green in both debug and release, cold `cargo clippy -p pecos-bp --all-targets --all-features -- -D warnings` exit 0, `cargo fmt --all -- --check` clean.

Issue #704, opened 2026-09-04
StabVec measurement is O(T^2) for structurally divergent terms, the common case for non-Clifford circuits


ciaranra:
## Summary

`StabVecGeneric` measurement is O(T^2) in the term count whenever the decomposition's terms do not share their CH-form structural matrices, and structure diverges under ordinary non-Clifford circuits. Measurement on those circuits is now roughly three orders of magnitude slower than before the correctness fix.

This is the accepted cost of that fix, not a defect in it. The previous speed came from assuming an invariant the code did not hold, and it produced wrong probabilities. See the discussion in the PR closing #694.

## Measurements

Same circuit on both sides, release build, `debug_assertions` confirmed `false`. The circuit is ordinary Clifford+T: `H` on every qubit, then repeated `RZ(pi/4)`, `H`, `CX`.

| qubits | T gates | terms | before | after |
|---:|---:|---:|---:|---:|
| 8 | 6 | 64 | 7.22 us | 502 us |
| 10 | 8 | 256 | 22.4 us | 10.08 ms |
| 12 | 10 | 1024 | 94.1 us | 137 ms |

Single run per point, no warmup, so the multipliers are indicative rather than tight. The quadratic scaling is the load-bearing observation and it matches an independent measurement taken on deliberately divergent decompositions.

## Why divergence is the common case

Terms acquire independent structural matrices as soon as a non-C-type gate is applied to terms whose gamma values differ, which is what happens the first time an `H` follows an `RZ` decomposition. Nothing subsequently re-homogenizes them. Since decomposing non-Clifford rotations into terms is StabVec's reason to exist, this is the primary workload rather than an edge case.

The shared-structure fast paths still apply where the invariant genuinely holds, and gate application is faster than before the fix, since the C-type path now clones one `Arc` per term instead of five.

## Why the obvious fix does not work

Homogenizing `F`/`G`/`M`/`v`/`s` after projection would restore the linear path, but it is not sound: projection can leave terms representing genuinely different residual stabilizer states, so forcing a shared frame changes the state. This was tested and rejected during the #694 work.

## What would need designing

An exact representation that stays cheap for structurally divergent terms. Possible directions, none evaluated:

- maintain orthogonality incrementally as terms diverge, so a coefficient-only norm stays valid by construction
- cache or incrementally update pairwise overlaps across measurements rather than recomputing them
- re-canonicalize terms into a shared frame where an equivalent shared representation provably exists, falling back to the exact path only when it does not
[...]

Issue #700, opened 2026-09-04
Twelve named Cliffords are not phase-exact on StabVec, and StabVec is absent from the phase-exactness conformance test


ciaranra:
## Summary

Twelve named single-qubit Cliffords are not phase-exact on `StabVecGeneric`, while `CHFormGeneric` and `StateVecSoA` are exact for all 24.

Measured against `Clifford::canonical_1q_matrix()` on the input `H·RZ(0.7)|0>`, StabVec mismatches for:

`H5`, `H6`, `SY`, `SYdg`, `F`, `Fdg`, `F2`, `F2dg`, `F3`, `F3dg`, `F4`, `F4dg`

with errors from 0.42 to 1.88.

## Mechanism

These twelve fall through to the `CliffordGateable` trait's projective default decompositions, which are correct up to a global phase but not phase-exact. That is exactly the state `H3` and `H4` were in before the #694 branch overrode them.

`CHFormGeneric` already overrides every one of these twelve with explicit `mul_phase` corrections (`crates/pecos-simulators/src/stab_vec/ch_form.rs:2136-2174` and `2296-2302`). That is the direct precedent for the fix, in the same crate and the same backend family.

## Why it is worth fixing

StabVec is an amplitude-carrying simulator: `amplitude()` and `state_vector()` are public, so a global phase is observable, and a subcircuit can be controlled after the fact. The phase-fixed conventional matrices are the repo standard following the #601/#642 convention work.

## Structural gap that let it persist

`StabVec` is absent from the phase-exactness conformance test in `crates/pecos-simulators/tests/phase_exact_single_qubit_tests.rs`, which covers the five dense simulators and CHForm. The #694 branch's phase-exact StabVec assertions live in the in-file `#[cfg(test)]` module instead.

Adding StabVec to that conformance test is what would have caught this class, and would also catch the sibling frame-composition issue. That is the more valuable half of this fix: without it the same gap reopens for the next gate added.

## Scope

Pre-existing on `dev`. Found during the adversarial review of the #694 branch.

ciaranra:
Sibling: #699 (frame composition phase error for 13 of 36 ordered gate pairs). Adding StabVec to `phase_exact_single_qubit_tests.rs` is the shared fix that would catch both classes.

ciaranra:
[...]

Issue #699, opened 2026-09-04
StabVec frame composition carries a wrong phase for 13 of 36 ordered gate pairs; h3 compensates for it


ciaranra:
## Summary

`StabVecGeneric`'s Clifford frame composition carries a wrong global phase for every ordered gate pair involving `y`, plus `x` then `z` and `z` then `x`. Measured on a superposed input, 13 of the 36 ordered pairs of StabVec's own overridden single-qubit gates disagree with the dense oracle.

The clearest case: `y(q); y(q)` produces `-I` instead of `I` (max amplitude difference 1.414 against the dense oracle). `sz(q)` then `y(q)` is off by exactly `i`, i.e. two eighth-turns of `frame_phase`.

## Why it matters beyond the phase itself

`StabVec::h3` currently compensates for this defect rather than being independently correct. Its override adds `+1` to `frame_phase`, while the trait documentation prescribes `H3 = exp(-i*pi/4) * Y * SZ` (`crates/pecos-simulators/src/clifford_gateable.rs:476-478`), which is `+7`. `h4` uses `+7` and is correct, because `sz` then `x` composes cleanly and needs no compensation.

So `h3` is right today only by double-correcting: its constant absorbs the `y` frame error. Fixing the cocycle in isolation would silently break `h3`, and only one test guards it. The two must be fixed together, or the coupling must be recorded explicitly.

## Detection

Definite-input tests cannot see this. The probe must use a superposed input and compare full complex amplitudes against the canonical matrix table, not against another PECOS backend.

## Scope

Pre-existing on `dev`. Found during the adversarial review of the #694 branch (`stabvec-two-qubit-root-terms`), which fixes `h3`/`h4` but not the underlying frame composition.

Related: the sibling issue on StabVec's remaining projective named Cliffords, and #694.

ciaranra:
Sibling: #700 (twelve named Cliffords still projective on StabVec). Both were found in the same adversarial review and share the same structural gap -- StabVec is absent from the phase-exactness conformance test. Fixing #699 in isolation will break `h3` on the #694 branch unless `h3`'s constant is corrected to `+7` in the same change.

ciaranra:
Two compensating constants now depend on this defect, and both must be revisited together when the cocycle is fixed:

- `StabVec::h3` adds **+1** frame octant.
[...]

PR #705, opened 2026-09-04
Make StabVec measurement exact for structurally divergent terms

### LINKED ISSUE #694: StabVec two-qubit roots give non-normalized and wrong states with multiple terms or a pending RZ
## Summary

`StabVec`'s two-qubit root gates (`sxx`, `syy`, `szz`) produce wrong states once the coefficient decomposition has several terms, or after a pending non-Clifford RZ. Both are measurement-visible. They are almost certainly related and are filed together.

## Defect 1: non-normalized state at >= 4 terms

```
StabVec::new(2)
h(q0); rz(0.37, q0)
h(q0); rz(0.37, q0)
szz((q0, q1))
```

Result: `||psi|| = 1.1206` where `StateVecSoA` gives `1.0`. Also observed: `sxx` at 8 terms gives `0.9904`, `syy` at 16 terms gives `0.9422`. Norm error grows with term count.

Prime suspect: the Arc-uniformity fast guard in `crates/pecos-simulators/src/stab_vec.rs` (around line 482) compares only `terms[0]` against `terms[1]`, so a decomposition whose terms are not uniform past the first two takes a fast path that does not hold.

## Defect 2: wrong state after a pending non-Clifford RZ

```
StabVec::new(2)
h(q0); rz(0.37, q0)
sxx((q0, q1))
```

Observed: `[0.09197+0.49147i, -0.09197+0.49147i, 0.49147-0.09197i, 0.49147+0.09197i]`

Correct: `[0.41256+0.28249i, 0.28249+0.41256i, 0.41256-0.28249i, 0.28249-0.41256i]`

(derived independently from `SXX = ((1+i)I + (1-i)XX)/2` and confirmed against `StateVecSoA`)

The two states are **not related by any global phase** — this is a wrong state, not a phase convention issue. `h3` and `h4` inherit it.

## Scope

Pre-existing on `dev`; not introduced by any current PR. Found during the adversarial review of #668, by execution against an independently derived reference.

## Suggested approach

Start with the uniformity guard: if it can report "uniform" for a decomposition that is not, both symptoms follow — defect 1 from applying a single scalar where per-term treatment is needed, defect 2 from the pending-RZ term structure violating the same assumption. A regression test should sweep term counts (2, 4, 8, 16) for each of the six two-qubit roots against `StateVecSoA`, asserting both norm and full state, and include the pending-RZ case.

ciaranra:
Closes #694.

## What was wrong

`StabVecGeneric` produced non-normalized states and wrong amplitudes once a decomposition held four or more terms, and a wrong state when a pending RZ was outstanding. Five separate defects were involved, four of which were found while fixing the first.

1. **The C-type fast path replaced every term's structure.** It checked only `F` for pointer equality and then overwrote each term's full structure from term 0. The reported reproductions already diverged in `v`/`s` at term 1 while sharing `F`/`G`/`M`. The path now validates the actual `G`/`M` invariant across every term, preserves each term's own `F`/`G`/`v`/`s`, and falls back per term when validation fails.

2. **CH-form root decompositions carried spurious phases.** `szz`, `szzdg`, `sxx`, and `sxxdg` each applied an extra `mul_phase`. Under the conventional-matrix standard those decompositions are already exact: `CZ · (S ⊗ S)` is `diag(1, i, i, 1)`, and `H·S·H` is the conventional `SX`, so `CX · SX(q) · CX` needs no residual factor. Combined with the fast path applying the phase to term 0 only, the stray factor became a per-term **relative** phase rather than a global one. That is the mechanism behind this issue.

3. **Pending RZ was applied after SXX/SYY instead of before.** Non-commuting pending rotations are now materialized first; SZZ/SZZdg keep their commuting optimization.

4. **H3/H4 needed their own pending-RZ handling** and exact frame-phase corrections.

5. **Forced projection did not zero incompatible terms.** CH-form's `mz_forced` honours a "force only nondeterministic outcomes" contract, which is correct for CH-form standing alone but wrong when StabVec uses it as a per-term projector. A strict `project_z` entry point was added; `mz_forced` and `mz` are unchanged.

## Measurement correctness

An adversarial review then established that the issue's headline symptom was still reachable through the public `mz`, for three further reasons, all pre-existing:

- Renormalization used `sum |c_j|^2` on the assumption that distinct gamma implies orthogonal terms. It does not.
- `merge_identical_terms` merged terms sharing a gamma/omega key without requiring shared structural matrices.
- The `n >= 7` probability path applied term 0's shared constraints to the whole sum without a structure guard, giving silently wrong measurement statistics.

All three are fixed. Norm and `Pr(0)` are now computed from the same pairwise CH-form overlaps, using the precomputed shared-structure path only when pointer equality proves its precondition. Merging requires shared `F`/`G`/`M`/`v`/`s`. Empty post-projection decompositions receive a zero-coefficient sentinel instead of relying on a `debug_assert`, which previously would have panicked in release.

Homogenizing structure after projection was considered and rejected: projection can leave terms representing genuinely different residual stabilizer states, so forcing a shared frame would change the state.

Randomized harness, 3,000 three-qubit circuits and 6,000 forced projections against `StateVecSoA`:

| | before | after |
|---|---:|---:|
| wrong states | 4,166 | 0 |
| non-unit norms | 2,135 | 0 |
| worst norm error | 0.894284 | 8.88e-16 |

## Performance regression

Exact measurement over structurally divergent terms is O(T^2), and structure diverges as soon as an `H` follows an `RZ` decomposition, which is ordinary for non-Clifford work. Measured on the same Clifford+T circuit before and after, release build, `debug_assertions` confirmed off:

| qubits | T gates | terms | before | after |
|---:|---:|---:|---:|---:|
| 8 | 6 | 64 | 7.22 us | 502 us |
| 10 | 8 | 256 | 22.4 us | 10.08 ms |
| 12 | 10 | 1024 | 94.1 us | 137 ms |

The shared-structure fast paths for gates, projection, and probability remain, and gate application is faster than before because the C-type path now clones one Arc per term instead of five. But measurement on divergent structure is substantially slower, and the cost is quadratic.

This is deliberate: the previous speed came from an invariant the code did not hold, and the resulting probabilities were wrong. A representation that is both exact and fast for divergent structure is separate design work, tracked in #704.

## Also in this change

- Dead C-type omega propagation and the representative search were removed; mutation testing showed both were unreachable once the phases were corrected.
- Debug-only contract assertions were added for the `F`/`G`/`v`/`s` invariant the C-type path relies on.
- Redundant outer SYY/SYYdg pending-RZ flushes were removed; `sxx` performs the first non-commuting operation and the required flush.
- CHForm root coverage extended to reversed, distant, and batched pairs.

## Verification

`cargo test -p pecos-simulators` passes in debug and release, including doctests. `cargo clippy -p pecos-simulators --all-targets --all-features -- -D warnings` passes cold. `cargo fmt --all -- --check` passes.

Every guard is mutation-verified: reverting the C-type uniformity check, the merge structure guard, the probability structure guard, coefficient-only normalization, the zero-term sentinel, and each contract assertion independently all produce a failing test.

Two pre-existing defects found during review are filed separately as #699 and #700 and are not addressed here.

### DIFF
--- a/crates/pecos-simulators/src/stab_vec.rs
+++ b/crates/pecos-simulators/src/stab_vec.rs
@@ -172,12 +172,70 @@ impl<S: IndexSet, R: SeedableRng + Rng + Debug + Clone> StabVecGeneric<S, R> {
         }
     }
 
-    /// Merge terms with identical gamma and omega. This is exact (no approximation).
-    /// Terms with the same gamma and omega produce identical amplitudes, so their
-    /// coefficients can be summed. Reduces T without loss of precision.
-    /// Only worth calling when duplicates are likely (e.g., after measurement projection).
+    /// Whether every term shares all CH-form structure that is independent of
+    /// gamma and omega.
+    fn has_shared_projection_structure(&self) -> bool {
+        self.terms.first().is_none_or(|(_, first)| {
+            self.terms[1..]
+                .iter()
+                .all(|(_, ch)| ch.shares_projection_structure(first))
+        })
+    }
+
+    /// Compute the exact norm and Z=0 probability of the represented state.
+    ///
+    /// Both quantities use the same pairwise CH-form overlaps. The shared-
+    /// structure implementation is selected only when pointer equality proves
+    /// its precondition; structurally divergent terms use the general overlap.
+    fn exact_norm_and_prob0(&self, q: usize) -> (f64, f64) {
+        let shared_structure = self.has_shared_projection_structure();
+        let shared_constraints =
+            shared_structure.then(|| self.terms[0].1.precompute_shared_constraints());
+        let omegas: Vec<_> = if shared_structure {
+            self.terms
+                .iter()
+                .map(|(_, ch)| ch.omega_complex())
+                .collect()
+        } else {
+            Vec::new()
+        };
+
+        let mut norm_sq = 0.0;
+        let mut twice_prob0 = 0.0;
+        for (coefficient, ch) in &self.terms {
+            let weight = coefficient.norm_sqr();
+            norm_sq += weight;
+            twice_prob0 += weight * (1.0 + ch.expectation_value_zq(q));
+        }
+        for j in 0..self.terms.len() {
+            for k in (j + 1)..self.terms.len() {
+                let (inner, inner_z) = if let Some(constraints) = &shared_constraints {
+                    self.terms[j].1.inner_product_pair_precomputed(
+                        &self.terms[k].1,
+                        q,
+                        constraints,
+                        omegas[j],
+                        omegas[k],
+                        Some(&self.gamma_diff_qubits),
+                    )
+                } else {
+                    self.terms[j].1.inner_product_pair(&self.terms[k].1, q)
+                };
+                let coefficient_product = self.terms[j].0.conj() * self.terms[k].0;
+                norm_sq += 2.0 * (coefficient_product * inner).re;
+                twice_prob0 += 2.0 * (coefficient_product * (inner + inner_z)).re;
+            }
+        }
+        (norm_sq, 0.5 * twice_prob0)
+    }
+
+    /// Merge shared-structure terms with identical gamma and omega.
+    ///
+    /// The shared F/G/M/v/s precondition makes matching gamma and omega
+    /// sufficient to prove identical amplitudes. Only worth calling when
+    /// duplicates are likely (e.g., after measurement projection).
     fn merge_identical_terms(&mut self) {
-        if self.terms.len() <= 4 {
+        if self.terms.len() <= 4 || !self.has_shared_projection_structure() {
             return;
         }
 
@@ -377,6 +435,15 @@ impl<S: IndexSet, R: SeedableRng + Rng + Debug + Clone> StabVecGeneric<S, R> {
         self.apply_rz_immediate(angle, q);
     }
 
+    /// Materialize a pending RZ together with its current Clifford frame before
+    /// applying a gate that does not commute with Z on this qubit.
+    fn flush_noncommuting_pending_rz(&mut self, q: usize) {
+        if self.pending_rz[q] != Angle64::ZERO {
+            self.flush_cliff_frame(q);
+            self.flush_pending_rz(q);
+        }
+    }
+
     /// Flush the Clifford frame on qubit q by applying its H+S generator sequence.
     fn flush_cliff_frame(&mut self, q: usize) {
         let cf = self.cliff_frame[q];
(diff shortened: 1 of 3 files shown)


ciaranra:
Round 4, after an independent adversarial review of the previous state.

**The measurement no longer recomputes a value it already has.** After projecting onto outcome b, the projected norm is exactly that outcome's probability (for orthogonal projectors, `||P_b psi||^2 = <psi|P_b|psi>`), which `measure_qubit` computes for the outcome decision. Renormalization now reuses `prob0` (or `state_norm_sq - prob0`) instead of paying a second O(T^2) Gram pass.

Implementing it surfaced a subtlety a naive substitution would have got wrong: CHForm returns a *normalized* representative for nondeterministic projection, so the `1/sqrt(2)` was moved into the StabVec coefficient to preserve both the identity and CHForm's normalization invariant.

Public `mz`, release, median of three:

| terms | shared before | shared after | divergent before | divergent after |
|---:|---:|---:|---:|---:|
| 256 | 2.475 ms | 16.8 us | 15.6 ms | 7.31 ms |
| 1024 | 44.6 ms | 69.2 us | 252 ms | 114 ms |
| 4096 | 760 ms | 262 us | 4.06 s | 1.83 s |

Shared structure improves 147x to 2,897x. Divergent structure improves about 2.2x; the remaining cost is the probability Gram pass, which is intrinsic to exactness under divergent structure and is tracked in #704.

**All 24 named single-qubit Cliffords are now phase-exact on StabVec**, and `StabVec` has been added to `crates/pecos-simulators/tests/phase_exact_single_qubit_tests.rs` covering all 24 on both the single-qubit and three-qubit multi-target paths, comparing full complex amplitudes. That test is the structural half of #700 and is what would have caught the remaining defect.

**H6's fix is compensating, not independently correct.** Its default `sx(q).y(q)` composition comes out `-i` times canonical because of the Y-frame cocycle defect in #699, and H6 adds `+2` frame octants to cancel it -- the same shape as H3's existing `+1`. Both are recorded on #699 so a future cocycle fix removes them together; the new conformance test makes that failure loud rather than silent.

The `diff_qubits` contract is now enforced by a debug assertion at the callee boundary, covering every caller, rather than resting on a non-local argument.

Correctness is unchanged: 6,000 forced projections, 0 wrong states, 0 non-unit norms, 0 NaN/inf. Worst norm-squared error moved from 2.220e-15 to 2.649e-13 -- reusing the measured probability inherits that value's error instead of recomputing it. Still far inside tolerance, but it did move.

Verified: `cargo test -p pecos-simulators` green in **both** debug and release (the C-type contract assertions are `#[cfg(debug_assertions)]`, so release alone does not exercise them), cold `cargo clippy -p pecos-simulators --all-targets --all-features -- -D warnings` exit 0, `cargo fmt --all -- --check` clean. Mutation-checked: removing H6's `+2` fails the new conformance test at 1.127 (single-qubit) and 0.972 (three-qubit multi-target).

PR #709, opened 2026-09-04
Build pecos-stab-tn at opt-level 2 in dev and test profiles like the other simulation crates

ciaranra:
## Why

After #683/#684 the PR gate's critical path is `pr-core-rust`, and inside it the workspace test phase: 2486 CPU-seconds of test work on a 4-core runner, ~10.4 min however it is scheduled (a nextest trial, #698, confirmed the phase is CPU-bound, not serialization-bound). The two largest contributors are the surface-code LER equivalence test and the `pecos-stab-tn` test binaries. The root `Cargo.toml` already builds the simulation crates at `opt-level = 2` in dev/test builds for exactly this reason, but `pecos-stab-tn` was not on the list, so its tensor-network code ran unoptimized under test.

Measured locally (14 cores):

| | opt-level 0 (today) | opt-level 2 |
|---|---|---|
| `pecos-stab-tn` lib tests (352) | 159 s | 5.6 s |
| all `pecos-stab-tn` test targets (lib, verification, exact_default_measurement) | ~220 s serial on CI | 8 s |
| extra compile for the crate | | ~33 s |

Also tried `[profile.dev.package."*"] opt-level = 2` for every external dependency: no further gain on stab-tn, 6% on the LER test, and every dependency recompiles slower. Not taken.

## What

One entry in the existing `[profile.dev.package.*]` block plus a comment with the numbers. `profile.test` inherits the per-package overrides (verified by reading the rustc invocations: `-C opt-level=2` for the listed crates in `cargo test` builds). Debug assertions stay on. Crates that depend on `pecos-stab-tn` (the stab-mps and mast Selene plugins, pecos-rslib-exp, benchmarks) get the optimized crate in their debug builds.

## Verification

- `cargo test --locked -p pecos-stab-tn` with the change: lib 352 passed / 8 ignored, verification 95 passed / 9 ignored, exact_default_measurement 9 passed / 20 ignored; 46 s wall including the 37 s compile.
- pre-commit on `Cargo.toml`: clean.
- Measured on CI (gate run 33901282224, warm Rust cache): `pr-core-rust` 17 min end to end, down from 25 min on the previous warm-cache run and ~30 min before #683. Workspace test phase 5.1 min (per-binary times sum to 284 s, was 656 s; longest binary is now the LER test at 191 s), follow-up phases 1.6 min, compile 6.6 min.


ciaranra:
Verified independently. Both load-bearing claims hold.

## `profile.test` really does inherit the override

This is the part the change depends on, so I checked the actual rustc invocation rather than trusting the profile documentation.

On current `dev`, `cargo test -p pecos-stab-tn` compiles the crate with **no `-C opt-level` flag at all**, so it builds at the default 0. With the one-line addition, the same command compiles it with:

```
-C opt-level=2
-C debug-assertions=on
```

So the override reaches the test profile, and the claim that debug assertions stay enabled is correct — worth confirming explicitly, since an optimisation change that silently disabled them would weaken every test in the crate.

## The speedup reproduces

Same machine, same binary, before and after:

| | wall | result |
|---|---:|---|
| current `dev`, opt-level 0 | 63.9 s | 352 passed, 8 ignored |
| with the change, opt-level 2 | 2.3 s | 352 passed, 8 ignored |

That is 28x, matching the ratio in the description (159 s to 5.6 s) on a faster host. Same pass and ignore counts on both sides, so nothing is being skipped to achieve it.

## Reads correctly otherwise

The crate belongs on that list on the same grounds as the entries already there, and the comment records the measurement and its date, which is what makes the entry auditable later. Rejecting `[profile.dev.package."*"]` because it slows every dependency recompile for no gain on this crate is the right call, and worth having written down so it is not retried.

No objection to merging.

PR #717, opened 2026-09-05
Make SparseBinaryMatrix copy-on-write per row and column instead of per matrix

ciaranra:
Phase 1 of the #704 performance work. **Pure optimisation: no amplitude, outcome, or probability changes.**

## Correctness

The gate is the harness landed in #716, which cross-checks `StateVecSoA` on full complex amplitudes and pins exact values with `f64::to_bits()` equality. It was committed before this work specifically so this change could not be tuned against it.

`crates/pecos-simulators/tests/data/stab_vec_correctness_bits.txt` is **byte-unchanged** -- verified against an md5 recorded before the work began, and again against `dev`'s copy after merging. The harness passes including its slow 4,096-term pinned case. A one-ULP difference fails that pin, so bit-identity here is checked rather than asserted.

## What changed

`SparseBinaryMatrix` now uses row- and column-granular `Arc<S>` copy-on-write. Matrix clones share unchanged sets and detach only the rows or columns actually mutated, instead of cloning the whole `Vec<BitSet>`. Added allocation-free `col_xor_assign` / `col_xor_from`, and `right_cx`, `right_cz` and `right_s` in `ch_form.rs` now use them rather than allocating a temporary `Vec<usize>` per call.

It also **removes two `unsafe` blocks**: an existing row-aliasing workaround is replaced with safe split borrowing.

## Measurements

**The speedup is confined to circuits that reach the rewritten helpers.** An adversarial review
established this, and the original numbers here overstated the scope:

| workload | before -> after | |
|---|---|---:|
| H+T divergent, 2048 terms (**calls none of the changed code**) | 3.398 -> 3.440 s | +1.2% |
| H+T divergent, 4096 terms | 13.624 -> 13.811 s | +1.4% |
| mixed Clifford+T, 400 circuits, 8q depth 70 (**does** call it) | 6.48 -> 5.48 s | **-15.4%** |

Pure H+T divergent circuits regress slightly: `Vec<S>` -> `Vec<Arc<S>>` adds a dependent load and
reduces contiguity in `inner_product_fast`, which the corrected profile puts at 91.5% of runtime.

The mechanism is **eliminating the whole-`Vec<BitSet>` deep copy per matrix clone**, not allocator
time. The original description attributed the win to allocation churn while quoting a profile showing
allocator routines at ~4.6% -- those two statements cannot both be true.

Shared-structure timings were also measured and are reported rather than hidden: +9.7% at 256 terms,
+53.5% at 1,024, -3.2% at 4,096, +3.8% at 16,200 -- sub-millisecond in absolute terms except at
16,200 (3.73 ms -> 3.87 ms).

## A correction to #704's profile

The profile that motivated this phase was wrong, and the record on #704 has been corrected.

It was taken with `perf`'s default hardware `cycles` event on an Intel i9-14900KF -- a hybrid CPU with 8 P-cores and 16 E-cores, where that event needs per-PMU handling. Re-profiled with the `cpu-clock` software event:

| symbol | reported (cycles) | corrected (cpu-clock) |
|---|---:|---:|
| `inner_product_fast` | 44.7% | **91.5%** |
| allocator routines | ~32% | **~4.6%** |

So allocation churn was never a third of the runtime. This change is still worth landing on its measured result and the removed `unsafe`, but the honest conclusion is that **the O(T^2) Gram pass is not part of the problem, it is essentially all of it** -- which is what #704 now says.

## Verification

`cargo test -p pecos-simulators` green in both debug and release, the slow pinned harness case passes explicitly, cold `cargo clippy -p pecos-simulators --all-targets --all-features -- -D warnings` exit 0, `pre-commit run --all-files` exit 0 with nothing modified.


ciaranra:
Closed the adversarial review's findings.

## The correctness gate now executes the changed code

The review's central finding was that the harness did not exercise this diff: `right_cx`/`right_cz`/`right_s` and both new column primitives were called **zero** times by the full corpus, so "reference data byte-unchanged, therefore bit-identical" was vacuously true for the `ch_form.rs` half and both new primitives. **12 of 13 injected defects survived.**

Three corpus cases were added -- CX/CZ/S interleaved with T and H, which is what produces the multi-element pivot sets `right_cx` requires. They now execute `right_cx`/`right_cz`/`right_s` **152/90/74** times and the column primitives **456/254** times.

Each new case passed the `StateVecSoA` oracle at the unchanged `1e-12` tolerance **before** its bits were pinned:

| case | checkpoints | max error |
|---|---:|---:|
| `mixed_cx_pivots` | 14 | 1.671e-16 |
| `mixed_cz_phase_pivots` | 20 | 3.342e-16 |
| `mixed_shared_then_divergent` | 27 | 1.755e-16 |

The eight existing fast cases and the slow case are **byte-unchanged** -- the reference file gained 64 lines and lost none.

Independently re-verified: deleting `col_xor_assign`'s row-view update, which previously survived everything, now fails both the new `matrix_cow::cow_col_xor_assign` regression **and** the corpus itself (`mixed_cx_pivots/10:H(2)`, off by 0.6533). Restored and re-confirmed green.

## Aliasing coverage

12 per-mutator copy-on-write regressions added -- mutate a clone, assert the original is untouched -- one each for `set`, `toggle`, `row_xor_assign`, `row_xor_from`, `col_xor_assign`, `col_xor_from`, `swap_rows`, `swap_cols` and the rest. Previously only `toggle` was covered, while removing the detach from any of the others survived the unit tests in both profiles. 14 distinct mutants are now rejected.

Also: `debug_assert_consistent` is bidirectional (it only checked rows against cols, which is the direction column-primary primitives are blindest in), `col_xor_from` and `row_xor_from` have dimension checks, the `right_cx(q, q)` precondition is asserted at the call site, and `reset_to_zero`/`reset_to_identity` no longer deep-copy contents they immediately overwrite.

## Corrected performance narrative

The review was right that the original description was wrong in attribution and scope:

| workload | before -> after | |
|---|---|---:|
| H+T, 2048 terms (**calls none of the changed code**) | 3.398 -> 3.440 s | +1.2% |
| H+T, 4096 terms | 13.624 -> 13.811 s | +1.4% |
| mixed Clifford+T (**does** call it) | 6.48 -> 5.48 s | **-15.4%** |

The gain comes from eliminating whole-`Vec<BitSet>` deep copies during matrix cloning, not from allocator time -- allocation churn alone cannot explain a 12-15% delta against a profile showing allocator routines at ~4.6%. Pure H+T divergent circuits regress slightly from the added indirection and reduced contiguity in `inner_product_fast`. The PR description above will be corrected before merge.

## Verification

1,238 release tests and 1,242 debug tests pass, the slow pinned case passes in both profiles, cold clippy `--all-targets --all-features -- -D warnings` exit 0, and `pre-commit run --all-files` exit 0 with nothing modified.

PR #724, opened 2026-09-05
Correct the deferred Clifford frame's CX and CZ phase rules for its Pauli convention

### LINKED ISSUE #721: StabVec returns the wrong global phase for some pure-Clifford two-qubit compositions
## Summary

`StabVecGeneric` returns the wrong global phase for some pure-Clifford circuits. Three gates are enough, with no rotations involved.

## Reproducers

```rust
let mut sv = StabVec::builder(2).seed(1).build();
sv.x(&[0.into()]);
sv.x(&[1.into()]);
sv.cz(&[(0.into(), 1.into())]);
```

| basis | StabVec | StateVecSoA |
|---|---|---|
| `\|11>` | **+1.0000** | **-1.0000** |

`StateVecSoA` is correct, and this is checkable by hand: `X (x) X \|00> = \|11>`, and `CZ \|11> = -\|11>`, so the amplitude must be `-1`.

A second, structurally different case:

```rust
sv.x(&[0.into()]);
sv.cx(&[(0.into(), 1.into())]);
sv.cz(&[(1.into(), 0.into())]);
```

Same result: StabVec `+1`, StateVecSoA `-1` on `\|11>`.

## Scope

Found in a randomized sweep of 200 five-qubit depth-30 Clifford+T circuits: **12 of 200 disagree with `StateVecSoA` on raw amplitudes, and all 12 are pure global `-1` factors** -- after removing the phase, the worst remaining amplitude error is exactly `0`. So this is the entire residual disagreement between the two simulators once #720 is fixed.

T gates are not required. The reduced examples above are Clifford-only.

## Why it matters

This repo treats global phase as observable for amplitude-carrying simulators, and has been converging on that standard: #601 and #668 made the named gates phase-exact, #705 added `StabVec` to `phase_exact_single_qubit_tests.rs` for all 24 named single-qubit Cliffords, and #714 fixed the measurement path. `amplitude()` and `state_vector()` are public, and a subcircuit can be controlled after the fact.

## Why the existing tests miss it

`phase_exact_single_qubit_tests.rs` pins **single-qubit** gates only. These failures are two-qubit compositions -- the phase is correct for each gate individually and wrong for the product. Nothing pins two-qubit composition phase for `StabVec`.

The structural fix is a two-qubit phase-exactness conformance section, in the same spirit as the single-qubit one, comparing full complex amplitudes against the canonical matrix table rather than against another PECOS backend.

## Relationship to other issues

Distinct from #720 (a pending rotation conjugated by a two-qubit gate after `x()`/`y()`), which produced genuinely wrong states rather than a global phase, and which is fixed separately. With that fix in place these global-phase cases are all that remain.

ciaranra:
Closes #721.

## The defect

`CliffordFrame` represents Paulis as `E(x,z) = Z^z X^x`, so its `Y` representative is `iY`, not the standard `Y`. The CX and CZ phase rules were written against **standard-Y signs**, so a deferred frame carrying `X` or `Y` picked up the wrong scalar when it propagated through a two-qubit gate.

Three Clifford gates reproduce it -- no rotations:

```rust
sv.x(&[0.into()]); sv.x(&[1.into()]); sv.cz(&[(0.into(), 1.into())]);
```

`|11>` came out `+1` where the answer is `-1`. Checkable by hand: `X (x) X |00> = |11>` and `CZ|11> = -|11>`.

## The correction

From the canonical conjugations under this convention:

- **CX** preserves Z-before-X ordering, so its phase correction is **zero**.
- **CZ** introduces one anticommutation when both input X bits are set, giving **`(-1)^(x1*x2)`**.

`StabVec` materialises `E^T` because it executes the generator sequence in listed order; transposing `U† E U = w^k E'` gives `U E^T U† = w^k E'^T`, which also explains the existing root-phase conventions and dagger signs.

The shared `PauliFrameGate` transition now updates both frames and their scalar together, including convention and dagger handling, and **both** amplitude-carrying frame consumers use it -- `state_vec_sparse_soa.rs` previously carried duplicated phase arithmetic and is migrated onto the same path rather than fixed twice.

## Per-path audit

| path | verdict |
|---|---|
| `cx` | fixed -- standard-Y signs removed; correction is zero |
| `cz` | fixed -- missing signs for XX and YY frames |
| `cy` | already correct: materialises frames before applying |
| `swap` | already correct: the three CX sign errors cancel; SWAP preserves total Y count |
| `iswap`, `iswapdg` | affected through their CX decompositions; fixed |
| `g`, `gdg` | affected through CZ; fixed |
| `sxx`, `sxxdg` | already correct: element-convention phase `6(z1 xor z2)`, negated for the dagger |
| `syy`, `syydg` | already correct: composed S/SXX/Sdg phases and dagger reversal are exact |
| `szz`, `szzdg` | already correct: phase `2(x1 xor x2)`, negated for the dagger |

## The structural half

`phase_exact_single_qubit_tests.rs` pinned **single-qubit** gates only, which is exactly why this survived: every gate's phase was individually right and the *composition* was wrong.

It now also covers all 14 named two-qubit Cliffords against the **canonical matrix tables** -- not against another PECOS backend -- across all 24x24 deferred local Cliffords, four orthogonal superposed inputs (a definite input cannot see a global phase), `(0,2)` and `(2,0)` orderings, all two-gate products with Pauli frames, and disjoint batched pairs. The canonical tables for all 14 two-qubit Cliffords were added to `pecos-core` and are themselves cross-checked against the existing symbolic unitary definitions.

## Results

Seeded sweeps against `StateVecSoA`, five qubits, depth 30, tolerance `1e-12`:

| sweep | before (raw / phase-aligned) | after |
|---|---:|---:|
| 200 Clifford circuits | 10 / 0 | **0 / 0** |
| 200 Clifford+T circuits | 6 / 0 | **0 / 0** |

Independently re-verified: a 400-circuit sweep over 16 gates including CY, SWAP, iSWAP and the two-qubit roots gives **0/400 raw mismatches, worst error exactly 0**. Both reproducers now return `-1`.

**The existing randomized differential test now asserts raw agreement.** It previously had to concede global-phase differences and assert only on failures surviving phase removal -- that concession is no longer needed, which is the clearest signal this closes the class.

Mutation-verified: restoring the old CX/CZ rules fails the new conformance tests, including CZ after `(X,X)` and a CX product after `(Y,Y)`, with maximum amplitude error about 0.7071.

## Verification

1,261 release tests and 1,265 debug tests pass across `pecos-simulators` and `pecos-core`, the slow pinned corpus passes in both profiles, cold clippy `--all-targets --all-features -- -D warnings` exit 0, and `pre-commit run --all-files` exit 0 with nothing modified.

The #716 harness and its reference data are **untouched** -- no pinned amplitude changed, and `regenerate_reference_data` was never invoked.

### DIFF
--- a/crates/pecos-core/src/clifford.rs
+++ b/crates/pecos-core/src/clifford.rs
@@ -269,6 +269,76 @@ impl Clifford {
         self.num_qubits() == 2
     }
 
+    /// Returns the canonical phase-fixed matrix for a named two-qubit Clifford.
+    ///
+    /// Rows and columns use `|q0 q1>` order, with q0 the most significant bit.
+    /// Controlled gates embed the canonical Pauli table; roots reuse the
+    /// `GateType` table. SWAP exchanges `|01>` and `|10>`, iSWAP multiplies
+    /// those exchanges by i, and G is CZ (H tensor H) CZ. Daggers are adjoints.
+    #[must_use]
+    pub const fn canonical_2q_matrix(self) -> Option<crate::gate_type::TwoQubitGateMatrix> {
+        let mut matrix = [0.0; 32];
+        match self {
+            Self::CX | Self::CY | Self::CZ => {
+                let pauli = match self {
+                    Self::CX => gate_type_matrix(GateType::X),
+                    Self::CY => gate_type_matrix(GateType::Y),
+                    _ => gate_type_matrix(GateType::Z),
+                };
+                matrix[0] = 1.0;
+                matrix[10] = 1.0;
+                let mut row = 0;
+                while row < 2 {
+                    let mut col = 0;
+                    while col < 2 {
+                        let dst = 2 * (4 * (row + 2) + col + 2);
+                        let src = 2 * (2 * row + col);
+                        matrix[dst] = pauli[src];
+                        matrix[dst + 1] = pauli[src + 1];
+                        col += 1;
+                    }
+                    row += 1;
+                }
+            }
+            Self::SWAP | Self::ISWAP | Self::ISWAPdg => {
+                matrix[0] = 1.0;
+                matrix[30] = 1.0;
+                if matches!(self, Self::SWAP) {
+                    matrix[12] = 1.0;
+                    matrix[18] = 1.0;
+                } else {
+                    let sign = if matches!(self, Self::ISWAP) {
+                        1.0
+                    } else {
+                        -1.0
+                    };
+                    matrix[13] = sign;
+                    matrix[19] = sign;
+                }
+            }
+            Self::G | Self::Gdg => {
+                // G is real, symmetric and self-inverse.
+                let entries = [
+                    0.5, 0.5, 0.5, -0.5, 0.5, -0.5, 0.5, 0.5, 0.5, 0.5, -0.5, 0.5, -0.5, 0.5, 0.5,
+                    0.5,
+                ];
+                let mut i = 0;
+                while i < 16 {
+                    matrix[2 * i] = entries[i];
+                    i += 1;
+                }
+            }
+            Self::SXX => return GateType::SXX.canonical_2q_matrix(),
+            Self::SXXdg => return GateType::SXXdg.canonical_2q_matrix(),
+            Self::SYY => return GateType::SYY.canonical_2q_matrix(),
+            Self::SYYdg => return GateType::SYYdg.canonical_2q_matrix(),
+            Self::SZZ => return GateType::SZZ.canonical_2q_matrix(),
+            Self::SZZdg => return GateType::SZZdg.canonical_2q_matrix(),
+            _ => return None,
+        }
+        Some(matrix)
+    }
+
     /// Returns the canonical phase-fixed matrix for a single-qubit Clifford.
     ///
     /// Each representative has its exact Clifford-group order: H is H1 and
--- a/crates/pecos-core/src/clifford.rs
+++ b/crates/pecos-core/src/clifford.rs
@@ -275,6 +275,10 @@ impl Clifford {
     /// Controlled gates embed the canonical Pauli table; roots reuse the
     /// `GateType` table. SWAP exchanges `|01>` and `|10>`, iSWAP multiplies
     /// those exchanges by i, and G is CZ (H tensor H) CZ. Daggers are adjoints.
+    ///
+    /// Covers all 14 two-qubit variants and returns `None` for single-qubit
+    /// variants. Unlike this method, [`GateType::canonical_2q_matrix`] covers
+    /// only the six Pauli roots; its entries for the other eight gates are `None`.
     #[must_use]
     pub const fn canonical_2q_matrix(self) -> Option<crate::gate_type::TwoQubitGateMatrix> {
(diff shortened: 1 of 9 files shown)


ciaranra:
Closed the adversarial review's findings.

## What the review confirmed, and did not need changing

It derived all 14 canonical two-qubit matrices **independently** in Python from the documented semantics -- never from `to_unitary_rep_on_qubits` -- and they match entry-by-entry including global phase, max error 1.1e-16, and also match what all three simulators actually implement. Seven separate table corruptions were each killed by the new conformance tests, so the tables are not self-validating.

It also confirmed the `SparseStateVecSoA` migration is behaviour-preserving (1,500 randomized 5-qubit depth-40 circuits, byte-identical output against `origin/dev`), that no fifth consumer was missed (`push_through_*` now has zero call sites outside `clifford_frame.rs`), and that the "already correct" verdicts hold **in general** rather than on samples -- SWAP's cancellation replayed for all 16 Pauli frame pairs, and the root-gate phase identities checked across 9 gates x 16 pairs x both conventions with zero failures.

## The must-fix: a test that could not see this PR's own defect

`deferred_frames_preserve_phase_across_disjoint_batched_pairs` was **vacuous for exactly the defect class #721 fixes**. Injecting `frame_phases[c] = (frame_phases[c] + 4) % 8` into the *batched* CZ loop -- issue #721 verbatim -- passed the entire `pecos-simulators` suite, exit 0. The same injection on the single-pair path was killed by 9 tests.

The mechanism: the test used exactly **two** pairs and mirrored the local Cliffords across them. Every kernel's phase depends only on symmetric functions of the two frames, so both pairs receive the same delta and a per-pair error contributes `2*delta mod 8` -- which is **0** when delta = 4. The sign error cancelled against itself. `batch_2q_gates_match_sequential` used two pairs as well and shared the hole.

Both now use **three pairs on six qubits**, so a per-pair error contributes `3*4 = 12 = 4 mod 8` and cannot cancel.

Independently re-verified: the same injection now fails **both** tests. Restored and re-confirmed green.

## Also addressed

- `propagate_transposed` is correct only because every `PauliFrameGate` variant has a symmetric canonical matrix. That was a comment-only invariant -- adding an asymmetric variant would have silently produced wrong `StabVec` phases. There is now an exhaustive symmetry assertion over all nine variants, written so a new variant forces an update.
- The `Clifford::canonical_2q_matrix` / `GateType::canonical_2q_matrix` coverage split (14 gates vs 6 roots) is documented on both, since both are `pub` and already called from `pecos-quantum` and `pecos-gpu-sims`.
- `cy`'s doc matrix showed control-as-LSB while the code and every neighbouring doc use control-as-MSB. Corrected to match the code.
- The `StabVecGeneric` doc block sat above a `use` statement, so it documented the import rather than the struct. The import moved above it; the doctest passes.

## Verification

2,546 release tests and 2,550 debug tests pass, cold clippy `--all-targets --all-features -- -D warnings` exit 0, `cargo fmt --all -- --check` clean, and `pre-commit run --all-files` exit 0 with nothing modified (the Codex sandbox could not run Black; it was run outside).

The #716 harness and its pinned reference data remain untouched.

PR #728, opened 2026-09-06
Seed the two unseeded RNG uniformity tests so they cannot fail by chance

ciaranra:
## Summary

`test_randint_uniformity` and `test_choice_uniformity` in `python/pecos-rslib/tests/test_numpy_random_comparison.py` run a chi-square test at `p > 0.01` on unseeded draws. By construction that fails about one CI run in a hundred; it took down the `pr-core-python (rest)` shard on #711 (p=0.0093) for a change that does not touch the RNG.

Every other statistical test in the file already calls `pc.random.seed(...)` "for deterministic test behavior". This applies the same convention to the two that were missed, with seeds not used elsewhere in the file.

## Verification

```
uv run --frozen pytest python/pecos-rslib/tests/test_numpy_random_comparison.py -k uniformity
```

4 passed. `pre-commit run --files` on the changed file: all hooks pass.

### DIFF
--- a/python/pecos-rslib/tests/test_numpy_random_comparison.py
+++ b/python/pecos-rslib/tests/test_numpy_random_comparison.py
@@ -239,6 +239,8 @@ class TestRandintComparison:
 
     def test_randint_uniformity(self) -> None:
         """Test that randint produces uniform distribution."""
+        # Use a fixed seed for deterministic test behavior
+        pc.random.seed(17)
         low, high = 0, 10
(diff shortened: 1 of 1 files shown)


ciaranra:
Superseded by #730, which merged first and seeds the same two tests in the same file. `dev` already has `pc.random.seed(4)` on `test_choice_uniformity` and `pc.random.seed(24)` on `test_randint_uniformity`, which is why this now shows as conflicting. Nothing here is additive, so closing rather than rebasing.

PR #731, opened 2026-09-06
Record typed named results on the QIS route and turn program exits and panics into shot outcomes

ciaranra:
## Why

Guppy programs are moving onto the QIS engine (Selene runtime) as their default route. Two gaps in that engine blocked the move; both are contract defects surfaced by running the full Guppy test tree on the QIS route:

- `result("tag", value)` lowers to `print_bool`, `print_int`, `print_uint`, `print_float` and their `_arr` forms. The QIS FFI exported only `print_bool`, `print_int` and `print_bool_arr`, and `print_int` recorded only the values 0 and 1, silently dropping every other integer. Float and array-typed results could not even be linked.
- `panic(code, message)` called `std::panic!` inside an `extern "C"` function. A Guppy division by zero therefore aborted the whole Python process instead of failing the shot.

## What changes

**Typed named results.** Every `print_*` entry point exists on both string ABIs (the length-prefixed direct form and the Selene-shim plain form, which now forwards `i64`, `u64`, `f64` and their arrays instead of printing them). Storage is typed per tag and follows the declared call types; a tag that mixes bool and integer calls widens to integer, while signed versus unsigned and float versus anything else are loud errors that fail the shot. The JSON transport carries an explicit element type and the reader rejects unknown types. Shot data keeps the existing rule (one element is a scalar, otherwise a vector), bools stay `U32`, integers become `I64`/`U64`, floats `F64`. Programs that only record bools produce byte-identical shot data and traces.

**Detector convention preserved.** An integer result whose elements are all 0 or 1 stores a bool named-result trace and drains its measurement reads exactly as the old `print_int` did for 0/1, so cultivation-style constant detectors keep working in `DetectorErrorModel.from_guppy`. Other integers and all floats store no trace and drain nothing, which keeps the previous provenance attribution for later bool detectors.

**Exits and panics.** Guppy `exit` (signals 0 to 1000) and `panic` (signal + 1000) both lower to the same `panic` symbol and differ only by code. The FFI now records the code and message on the execution context and transfers control back through the shim's existing `setjmp` wrapper. An exit ends the shot normally with the results recorded so far, and later shots run; a panic fails the shot with the program's own message (`Attempted division by 0` for a Guppy division by zero). The shim no longer treats sub-1000 codes as "recoverable" success. The `Drop` mitigation for the macOS TLS teardown hang is kept.

Static operation collection now runs with an execution context too, so typed outputs and program errors are captured there; that context resets pending reads at collection start and is freed at collection return.

## Verification

- `cargo test -p pecos-qis-ffi` (118 unit tests incl. a full first-type by second-type widening matrix, exit/panic classification for codes -1, 0, 1, 999, 1000, 1001, 1002, provenance attribution, context ownership) and `cargo test -p pecos-qis` (51).
- New `tests/guppy/test_qis_typed_results.py`: int, float, bool, array, repeated and measurement-dependent tags with identical register sets across shots; `exit` keeping results across shots; division by zero raising and the process surviving; the Base QIS plugin path through the C shim for `i64`, `u64`, `f64`.
- Full `guppy`, `pecos`, `qec` lanes: 5,107 passed, 0 failed. `pecos-rslib` lane 2,296 passed. `pre-commit`, `cargo fmt`, `cargo clippy --workspace --all-targets -D warnings` clean.
- Three rounds of independent adversarial review; the pinned oracle `test_selene_interface_integration.py::test_qis_trace_capture_supports_leakage_measurement_futures` is unchanged from `dev`.

## Known limits

- A conditional `exit` before a program's last `result()` call yields a per-shot register-count mismatch, the same shape rule that applies to untagged measurements inside loops.
- The exit/panic code convention is guppylang's and is applied at the C ABI to every producer.
- The mixed-type shot error from `pecos-results` names neither the register nor the shot; follow-up.


ciaranra:
Reviewed with two independent arms on separate lenses, one on FFI and memory safety, one on the result contract. I verified the items in the first section against the code myself; the second section is reported to me but unverified, and marked as such. No changes made, and I did not touch the branch.

## Verified

**`setjmp` is called in a forbidden context.** `crates/pecos-qis/src/c/selene_shim.c:469` and `:515` both do `int error_code = setjmp(user_program_jmpbuf);`. C restricts where that macro may appear: the entire controlling expression of a selection or iteration statement, one side of a comparison or `!` in such an expression, or an expression statement possibly cast to `void`. An initializer is not among them, and the standard makes it undefined behaviour (C23 7.13.1.1p4-5).

This is **pre-existing on `dev`** at lines 452 and 500, so it is not introduced here. What changes is that this PR makes the value load-bearing: the wrappers now return `pecos_program_exited() ? 0 : (uint64_t)error_code`, so exit-versus-panic classification depends on a value obtained through UB. The fix is small and local, for example `if (setjmp(user_program_jmpbuf) == 0) { ... }` with the recorded error deciding the failure return. Worth taking while this code is already being changed.

**The trace decision is value-dependent.** Confirmed at `crates/pecos-qis-ffi/src/lib.rs:203-212`: an `I64`/`U64` result traces only when `values.iter().all(|&value| matches!(value, 0 | 1))`. The comment says this is deliberate, and the storage type is unaffected, so I take the intent as understood. The part worth a second look is the downstream consequence the semantics arm reports, in the next section.

**The JSON transport changes shape, not just content.** `crates/pecos-qis-ffi-types/src/named_results.rs:6` derives `#[serde(tag = "type", content = "values")]`, so `{"b":[true]}` becomes `{"b":{"type":"bool","values":[true]}}`. That is a structural replacement rather than an additive field, including for bool-only programs. The description's compatibility claim is about shot data and traces, which looks right by inspection, but the wire format is not backward compatible. The arm reports that the reader on `dev` (`origin/dev:crates/pecos-qis/src/executor.rs:533`) expects `BTreeMap<String, Vec<bool>>` and that `origin/dev:crates/pecos-qis/src/ccengine.rs:1686-1699` only debug-logs a parse failure before falling back to raw measurements, which would make a mixed-version installation lose named outputs quietly. I confirmed the serde attribute; I did not confirm the fallback behaviour.

## Reported to me, not independently verified

I am passing these on because they are specific and checkable, not because I have confirmed them.

**Value-dependence may change certified DEM width.** The semantics arm reports that compiler certification numbers only the emitted runtime traces when obtaining call arities (`python/quantum-pecos/src/pecos/qec/dem.py:765,784`), so an integer array that is binary-valued contributes width two and the same array with a non-binary element contributes width one. It says it executed the certification function directly with occurrences `[None, 0]` and obtained certified widths of three and two, with evaluation then raising `result 'tag' has 3 values; trace provenance expects 2` at `dem_spec.py:302`. If that holds, the same program yields a different certified schema depending on its data, which seems worth confirming before merge.

**Recovered termination may leak program heap allocations.** `heap_alloc` returns untracked `malloc` allocations (`crates/pecos-qis-ffi/src/ffi.rs:1332,1344`); a termination while an array is live jumps past its `heap_free`, and the recovery branch calls a no-op `selene_on_shot_end`. Repeated early exits would accumulate.

**Cross-shot tag types are not unified.** A program that conditionally calls `result("v", True)` on one shot and `result("v", 0)` on another performs no widening, since each shot executes one call; one shot exports U32 and the other I64, and column conversion rejects it. The old `print_int(0)` stored a bool, so this is a behaviour change for conditional code.

**Early exit before the first named output can collide with raw measurement names.** With no named call executed, `get_results()` exports `measurement_<id>` registers, which cannot be distinguished from a program that deliberately names a result `measurement_0`. Exiting and continuing shots would then carry the same key with different meanings, and column conversion would succeed.

**The widening matrix itself checks out.** The arm worked through all sixteen first-type by second-type pairs against `crates/pecos-qis-ffi-types/src/named_results.rs:55,71` and found the implementation matches the description, order-independent, with errors sticky. It rates the existing tests as genuinely covering negative behaviour rather than happy paths, and suggests the gaps are trace assertions on both outcomes of the 1/7 test at `test_qis_typed_results.py:65`, an old reader consuming new bool-only output, and conditional exits before the first result.

## Note

One thing my own framing got wrong, for the record: a Rust `panic!` reaching an `extern "C"` boundary aborts safely on current Rust rather than being undefined; it is a foreign exception entering a non-unwinding Rust boundary that is UB. So the remaining allocation-failure and logger paths inside the new exports are an abort risk, not a soundness hole.

The branch is behind `dev` and will need a refresh regardless.

ciaranra:
Thanks, all folded. Status per item, against the current tip:

**Verified section**
- `setjmp` in an initializer: fixed. Both wrappers now use it only as the controlling expression of an `if`, with the C23 rule cited, and the return status comes from the recorded termination rather than the `setjmp` value.
- Value-dependent trace decision: narrowed. Only a scalar integer 0/1 result stores a bool trace; integer arrays never trace, so certified DEM width no longer depends on data. This is exactly what the old `print_int` did (it had no array form).
- JSON transport shape: kept, and the body now states when a mixed reader and cdylib can occur (pinned FFI path override, stale artifact, partial rebuild). The new reader fails loudly on the old shape.

**Reported section**
- DEM width: the array case is gone with the narrowing above.
- Heap leak on recovered termination: fixed before this comment landed (allocations are owned by the execution context and freed at shot end on normal return, exit and panic; pinned by a 200-exit test).
- Cross-shot tag types: no cross-shot widening, by decision. The conversion error now names the register, the shot and both types, and the body documents that branch-dependent result types must record a consistent type. The `True` versus `0` shape is a behaviour change from the old bool coercion and is listed as such.
- Early exit before the first named output: documented shape limit, now pinned by a test on the exact register-count error. A user tag literally named `measurement_0` colliding with the raw fallback is possible; noted.
- Widening matrix: the trace assertions on both outcomes of the 1/7 test were added, and the outcome/mapping count mismatch (`test_excess_measurement_outcomes_error`'s subject) is now a loud error in the QIS engine as well.

Branch is rebased on your `dev` merge.

PR #732, opened 2026-09-06
Run Guppy and HUGR programs on the Selene QIS route by default and remove the HUGR interpreter from the Python API

ciaranra:
Stacked on #731 (typed named results and exit/panic handling on the QIS route); retarget to `dev` once that merges. Part 1 of 2 of removing the direct HUGR interpreter; part 2 deletes the Rust crate code.

## Decision

PECOS no longer executes HUGR programs itself. HUGR execution belongs to Selene and Guppy. A Guppy or HUGR program handed to `pecos.sim()` is lowered to QIS and runs on the QIS engine with the Selene simple runtime and Helios interface, the same engine `sim(Qis)` uses. Longer term every HUGR concern moves to the Python boundary and the Rust side consumes QIS only; this PR is the first step, and DEM generation from Guppy programs is unchanged throughout (it already traced through Selene and uses only the static HUGR helpers this PR keeps).

## What changes

- `sim(Hugr)` and `sim(Guppy(...))` build a thin holder of the HUGR bytes and configuration. Nothing is lowered or executed at `sim()` time. `.classical(QisEngineBuilder)`, `run()`, `build()` and the operation-trace entry points all lower through one helper that imports `pecos_rslib_llvm.compile_hugr_to_qis` at runtime (failing loudly when the package is missing) and hands the QIS to the engine. `.stack("neo")` on a HUGR program still routes through the HUGR-to-PHIR facade path, which never used the interpreter.
- Removed from Python: `pecos.hugr_engine`, `pecos.engines.HugrEngineBuilder`, `pecos.engines.HugrSimulation`, and the `pecos_rslib` classes behind them. The `Hugr` program wrapper stays.
- zluppy's `ZluppyEngine.run()` goes through `pecos.sim(pecos.Hugr(bytes)).qubits(n)`, with the qubit count taken from the compiler, and the zlup HUGR codegen now emits a module whose entry point is a `main` function definition, which the QIS compiler requires.
- The HUGR and Guppy user guide and the getting-started guide describe the real path; the Rust examples use the QASM facade; doc tests regenerated.

## Behaviour changes (all decided, each stated on purpose)

- Shot results for Guppy/HUGR programs follow the QIS engine's contract: untagged measurements appear as `measurement_<id>`, `result("tag", value)` calls under their tag. The interpreter's `q<n>`, `measurements` and `return` keys are gone. No compatibility adapter.
- `.qubits(N)` is mandatory for Guppy/HUGR programs, as it already was for QIS programs. Qubit capacity is the runtime's; a growing state-vector backend no longer rescues allocation past it.
- Untagged measurements inside loops give a per-shot variable register count on this route; programs should tag their results.
- Finite recursion runs (the interpreter rejected it). A Guppy division by zero raises `RuntimeError` with the program's message.
- WASM foreign objects stay a PECOS capability but are decoupled from HUGR: `.foreign_object()` on a HUGR program is refused with a "not yet wired into the QIS route" error until that wiring lands; QASM keeps it.
- HUGR programs need `pecos-rslib-llvm` and a built Selene runtime at run time.

## Tests

The Guppy test tree is the acceptance battery. 22 files that pinned the interpreter's result shape were rewritten to the contract above with their physics and semantics assertions kept at full strength (verified by an independent assertion-by-assertion audit); results are read through `result()` tags or the specific `measurement_<i>` keys, no fallbacks. 28 skips whose bodies now pass were removed; three remain, each naming its cause. The loop tests in the former interpreter-parity file run on both the default route and the explicit QIS route. Four test functions were replaced by stronger or equal ones; none deleted without an equivalent.

## Verification

Full `guppy`, `pecos`, `qec` lanes green on the stack; `pecos-rslib` lane 2,296; Selene and SLR lanes 461; Python doc tests 308; zlup 888; zluppy 76; `pre-commit`, `cargo fmt`, `cargo clippy -p pecos-rslib -D warnings` clean. Two rounds of implementation review plus a fresh-context verification run.


ciaranra:
Found while reviewing the stacked Rust PR #733; this one belongs here rather than there, since #733's own diff does not touch the Python API.

**An example notebook calls an API this PR removes.** `examples/measurement_throughput_benchmark.ipynb:40` does `from pecos import depolarizing_noise, hugr_engine`, and line 94 calls `hugr_engine().hugr_bytes(hugr_bytes).to_sim().noise(noise).run(shots).to_dict()`. The surrounding prose names it at lines 21 and 300, describing it as the "Traditional" path that interprets HUGR directly.

On `dev`, `hugr_engine` is exported at `python/quantum-pecos/src/pecos/__init__.py:315` and listed in `__all__` at line 423. Commit `92be52e8c` on this branch removes it, so that import fails afterwards.

The notebook's whole purpose is comparing the direct-interpreter path against symbolic sampling, so it is not a matter of swapping one call: the benchmark's "traditional" arm no longer exists. Either the notebook needs rewriting around the QIS route, or it should be removed along with the API it exercises.

Verified by checking the export on `dev` and confirming the removal is in this PR's diff rather than #733's. No changes made.

ciaranra:
Agreed, and fixed in the round pushed after your comment (commit 257173b1 on this branch): the notebook's "traditional" arm is now the Selene QIS route, `pecos.sim(pecos.Hugr(hugr_bytes)).qubits(N).noise(...)`, and the prose describing it as direct interpretation was rewritten to match. The benchmark still compares that arm against symbolic sampling, which is what it was for. The notebook JSON is validated by loading it, since notebooks are not executed in CI.

PR #733, opened 2026-09-06
Delete the direct HUGR interpreter and shrink pecos-hugr to the static loader and result-tag helpers

ciaranra:
Stacked on the Python-side PR (Guppy/HUGR on the Selene QIS route by default); retarget when that merges. Part 2 of 2 of removing the direct HUGR interpreter.

## What changes

- `crates/pecos-hugr` loses the interpreter engine, its builder, the control-flow and handler modules, the loop-boundary test and an orphaned fixture: 29 files, about 17,000 lines. The crate now holds only the static HUGR loader and the result-tag analysis that back the Guppy DEM bindings in `pecos-rslib` (`extract_result_tag_measurements_for_guppy`, `guppy_hugr_has_nontrivial_control_flow`, `resolve_result_tags_for_guppy`). Those two modules are unchanged; the crate's dependencies shrink to `anyhow`, `tket`, `tket-qsystem`.
- `pecos` feature `hugr` no longer depends on `pecos-hugr` and means static HUGR/DAG conversion only; feature `neo` no longer enables `pecos-neo/hugr`; the engine re-exports are gone from the facade and the prelude.
- `pecos-neo` loses its `hugr` feature, the `From<HugrEngineBuilder>` conversion and the interpreter auto-select arm; `sim_neo` now refuses a HUGR program with a message saying HUGR is lowered to QIS at the Python boundary. Its interpreter integration test is deleted.
- The doc-test crate drops its direct `pecos-hugr` dependency and the `hugr` feature; CLI and Justfile messages that promised HUGR execution now say what LLVM actually gates (QIR/HUGR compilation).
- One comment in `pecos-phir` that pointed readers at the interpreter now points at the boundary.

The symbols `HugrEngine`, `HugrEngineBuilder`, `hugr_engine`, `hugr_sim`, `CapturedResult`, `ClassicalValue`, `FutureId`, `ResultValue` and `RngContextId` from `pecos_hugr` exist nowhere in the workspace.

## Verification

Cold `cargo clippy --workspace --all-targets -D warnings` (after `cargo clean` of the changed crates) and the `crates/pecos` lane with `runtime,hugr,neo`; `cargo test -p pecos-hugr` (7), `-p pecos --features runtime,hugr,neo` (85), `-p pecos-neo` (1,350); `pecos-rslib` lane 2,296; `qec` lane 1,632 (the DEM path); Python doc tests 308; Rust doc-test crate 176; `cargo tree -p pecos-hugr --depth 1` shows the three dependencies; `Cargo.lock` diff is removals only. Independent fresh-context verification PASS.


ciaranra:
Reviewed with two independent arms on separate lenses, one on completeness of the deletion and one on whether the QIS route replaces what the interpreter did. I verified each finding below against the code myself; anything I could not confirm is marked as such. No changes made, and I did not touch the branch.

## Verified, worth acting on

**A full-matrix CRz test is collateral removal — it covers code that survives.** `crates/pecos-hugr/tests/crz_boundary.rs` contains two tests. `dynamic_radd_preserves_crz_winding` goes with the interpreter, correctly. But `rust_hugr_crz_boundary_preserves_full_matrix` (line 201) calls the `execute` helper at line 76, and that helper uses `pecos_quantum::hugr_convert::hugr_to_dag_circuit`, not `HugrEngine`. That converter remains public at `crates/pecos-quantum/src/hugr_convert.rs:709`. The test builds the complete 4x4 CRz matrix over five angles including `3*pi` and `tau`, checking a shared global phase, which is exactly the winding behaviour the angle convention work was about. Deleting the file removes that coverage for code this PR keeps. Suggest relocating the test rather than dropping it.

**A Python test still enables logging for a deleted module.** `python/quantum-pecos/tests/guppy/test_for_loop.py:45` sets `RUST_LOG=pecos_hugr::engine=debug`. `engine.rs` is deleted here, so the filter now names nothing.

**A generator script still documents the deleted engine as its purpose.** `scripts/generate_conditional_hugr.py:1` reads "Generate a HUGR file with conditionals for testing HugrEngine". The script is not orphaned, since its output still has consumers, but the stated purpose is now wrong.

## Verified corrections to the description

The claim that the nine symbols "exist nowhere in the workspace" is very slightly overstated: `HugrEngine` survives in the generator docstring above. Separately, `ResultValue` appears 18 times in `crates/pecos-phir-json/src/v0_1/classical_interpreter.rs`, but that is an unrelated type of the same name, not a leftover — worth noting only so a future grep is not misread.

I also checked whether the retained `.hugr` fixtures under `crates/pecos/tests/test_data/hugr/` are orphaned. **They are not.** `python/quantum-pecos/tests/guppy/test_hugr_compiler_parity.py:214` globs every `*.hugr` in that directory, so all of them still have a consumer. A basename search misses this; flagging it in case the same question comes up.

## Verified context, not a defect in this PR

**Classical RNG returns constant zero on the replacement route.** `crates/pecos-qis/src/c/selene_shim.c:395-417` implements `selene_random_seed` and `selene_random_advance` as no-ops and `selene_random_u32`, `selene_random_u32_bounded` and `selene_random_f64` as constant zero. This is **pre-existing on `dev` and untouched by this PR**, and it is the real runtime path rather than a test double: it is compiled to `libpecos_selene_shim.so` by `build_selene.rs:40,179` and is the only implementation of those symbols in the workspace.

The relevance here is that the deleted interpreter had working seeded RNG contexts, so after this stack there is no route with functioning classical RNG for a Guppy program that samples. That is a user-visible regression produced by removing the working alternative, even though the stub is not this PR's doing. It seems worth a deliberate decision on sequencing rather than discovering it later, and the fix belongs in the shim.

## Reported but not verified by me

The capability arm produced a full inventory of the 89 deleted Rust tests with an assessment of whether each has a surviving equivalent. I did not check them individually. Two it rated highest are worth a look:

- `test_excess_measurement_outcomes_error` (`origin/dev:crates/pecos-hugr/src/engine.rs:3159`) rejected an outcome with no queued measurement. It reports that QIS instead pairs mappings and outcomes with `zip` at `crates/pecos-qis/src/ccengine.rs:337`, which would silently discard extras rather than erroring.
- The loop-boundary test's specific guard — completion within ten rounds, exactly three H gates and one measurement — has no equivalent trace assertion it could find, though the scheduler defect it protected against cannot survive removal of that scheduler.

It also reports the old result schema is not backward compatible: raw `qN` values, a combined `measurements` array, and scalar entrypoint returns captured as `return`/`return_<port>` have no QIS equivalent, with named results now `measurement_<result_id>`. I have not confirmed that.

## Sequencing

Worth stating explicitly since the description mentions retargeting: `python/pecos-rslib/src/engine_builders.rs:16,585` and `src/sim.rs:170` on `dev` still name `pecos_hugr::HugrEngineBuilder` and call `pecos_hugr::hugr_engine()`, both of which this PR deletes. A normal merge of the full branch carries the Python parent along, so this is only a hazard if the deletion commit is cherry-picked or squashed away from it.

ciaranra:
Thanks. Per item, against the current tip:

- CRz full-matrix test: relocated byte-identical to `crates/pecos-quantum/tests/crz_boundary.rs` under the `hugr` feature (it only needs `hugr_to_dag_circuit`); `cargo test -p pecos-quantum --features hugr -- crz` lists it.
- `RUST_LOG=pecos_hugr::engine=debug` in `test_for_loop.py`: that whole manual-runner block is removed, on the Python-side PR since the file belongs there.
- Generator docstring: corrected; it now names the two tests that read the fixture.
- Description: the "exist nowhere" sentence is reworded and notes the unrelated `pecos-phir-json` `ResultValue`.
- Fixtures: agreed, the compiler parity test globs that directory; the description says so now.
- Classical RNG: confirmed, and worse than zeros. A Guppy program using `RNG` does not load on the QIS route at all: nothing exports `random_seed`/`random_int`/`random_rng`/`random_float`/`random_advance`, so the process dies with a symbol lookup error. Selene's reference stream for a program seed is a pure function of that seed (measured: seed 42 gives 1085446021, 0, 0.18373215361498296, 1312053665 on every shot regardless of the instance seed), and PECOS's own PCG32 in `pecos-random` already reproduces the first value under the convention pinned in `tests/pecos/unit/test_rng.py`. A separate engine PR, stacked on #731, adds the five exports backed by a context-owned PCG32, forwards the shim's `selene_random_*` to the same state, and pins the four Selene values plus a live selene-sim parity test. Recommendation: merge it before this one.
- `test_excess_measurement_outcomes_error`: the QIS engine now fails a batch whose outcome count differs from the queued count, in #731.
- Sequencing: understood; the stack merges as a whole, parent first.

PR #747, opened 2026-09-11
Fail loudly when a per-gate noise rate key names a Clifford action instead of the scheduled gate

### LINKED ISSUE #746: Per-gate noise rate keys that name a Clifford action instead of the scheduled gate are silently ignored
## Problem

Per-gate noise rates (`p1_gate_rates`, `p2_gate_rates`, and the `PerGateTypeNoise` tables `rates_1q`, `rates_2q`, `rates_1q_per_qubit`, `rates_2q_per_qubits`) are keyed by `GateType` and looked up by each noise location's scheduled gate type, with the scalar rate as the default. Nothing checks that a key matches any location in the circuit.

Since #742 noise locations follow the gate as scheduled. A Selene-traced circuit schedules `RZZ(pi/2)`, `RXY1Q(pi/2, 0)`, and `RZ(pi)`, not `SZZ`, `SX`, and `Z`. A user who writes `p2_gate_rates={"SZZ": r}` against such a circuit gets the scalar `p2` on every `RZZ` and no error. The configuration is silently a no-op.

The surface-code path is unaffected in practice because it lowers traced rotations to named gates before building, so its `"SZZ"`/`"SZZdg"` keys match. The raw-trace entry points (`DemSampler.from_circuit`, `DetectorErrorModel.from_circuit`, `build_adaptive_dem`) are the ones that silently ignore the key.

## Decision

Keys keep meaning the scheduled gate. A key with a non-zero rate that matches no scheduled gate in the circuit but does name the Clifford action of one (`SZZ` for `RZZ(pi/2)`, `SX` for `RXY1Q(pi/2, 0)`, `Z` for `RZZ(pi)`) is a configuration error that names the key, the table, the node, the scheduled gate, and the remedy: key by the scheduled gate, or lower the circuit with `lower_clifford_rotations()` first. A key that matches nothing at all stays allowed, so noise configurations can be shared across circuits with different gate sets. Applying rates by Clifford action was considered and rejected: it would make a second key namespace for one noise table.

The check runs once per build in every builder's preflight (`DemBuilder`, `DemSampler`, `SamplingEngine`, `SamplingEngineBuilder`, `MemBuilder`) over every gate-keyed table.

## Amendment after review

A key that names the Clifford action of a location whose scheduled gate differs is an error even when the key also matches other scheduled gates (a circuit scheduling both `SZZ` and `RZZ(pi/2)` with `{"SZZ": r}` was silently applying `r` to one pair only). Per-qubit keys match on gate and qubit scope; only the tables a build path consumes are validated.

ciaranra:
Closes #746.

## Problem

Per-gate noise rates are keyed by `GateType` and looked up by each noise location's scheduled gate type, with the scalar rate as the default. Nothing checked that a key matched any location. Since #742 noise locations follow the gate as scheduled, so on a Selene-traced circuit, which schedules `RZZ(pi/2)`, `RXY1Q(pi/2, 0)`, and `RZ(pi)`, a rate keyed `"SZZ"`, `"SX"`, or `"Z"` applied to nothing and raised nothing. The surface-code path never hit this because it lowers traced rotations to named gates before building; the raw-trace entry points did.

## Change

Keys keep meaning the scheduled gate. One validator in `dem_builder/types.rs` checks every gate-keyed table in the noise configuration (`p1_gate_rates`, `p2_gate_rates`, and the `PerGateTypeNoise` tables `rates_1q`, `rates_2q`, `rates_1q_per_qubit`, `rates_2q_per_qubits`): a key with a non-zero rate that matches no scheduled gate but names the Clifford action of one (`Named` or `PerQubit` on the location) is a `ConfigurationError` that states the table, the key, the node, the scheduled gate, and the remedy (key by the scheduled gate, or lower the circuit with `lower_clifford_rotations()` first). A key that matches nothing at all stays allowed so a noise configuration can be shared across circuits with different gate sets. Zero rates are ignored. Applying rates by Clifford action was considered and rejected as a second key namespace for one table.

Every builder runs the check in its preflight next to the existing replacement-branch validation: `DemBuilder`, `DemSampler` (including its raw-measurement builder), `SamplingEngine`, `SamplingEngineBuilder`, and `MemBuilder`. Python callers see `ValueError` through the existing mappings. Docstrings for the rate parameters in `dem.py`, `surface/decode.py`, `surface/circuit_builder.py`, and the bindings state the contract.

## Tests

Thirteen Rust tests in `unsupported_dem_gates_tests.rs` cover the mismatch through each builder, the applied rate (`{"RZZ": r}` changes the DEM), scheduled-gate precedence when both `SZZ` and `RZZ(pi/2)` are present, unrelated keys staying allowed, zero rates ignored, the per-qubit case (`RZZ(pi)` with a `Z` key), and every `PerGateTypeNoise` table. Gutting the validator fails eight of them. Python tests on the raw d=3 surface trace: `{"SZZ": r}` raises with a message naming `SZZ` and `RZZ`, `{"RZZ": r}` builds and differs from the scalar model, and `{"SZZ": r}` builds after `lower_clifford_rotations()`.

## Verification

`cargo test -p pecos-qec --no-fail-fast`: 1044 passed. `cargo clippy --workspace --all-targets -- -D warnings` on the touched files: clean. Python `tests/qec` and `pecos-rslib/tests`: 3954 passed. `pre-commit run --all-files`: 17 hooks passed.

## Second commit, after review

An independent review of the first commit found four edge cases in the validator, none a wrong DEM, all fixed here.

- **Mixed gate sets.** A circuit scheduling both `SZZ` and `RZZ(pi/2)` with `{"SZZ": r}` was accepted and silently applied `r` to one pair and the scalar to the other, which is the silent miss this change exists to remove. The rule is now: a non-zero key naming the Clifford action of a location whose scheduled gate differs is an error even when the key also matches other scheduled gates. Keys matching nothing at all stay allowed.
- **Remedy names the right table.** `RZZ(pi)` with `p1_gate_rates={"Z": r}` advised keying by `RZZ`, a two-qubit gate that the Python boundary rejects in `p1_gate_rates`. The message now names the table the scheduled gate belongs in, derived from its arity.
- **Only consumed tables are validated.** When per-gate noise is set, the scalar tables are never read and are no longer validated; the raw-measurement sampler path never applies per-gate noise and no longer validates it; `MemBuilder` reads no gate-rate table and validates none. Each site carries a one-line comment stating the rule.
- **Per-qubit keys carry their qubit scope.** A `rates_1q_per_qubit` or `rates_2q_per_qubits` key for a qubit or pair absent from the circuit matches nothing and is allowed; a key matches a location only when gate and qubits both match, in the pair order the lookup uses.
- Docstrings corrected: `from_guppy` and `GuppyDemBuilder` lower before building, so their keys name the lowered gates; only the raw-circuit entry points see scheduled rotations. `NoiseParameters` in `surface/decode.py` emits `SZZ`/`SZZdg` keys and the surface path lowers first.

Six new Rust tests cover the tightened rule, both remedy messages, absent-qubit keys, and the consume-only rule in each builder; restoring the old first-match exemption fails the mixed-circuit test. Pre-existing and out of scope: `MemStabSim` and `DemStabSim` without per-gate noise forward only the four scalar rates and never carried gate-rate tables. The validator costs under one percent of build time on a d=7, seven-round surface DEM, so it has no index.

Verification after the second commit: `cargo test -p pecos-qec --no-fail-fast` 1051 passed, clippy clean on the touched files, Python `tests/qec` and `pecos-rslib/tests` 3954 passed, pre-commit 17 hooks passed.

## Third commit, after a second review

The second review found that "validate only what the path consumes" had enshrined two silent drops. Raw-measurement mode in `DemSamplerBuilder` never applies per-gate noise, and `MemBuilder` never reads a gate-rate table; both now reject those configurations with a configuration error instead of validating nothing. The mismatch message now distinguishes a key that matches no scheduled gate from one that also matches scheduled gates elsewhere, and the remedy names the table the scheduled gate belongs in, selected by arity through a total function rather than string matching with an assertion. The mixed-circuit test schedules its matching `SZZ` before the measurements so the key genuinely contributes on that pair, and the Python assertions require the remedy table in the message.

Left as pre-existing and tracked separately: per-gate noise supersedes the scalar configuration on `DemBuilder` and `SamplingEngineBuilder` (the sampler builder relies on that by syncing the scalar config to the per-gate base), and the pairing between "reads table X" and "validates table X" is enforced by comments; #752 proposes a validated-noise type that makes it structural.

Verification after the third commit: `cargo test -p pecos-qec --no-fail-fast` 1053 passed, clippy clean on the touched files, Python `tests/qec` and `pecos-rslib/tests` 3954 passed, pre-commit 17 hooks passed.

### DIFF
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -945,13 +945,16 @@ impl<'a> DemBuilder<'a> {
     }
 
     fn validate_noise_configuration(&self) -> Result<(), DemBuilderError> {
-        self.noise
-            .validate_gate_rate_keys(&self.influence_map.locations)
-            .and_then(|()| {
-                self.per_gate.as_ref().map_or(Ok(()), |noise| {
-                    noise.validate_gate_rate_keys(&self.influence_map.locations)
-                })
-            })
+        // Validate exactly what this path consumes: per-gate noise or scalar tables.
+        self.per_gate
+            .as_ref()
+            .map_or_else(
+                || {
+                    self.noise
+                        .validate_gate_rate_keys(&self.influence_map.locations)
+                },
+                |noise| noise.validate_gate_rate_keys(&self.influence_map.locations),
+            )
             .map_err(|error| DemBuilderError::ConfigurationError(error.to_string()))?;
         if let Some(weights) = &self.noise.p2_weights {
             weights
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/builder.rs
@@ -832,7 +832,7 @@ impl<'a> DemBuilder<'a> {
         self.validate_supported_gates()?;
         self.validate_measurement_count()?;
         self.validate_metadata_refs()?;
-        self.validate_replacement_branch_approximation()?;
+        self.validate_noise_configuration()?;
         self.validate_measurement_crosstalk_dem_mode()?;
         self.validate_idle_noise()?;
         self.build_inner()
@@ -868,7 +868,7 @@ impl<'a> DemBuilder<'a> {
     /// replacement-branch configuration, or failed signature conversion.
     pub fn build(&self) -> Result<DetectorErrorModel, DemBuilderError> {
         self.validate_supported_gates()?;
-        self.validate_replacement_branch_approximation()?;
+        self.validate_noise_configuration()?;
         self.validate_measurement_crosstalk_dem_mode()?;
         self.validate_idle_noise()?;
         self.build_inner()
@@ -944,7 +944,15 @@ impl<'a> DemBuilder<'a> {
         Ok(dem)
     }
 
-    fn validate_replacement_branch_approximation(&self) -> Result<(), DemBuilderError> {
+    fn validate_noise_configuration(&self) -> Result<(), DemBuilderError> {
+        self.noise
+            .validate_gate_rate_keys(&self.influence_map.locations)
+            .and_then(|()| {
+                self.per_gate.as_ref().map_or(Ok(()), |noise| {
+                    noise.validate_gate_rate_keys(&self.influence_map.locations)
+                })
+            })
+            .map_err(|error| DemBuilderError::ConfigurationError(error.to_string()))?;
         if let Some(weights) = &self.noise.p2_weights {
             weights
                 .validate_replacement_locations(
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/dem_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/dem_sampler.rs
@@ -467,6 +467,10 @@ impl SamplingEngine {
             return Err(super::DemBuilderError::UnsupportedGate(error.clone()));
         }
 
+        noise
+            .validate_gate_rate_keys(&influence_map.locations)
+            .map_err(|error| super::DemBuilderError::ConfigurationError(error.to_string()))?;
+
         if let Some(weights) = &noise.p2_weights {
             weights
                 .validate_replacement_locations(
@@ -2215,6 +2219,17 @@ impl<'a> SamplingEngineBuilder<'a> {
         if let Some(error) = self.influence_map.unsupported_gate() {
             return Err(super::DemBuilderError::UnsupportedGate(error.clone()));
         }
+        super::types::validate_scalar_gate_rate_tables(
+            &self.p1_gate_rates,
+            &self.p2_gate_rates,
+            &self.influence_map.locations,
+        )
+        .and_then(|()| {
+            self.per_gate.as_ref().map_or(Ok(()), |noise| {
+                noise.validate_gate_rate_keys(&self.influence_map.locations)
+            })
+        })
+        .map_err(|error| super::DemBuilderError::ConfigurationError(error.to_string()))?;
         if let Some(weights) = &self.p2_weights {
             weights
                 .validate_replacement_locations(
--- a/crates/pecos-qec/src/fault_tolerance/dem_builder/dem_sampler.rs
+++ b/crates/pecos-qec/src/fault_tolerance/dem_builder/dem_sampler.rs
@@ -2219,17 +2219,20 @@ impl<'a> SamplingEngineBuilder<'a> {
         if let Some(error) = self.influence_map.unsupported_gate() {
             return Err(super::DemBuilderError::UnsupportedGate(error.clone()));
         }
-        super::types::validate_scalar_gate_rate_tables(
-            &self.p1_gate_rates,
-            &self.p2_gate_rates,
-            &self.influence_map.locations,
-        )
-        .and_then(|()| {
-            self.per_gate.as_ref().map_or(Ok(()), |noise| {
-                noise.validate_gate_rate_keys(&self.influence_map.locations)
-            })
-        })
-        .map_err(|error| super::DemBuilderError::ConfigurationError(error.to_string()))?;
+        // Validate exactly what this path consumes: per-gate noise or scalar tables.
+        self.per_gate
+            .as_ref()
+            .map_or_else(
+                || {
+                    super::types::validate_scalar_gate_rate_tables(
+                        &self.p1_gate_rates,
+                        &self.p2_gate_rates,
+                        &self.influence_map.locations,
+                    )
+                },
+                |noise| noise.validate_gate_rate_keys(&self.influence_map.locations),
+            )
+            .map_err(|error| super::DemBuilderError::ConfigurationError(error.to_string()))?;
         if let Some(weights) = &self.p2_weights {
             weights
                 .validate_replacement_locations(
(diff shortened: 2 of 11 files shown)


ciaranra:
Merging current `dev` (`5530ba878`) into this branch produces one conflict and three test failures. The conflict is mechanical; the failures are a real semantic collision that needs a decision before this can land.

## Conflict: two independent guards, both belong

`MemBuilder::build` gained a rejection guard on both sides at the same spot:

- this branch: reject nonzero `p1_gate_rates` / `p2_gate_rates`
- `dev`: reject `BranchImpact` / `ExactBranchReplay` replacement branches

They are not competing versions of one check. Taking either side alone compiles, passes most tests, and silently deletes a guard. The resolution is to keep both `if` blocks and merge the two doc sentences.

## Failure: the gate-rate guard is now factually wrong

All three failures share one cause:

```
phase_u_review.rs:25  ConfigurationError("MemBuilder applies scalar rates only;
                       nonzero p1_gate_rates or p2_gate_rates are not supported")
```

Failing: `phase_u_measurement_noise_resolves_operation_rate`, `phase_u_explicit_scheduled_override_takes_precedence`, `measurement_noise_shares_two_qubit_rates_and_idle_policy`.

The guard's premise -- that `MemBuilder` ignores per-gate rates -- was true when this branch was written and is no longer true. `MemBuilder`'s one-qubit fault processing calls `self.noise.rates_1q_for_operation(loc.gate_type, loc.noise_gate_type)`, so MEM resolves per-gate rates. #760 (`b2ce0a4fc`, "resolve phase-shaped U noise consistently across MEM, NEO, and samplers") made that change and added `crates/pecos-qec/tests/phase_u_review.rs`, which landed on `dev` after this branch forked -- so this branch never saw those tests.

The tests demonstrate the consumption independently:

```rust
for (base, rz, expected) in [(0.3, 0.0, 0.0), (0.0, 0.3, 0.2)] {
    let noise = NoiseConfig::new(base, 0.0, 0.0, 0.0).set_p1_gate_rate(GateType::RZ, rz);
```

With a base rate of `0.0` and only an RZ gate rate of `0.3`, the expected measurement-flip probability is `0.2`. If gate rates were ignored it would be `0.0`.

## Suggested direction

Drop or narrow the `MemBuilder` gate-rate guard. The tests encode real behaviour and should not be changed to accommodate it. Narrowing is consistent with this branch's own second commit ("validate only consumed tables") -- the MemBuilder guard simply predates MEM becoming a consumer.

The `dev` replacement-branch guard is unaffected and should stay.

Everything else is clean on the merged tree: `python-ci-lint`, workspace clippy with `-D warnings`, and `cargo fmt --check` all pass. `pytest-ci-core` aborts on the same collision.

No changes were pushed; the branch on origin is untouched.

ciaranra:
Merged current `dev` and resolved the collision reported above. I have taken authorship of this branch.

## Changes

**The `MemBuilder` gate-rate guard is removed.** Its premise -- that MEM ignores per-gate rate tables -- is no longer true. MEM resolves them on both paths: `rates_1q_for_operation` (`mem_builder.rs:243`) and `rates_2q_for_operation` (`:338`). The guard was rejecting configurations the builder handles correctly. The comment claiming the two-qubit path reads `p2 / 15` "without gate-rate lookups" was contradicted by the line below it and is gone with it.

The `dev` replacement-branch guard is kept, and the raw-measurement guard in `sampler.rs` from this branch's third commit is untouched -- that half of the commit remains valid.

**The doc now states the actual contract**, naming the two resolvers, so the next reader does not have to rediscover whether MEM honours rate tables.

**`gate_rate_mem_builder_rejects_nonzero_tables` is replaced, not deleted**, by `gate_rate_mem_builder_honors_nonzero_tables`. The old test asserted the wrong contract; deleting it would leave no coverage where coverage existed. The replacement checks that nonzero tables build *and* that raising the scheduled RZZ rate raises the measurement-flip total -- asserting only that `build()` succeeds would pass even if the rates were parsed and dropped.

Mutation check: pointing the two-qubit lookup at a gate type that never matches makes the test fail with `0.006917442001781538 !> 0.006917442001781538` -- identical totals, so the assertion pins the effect rather than the return value.

## Merge conflict resolution

`MemBuilder::build` had gained a guard on both sides at the same spot: this branch's gate-rate check and `dev`'s `BranchImpact` / `ExactBranchReplay` check. They are independent, so both were kept and the doc sentences merged. Taking either side alone would have compiled and passed most tests while silently deleting a guard.

## Verification

- `just rstest debug`: 12,835 passed; only the known `gpu_density_matrix_two_qubit_roots_preserve_rotation_channels` tolerance failure present on `dev`.
- `just pytest-ci-core`: 9,269 passed, no failures.
- `just python-ci-lint`, `pre-commit run --all-files`, workspace clippy with `-D warnings`, `cargo fmt --check`: all clean.

Two failures seen during this work were stale build artefacts, not defects: the worktree's `pecos-rslib` and `pecos-rslib-exp` extensions predated the `dev` merge (the latter dated 09-10 against a 09-14 merge), so CRZ boundary tests failed against binaries lacking #743. Both pass after rebuilding.

`phase_u_neo_and_dem_agree_on_noise` is `#[cfg(feature = "neo")]`; it was confirmed to actually execute and pass under `just rstest debug` rather than being silently skipped.

ciaranra:
Second review round complete, plus a disclosed behaviour change.

## Behaviour change: MEM inherits DemBuilder's strictness

Sharing `validate_gate_rate_keys` means `MemBuilder` now rejects everything `DemBuilder` rejects -- which is the point -- but that includes a case MEM previously accepted and handled correctly.

For a circuit scheduling both `RZZ` and `SZZ`, supplying explicit rates for **both**:

```rust
NoiseConfig::new(0.0, 0.0, 0.0, 0.0)
    .set_p2_gate_rate(GateType::RZZ, 0.01)
    .set_p2_gate_rate(GateType::SZZ, 0.02)
```

is now rejected, with a message advising the caller to "add a key for RZZ in p2_gate_rates" -- a key that is already present. Reproduced directly:

```
p2_gate_rates key SZZ matches scheduled SZZ gates but also names the Clifford
action of RZZ at node 3. ... add a key for RZZ in p2_gate_rates, or lower the
circuit with lower_clifford_rotations() before building.
```

Previously MEM resolved each rate by its scheduled gate type and both entries had independent consumers.

This is not new policy. It is `DemBuilder`'s deliberate rule from `a3900b038` ("Reject a gate-rate key naming another scheduled gate's Clifford action even when it also matches"), now applied consistently. The judgement here is that a loud rejection is better than the previous behaviour, which was to silently ignore a mis-keyed entry and produce a wrong noise model without a word. The over-strictness is real but pre-existing, and narrowing it belongs in one change covering both builders rather than a MEM-only special case: tracked in #773.

Anyone keying both a Clifford action and its scheduled gate will need to adjust. That is the cost of the consistency.

`gate_rate_mem_builder_inherits_dem_strictness_for_mixed_scheduled_gates` pins this. Its doc comment states explicitly that it records current behaviour rather than a desired contract, and that it changes with #773 if that rule is narrowed -- so the next person to narrow it sees a record, not an apparent contract they are breaking.

## Verification

Current `dev` merged in. `just rstest debug`: 12,861 passed, only the known `gpu_density_matrix_two_qubit_roots_preserve_rotation_channels` tolerance failure. `just pytest-ci-core`: 9,255 passed. Workspace clippy with `-D warnings`, `cargo fmt --check`, and `pre-commit run --all-files` all clean.

## Review scope note

The second review covered only the post-review commit (`100485732`), since the earlier round had already covered the rest of the branch and the intervening range was almost entirely `dev`-merge traffic from other work. Its one finding is the mixed-gate case above.

PR #785, opened 2026-09-16
Add optional Frontier and BP-Trellis Python batch decoders

qciaran:
Frontier and BP-Trellis can now decode Python SampleBatch data and fused DemSampler shots with multiple workers. Their configuration factories and native workers live in the optional pecos-rslib-exp extension:

```python
from pecos_rslib_exp import frontier, bp_trellis

result = batch.decode(dem_text, frontier(k=64), workers=4)
```

Both factories expose the underlying decoder configuration, including column ordering and BP-Trellis escalation. Standard pecos.decoders imports work without the experimental package; explicitly importing frontier or bp_trellis from that module lazily loads the optional extension.

The stable extension accepts a versioned Python decoder provider and retains shared batch execution, retries, scoring, and prediction handling. Each worker builds its own native decoder. Native construction and decoding release the GIL, with a Python call boundary per shot. Experimental specs currently support batch and fused sampling APIs, rather than stable DecoderSpec parsing or composition. The publishable pecos-decoders crate has no experimental dependencies.

The Guppy DEM workflow demonstrates both decoders when the optional extension is available and remains runnable without it. The decoder guide documents configuration and the optional imports.

Validation:
- 129 standard decoder tests passed, 2 skipped; the provider suite subsequently passed all 5 tests with an added package-publication regression check.
- 49 experimental tests passed, covering parallel parity, direct binding parity, wide observables, retries, and GIL release.
- Full Guppy workflow passed with and without the experimental extension.
- Clippy for both Python extensions with all targets and warnings denied; rustfmt, Ruff, Black, and diff checks passed.


ciaranra:
Continued in #789. This pull request closed when its head branch was renamed from `codex/trellis-python-batch-decoding` to `trellis-python-batch-decoding`; the commits are unchanged.

